PR #7, opened 2015-01-29
Adding option to enable caching telemetry when being offline

SofianHn:
For Node.js application that expect intermittent internet connection, this change enables the App Insights SDK to be tolerant to the user being offline. When this option is enabled, the SDK caches the events in the tmp directory and upon the next successful post, it retries to submit all the data that occured while being offline. 

This change contains: 
1- A fix to some typescript definition files. npm test is not working due to some type mismatch. 
2- Adding cacheOnError config option support and adding the appropriate E2E tests.


scsouthw:
Thanks, Sofian!

PR #10, opened 2015-02-22
Fix up TypeScript typings in several files

lukehoban:
There were several TypeScript typing issues in the codebase, this PR addresses a number of them through two primary changes:

* Update the copy of `async.d.ts` to the latest from DefinitelyTyped, which has a few fixes
* Turn on `--noimplicitany` in the `npm test` script, so that additional potential errors are identified, and fixes all the places where this was triggering warnings


scsouthw:
Thanks, Luke!

PR #21, opened 2015-04-28
Feature/v15 schema and auto collection

scsouthw:
There are a lot of changes in here, several of which break the existing API. This aligns the API more closely with the Android/iOS SDKs and adds better support for customizing the auto-collection features. Also, there is a prototype implementation for collecting system performance metrics (CPU/Memory/RPS) and a stub for collecting console activity as traces (is anyone interested in that?). Another big change is that this moves away from dependency on the browser JS SDK (which was downloaded on install). This works better with commonJs and simplifies setup.

The new API looks something like this:
```javascript
import AppInsights = require("./ApplicationInsights");
AppInsights.setup("b7040ad8-a016-4057-903f-35edb58a6007")
    .setAutoCollectRequestsEnabled(true) // default is true 
    .setAutoCollectPerformanceEnabled(true) // default is true
    .setAutoCollectExceptionsEnabled(true) // default is true
    .enableVerboseLogging() // optional
    .start(); // no telemetry will be sent until .start is called allowing auto-collection initialization to be prevented if necessary

AppInsights.instance.trackEvent("test event");
AppInsights.instance.trackException(new Error());
AppInsights.instance.trackMetric("test metric", 3);
AppInsights.instance.trackTrace("test trace");
```

@lukehoban @SofianHn @jeffwilcox @ctaggart @mrmckeb @mousetraps I'd greatly appreciate any feedback/input on the changes. It's not quite ready to merge yet and needs a lot more test coverage. I'm moving to mocha/chai/sinon for tests which should make things easier.


package.json
```diff
@@ -1,17 +1,18 @@
 {
   "name": "applicationinsights",
   "license": "MIT",
-  "bugs": "https://github.com/Microsoft/AppInsights-node.js/issues",
+  "bugs": "https://github.com/Microsoft/ApplicationInsights-node.js/issues",
   "version": "0.12.5",
```
jeffwilcox: If there are breaking changes, shouldn't there be an increment of some variety for this release?

server.ts
```diff
@@ -0,0 +1,35 @@
+///<reference path='.\Declarations\node\node.d.ts' />
```
jeffwilcox: And a general comment, attaching to `server.ts` but applicable to any: might be worth placing the MIT license header on each file.

server.ts
```diff
@@ -0,0 +1,35 @@
...
+AppInsights.instance.trackTrace("test trace");
+
+var count = 0;
+var last = 0;
+var server = http.createServer((req: http.ServerRequest, res: http.ServerResponse) => {
+    count++;
+    var now = +new Date;
+    if(now - last > 1000) {
+        last = now;
+        console.log(count + " requests processed");
```
jeffwilcox: Instead of using `console.log`, might be worth using a npm module such as `debug` (https://www.npmjs.com/package/debug); Express and some other middleware engines use it and make it easier to show/hide on demand with environment configuration.

Tests/applicationInsights.tests.ts
```diff
@@ -0,0 +1,64 @@
...
+        afterEach(() => AppInsights.instance = undefined);
+        after(() => warnSpy.restore());
+
+        it('should not warn if setup is called once', () => {
+            AppInsights.setup("key");
+            assert.ok(warnSpy.notCalled, "warning was not raised");
+        });
+
+        it('should warn if setup is called twice', () => {
+            AppInsights.setup("key");
```
jeffwilcox: And super random consistency question/pt: not sure why `"key"` is using double quotes mixed with single quotes.

scsouthw: Thanks, https://github.com/Microsoft/ApplicationInsights-node.js/commit/9d7bce1329edd520f38e97ae157d02828be1faac makes double quote use consistent

server.ts
scsouthw: I was told at some point to either add the license per-file or add a license.txt to the repository root. I'll double check though.

package.json
scsouthw: definitely, thanks for the reminder. This will be 0.15 to align with the c#/browser JS release.

PR #30, opened 2015-08-18
Remove sync fileops

SofianHn:
Adding tests for the offline mode feature


scsouthw:
We should keep a sync file saving mechanism for handling crashes even if we add an async one for this purpose. File operations don't complete if node is crashing in v12. The process is terminated as soon as the crash handler returns.

SofianHn:
I will have another PR to save data sync only when crashing

PR #43, opened 2015-10-21
add logging to channel to determine when it is flushing during flakey…

lmillott:
… test (don't review this PR. It's just for finding out why tests are flakey on the travis build)


lmillott:
Aborting this PR. Will create a new one with a single cleaned up changeset

PR #44, opened 2015-10-22
Increase end to end test timeout from 2 to 10 seconds

lmillott:
The end to end tests are having intermittent failures on Travis builds. Increasing the timeout from 2 to 10 seconds seems to fix the issue, 4 builds in a row passed after increasing it.


lmillott:
this was for testing the tests on the travis build. I moved the changes to a new clean branch to avoid the history pollution

PR #47, opened 2015-11-05
JSON order of envelope properties

sergey-shandar:
Some channels require a specific order of envelope properties. In particular, the “data”, “tags” properties must all come after the "name" and "time" properties.

### DIFF
--- a/Library/Channel.ts
+++ b/Library/Channel.ts
@@ -20,7 +20,7 @@ class Channel {
         this._getBatchIntervalMs = getBatchIntervalMs;
         this._sender = sender;
     }
-    
+
     /**
      * Enable or disable offline mode
      */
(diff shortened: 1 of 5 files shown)


scsouthw:
:+1: thanks for fixing this @sergey-shandar!

sergey-shandar:
No problem. Do you know when we may have an official NPM package? I would like to fix 2 more issues:
- Change event name from “Microsoft.ApplicationInsights.Event” to “Microsoft.ApplicationInsights.{KEY}.Event”.
- Replace a sequence number format.

scsouthw:
@BogdanBe has been managing the release of new versions. He could help get these changes published and coordinate the release after your other fixes are in.

PR #68, opened 2016-02-11
Fixing bug that throws unhandled rejections

rzio:
In the previous version that was released today there was a 'bug' that threw unhandled promise rejections, as if they were exceptions. This changes the behavior of node, since things that were silently swallowed before now are thrown as unhandled exceptions and can crash node.


BogdanBe:
Thank you!

PR #71, opened 2016-03-17
Add capability to pass custom properties to the trackMetric api

lmillott:
We would like to add custom properties to our metrics telemetry. Currently the contracts data model for metrics in the SDK has properties on the MetricsData object, but the api doesn't allow them as arguments.


phmiller:
Great! PR looks good to me

Issue #74, opened 2016-03-22
Cannot track request with custom properties determined during request processing


lmillott:
In our node server we collect various properties about the request throughout processing and place them in a context object. Whenever we log any app insights telemetry we add this context to the custom property bag. The only type for which this doesn't work is requests.

For requests tracking, we want to do something like this:

```javascript
var context;
// process the request and add items to logging context throughout the pipeline
response.on('finish', function() {
    appInsights.client.trackrequest(request, response, context);
});
```

Unfortunately this doesn't work because in the trackRequest implementation it actually listens for the response.on('finish') event which has already fired at this point. Here's the relevant snippet from the trackRequest method in the sdk:
```javascript
if (response && response.once) {
    response.once("finish", function () { return processRequest(); });
}
```

Additionally we want to set the operationId, userId, and operationParentId to our own custom values so we can correlate the request telemetry with other telemetry from inside and outside this application:
```javascript
appInsights.client.context.tags[appInsights.client.context.keys.operationId] = loggingContext.requestId;
appInsights.client.context.tags[appInsights.client.context.keys.userId] = loggingContext.userId;
appInsights.client.context.tags[appInsights.client.context.keys.operationParentId] = loggingContext.clientRequestId;
```

But this doesn't work because the request is actually logged asynchronously and this context could have been reset by another request at that point.

I'm open to suggestions on how to implement this either in our client, or in the SDK itself. I think we basically need a trackRequest method that is synchronous.

PR #75, opened 2016-03-23
Track request sync

lmillott:
We need to be able to track requests in a synchronous call because we have custom properties that we want to add in the response.finish event, at which point it is too late to use the async track request method. Additionally you cannot set the context tag keys for the request reliably with the async way, because they could be reset by another request before the request is actually tracked.

More details in this issue: https://github.com/Microsoft/ApplicationInsights-node.js/issues/74

### DIFF
--- a/AutoCollection/Requests.ts
+++ b/AutoCollection/Requests.ts
@@ -67,9 +67,41 @@ class AutoCollectRequests {
             });
 	    }
     }
+    
+    /**
+     * Tracks a request synchronously (doesn't wait for response 'finish' event)
+     */
+    public static trackRequestSync(client: Client, request: http.ServerRequest, response:http.ServerResponse, ellapsedMilliseconds?: number, properties?:{ [key: string]: string; }, error?: any) {
+        if (!request || !response || !client) {
+            Logging.info("AutoCollectRequests.trackRequestSync was called with invalid parameters: ", !request, !response, !client);
+            return;
+        }
+        
+        // store data about the request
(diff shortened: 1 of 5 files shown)


AutoCollection/Requests.ts
```diff
@@ -67,53 +67,57 @@ class AutoCollectRequests {
...
 	    }
     }
-
+    
     /**
-     * Tracks a request
+     * Tracks a request synchronously (doesn't wait for response 'finish' event)
      */
-    public static trackRequest(client:Client, request:http.ServerRequest, response:http.ServerResponse, properties?:{ [key: string]: string; }) {
+    public static trackRequestSync(client: Client, request: http.ServerRequest, response:http.ServerResponse, ellapsedMilliseconds?: number, properties?:{ [key: string]: string; }, error?: any) {
```
kamilszostak: I'm new to Node.js and this SDK and I'm not sure what's the naming convention. But in C# is recommended to add "Async" suffix to async methods, not "Sync" to those which are executed synchronously.

kamilszostak:
:shipit:

AutoCollection/Requests.ts
lmillott: I modeled this after the node file system api, where it's sort of opposite from C#. For example fs.writefile writes a file asynchronously, and fs.writefilesync writes it synchronously. Also this way I don't have to make a breaking API change by renaming trackRequest to trackRequestAsync

kamilszostak: Got it. Thanks!

PR #76, opened 2016-03-24
Fill required StackFrame fields with default values

lukekim:
This change ensures the method and fileName fields on StackFrame are populated so the item will not be rejected by dc.services.visualstudio.com


MaxShehovtsov:
:shipit:

dnduffy:
:shipit:

PR #78, opened 2016-03-24
Upgrade TypeScript to 1.8.9+

lukekim:


package.json
```diff
@@ -38,6 +38,6 @@
     "mocha": "^2.2.4",
     "node-mocks-http": "^1.2.3",
     "sinon": "^1.14.1",
-    "typescript": "^1.0.1"
+    "typescript": "^1.8.9"
```
dnduffy: Where is the 1.8.9 version number recorded (I can only find documentation mentioning 1.8 and no revision number)?

lukekim: What do you mean? https://www.npmjs.com/package/typescript

dnduffy: Thank-you that was what I was looking for.

PR #83, opened 2016-04-07
Added ability to change offline cache directory. Offline file processing on start()

jsheely:
- Fixed directory typo. 
- Added ability to change offline cache directory
- Added offline file processing on start()

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -178,10 +178,10 @@ class Sender {
         });
     }
 
-    private _confirmDirExists(direcotry: string, callback: (err) => void): void {
-        fs.exists(direcotry, (exists) => {
+    private _confirmDirExists(directory: string, callback: (err) => void): void {
+        fs.exists(directory, (exists) => {
             if (!exists) {
-                fs.mkdir(direcotry, (err) => {
+                fs.mkdir(directory, (err) => {
                     callback(err);
                 });
             } else {
@@ -191,7 +191,7 @@ class Sender {
     }
 
     /**
-     * Stores the payload as a json file on disk in the temp direcotry
+     * Stores the payload as a json file on disk in the temp directory
      */
     private _storeToDisk(payload: any) {
 
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -13,13 +13,14 @@ import Logging = require("./Logging");
 class Sender {
     private static TAG = "Sender";
     // the amount of time the SDK will wait between resending cached data, this buffer is to avoid any throtelling from the service side
-    public static WAIT_BETWEEN_RESEND = 60 * 1000; 
+    public static WAIT_BETWEEN_RESEND = 60 * 1000;
     public static TEMPDIR: string = "appInsights-node";
-    
+    public static OFFLINE_DIRECTORY: string = null;
+
     private _getUrl: () => string;
     private _onSuccess: (response: string) => void;
     private _onError: (error: Error) => void;
-    private _enableOfflineMode: boolean; 
+    private _enableOfflineMode: boolean;
 
     constructor(getUrl: () => string, onSuccess?: (response: string) => void, onError?: (error: Error) => void) {
         this._getUrl = getUrl;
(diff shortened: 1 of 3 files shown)


jsheely:
Items to consider for the future.

- concat all ai.json file payloads into a single request send()
- add boolean option for initial offline flush.

Library/Sender.ts
```diff
@@ -128,59 +137,99 @@ class Sender {
...
+                        }
+                    } else {
+                        this._onErrorHelper(error);
+                    }
+                });
+            }
+        });
+    }
+
     private _confirmDirExists(direcotry: string, callback: (err) => void): void {
```
rotozoom: Consider fixing this 'direcotry' also.

Library/Sender.ts
```diff
@@ -128,59 +137,99 @@ class Sender {
...
+            if (exists) {
+                fs.readdir(tempDir, (error, files) => {
+                    if (!error) {
+                        files = files.filter(f => path.basename(f).indexOf(".ai.json") > -1);
+                        if (files.length > 0) {
+                            files.forEach(file => {
+                                var filePath = path.join(tempDir, file);
+                                fs.readFile(filePath, (error, payload) => {
+                                    if (!error) {
+                                        // delete the file first to prevent double sending
```
rotozoom: What is better: lost of all telemetry or send the same telemetry several times? In the case when file has already been deleted but send is failed are we going to create another file or just lost current content?

jsheely: I would think that double telemetry would be better than losing any results. You may be using telemetry to alert on errors. I would rather have two errors then no errors reported.

However neither option is perfect. A new routine that is capable of cleaning cleaning up files only after a send is successful based on the contents of that file. Otherwise leave file on disk.

Library/Sender.ts
```diff
@@ -128,59 +137,99 @@ class Sender {
...
-    
-    private _confirmDirExists(direcotry: string, callback: (err) => void): void {
-        fs.exists(direcotry, (exists) => {
+
+    /**
+     * Send all temp files from last crash.
+     * reads all files that exist, deletes them and tries to send its load
+     */
+    public sendAllFilesOnDisk(): void {
+        var tempDir = Sender.OFFLINE_DIRECTORY || path.join(os.tmpDir(), Sender.TEMPDIR);
```
kamilszostak: Could you check if `_enableOfflineMode` is set to `True `before sending files from disk? I've noticed that currently we only check it in the Sender.send() method. IMHO we should not store any data on disk if offline mode is disabled.

Library/Sender.ts
```diff
@@ -128,59 +137,99 @@ class Sender {
...
             }
         });
     }
-    
+
     /**
-     * Stores the payload as a json file on disk in the temp direcotry
+     * Stores the payload as a json file on disk in the temp directory
      */
     private _storeToDisk(payload: any) {
```
kamilszostak: Same as above - check if the Offline mode is enabled.

kamilszostak:
Thank you for contributing! Please resolve my comments and :shipit: 

Could you also create two new issues for those future items listed above?

Library/Sender.ts
dnduffy: I would suggest sending first and then deleting.

jeffwilcox:
Hi,
There's a `develop` branch for this project, but this PR is against `master`. Could a maintainer provide input into the branching structure for contributions?

kamilszostak:
@jeffmcaffer I described the branching structure in #115.

Library/Sender.ts
AlexBulankou: @KamilSzostak, there's a subtlety here on whether _enableOfflineMode flag should apply to saving or whether we should also consider it when reading from disk. I think I'm with you here, that _enableOfflineMode=False means, **don't touch my file system**

OsvaldoRosado:
Closing this PR as it's fallen quite out of date with `develop` by now - especially with all of the changes made to disk retry behavior in the interim. Will continue tracking the enhancement it intended to provide via #307

Issue #84, opened 2016-04-19
Client.sendPendingData doesn't call callback when there is no data to send.


rotozoom:
When there is no data to send Client.sendPendingData doesn't call callback which is passed as a paramerter. In that case it is impossible to determine whether we complete work with Application Insights Client or not.

PR #86, opened 2016-04-29
Fix issue #84

rotozoom:
Calling final callback in any case, when sendPendingData is called.


rotozoom:
To fix Issue #84

AlexBulankou:
:shipit:

PR #87, opened 2016-04-29
SDK logs absolute URL instead of relative URL

kamilszostak:


lmillott:
Looks good to me! :shipit:

PR #91, opened 2016-05-20
Add support for Error subclasses

### LINKED ISSUE #90: Exception type and failed method incorrect for objects that are instances of Error subclass
Some modules ([example](https://github.com/jaredhanson/passport-windowslive/blob/master/lib/errors/liveconnectapierror.js)) subclass `Error`.  When `obj` is an instance of such, [`Util.isError(...)`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/master/Library/Util.ts#L96) returns `false`.

    // "[object Object]" !== "[object Error]"
    return Object.prototype.toString.call(obj) === "[object Error]";

This issue causes the "Exception type" and "Failed method" properties in Azure Portal's Application Insights blade to display the error type and method name from [this line](https://github.com/Microsoft/ApplicationInsights-node.js/blob/master/Library/Client.ts#L90) in the Application Insights SDK for Node.js:

    Error at Client.trackException

...instead of from the original error.  Example:

    OriginalErrorTypeName at originalFunctionName

I believe the implementation of `Util.isError(...)` should be:

    return obj instanceof Error;

dtivel:
Fixes #90.

### DIFF
--- a/Library/Util.ts
+++ b/Library/Util.ts
@@ -94,7 +94,7 @@ class Util {
      * Check if an object is of type Error
      */
     public static isError(obj:any):boolean {
-        return Object.prototype.toString.call(obj) === "[object Error]";
+        return obj instanceof Error;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kamilszostak:
:shipit:

dtivel:
I lack write access to this repository.  If these changes are acceptable, can you please merge the pull request?

Issue #99, opened 2016-06-14
Just do not send any devicer type


SergeyKanzhelev:
[Here](https://github.com/Microsoft/ApplicationInsights-node.js/pull/98) do not set any device type as we do in .NET. It will assume `PC` than

PR #101, opened 2016-06-14
Don't send device.type, PC is used by default

kamilszostak:
#99

### DIFF
--- a/Tests/Library/Client.tests.ts
+++ b/Tests/Library/Client.tests.ts
@@ -247,6 +247,21 @@ describe("Library/Client", () => {
                 assert.equal(duration, 10);    
             });
             
+            it('should track request with correct tags on response finish event', function () {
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
:shipit:

Tests/Library/Client.tests.ts
```diff
@@ -247,6 +247,21 @@ describe("Library/Client", () => {
...
+                client.trackRequest(<any>request, <any>response, properties);
+                
+                // emit finish event
+                response.emitFinish();
+                
+                // validate
+                var args = trackStub.args;
+                var tags = args[0][1];
+                assert.equal(tags["ai.device.id"], "node");
+                assert.equal(tags["ai.device.type"], null);
```
dnduffy: So the server will default this to 'PC' or the reporting will default it?

PR #109, opened 2016-08-08
Fixes #102 Removes default ai.device.id from node

### LINKED ISSUE #102: Don't send device.id
The `ai.device.id` is not used anymore and we should leave it blank, instead of setting it to `node`.

sudheesh001:


dnduffy:
:shipit:

kamilszostak:
:shipit:

PR #110, opened 2016-08-29
Add an option to disable automatic stack trace collection

kamilszostak:


AutoCollection/Exceptions.ts
```diff
@@ -68,20 +73,23 @@ class AutoCollectExceptions {
...
 
-        var stack = error["stack"];
         var exceptionDetails = new ContractsModule.Contracts.ExceptionDetails();
         exceptionDetails.message = error.message;
         exceptionDetails.typeName = error.name;
-        exceptionDetails.parsedStack = this.parseStack(stack);
+
+        if (AutoCollectExceptions.IsStackTrackCollectionEnabled) {
+            var stack = error["stack"];
+            exceptionDetails.parsedStack = this.parseStack(stack);
```
dnduffy: If "stack" isn't going to be reused you could just make it a one liner.

applicationinsights.ts
```diff
@@ -97,12 +97,13 @@ class ApplicationInsights {
     /**
      * Sets the state of exception tracking (enabled by default)
      * @param value if true uncaught exceptions will be sent to Application Insights
+     * @param enableStackTrackCollection if true a stack trace will be added to each exception  
      * @returns {ApplicationInsights} this class
      */
-    public static setAutoCollectExceptions(value: boolean) {
+    public static setAutoCollectExceptions(value: boolean, enableStackTrackCollection = true) {
```
dnduffy: Now that there are two parameters, and the second one has a detailed name, would it be good to also give the first one a detailed name?

dnduffy:
👍

kamilszostak:
I created a more generic solution to handle this problem - see https://github.com/Microsoft/ApplicationInsights-node.js#telemetry-processor

PR #112, opened 2016-08-29
Add Telemetry processors

kamilszostak:


Library/Client.ts
```diff
@@ -236,6 +259,32 @@ class Client {
...
+                if (processor) {
+                    if (processor.apply(null, [envelope]) === false) {
+                        accepted = false;
+                        break;
+                    }
+                }
+            }
+        } catch (error) {
+            accepted = false;
+            Logging.warn("One of telemetry processors failed, telemetry item will not be sent.", error, envelope);
```
dnduffy: Is there ever a situation in which you would want to try each processor individually and send at least partially successful data rather than nothing?

README.md
```diff
@@ -60,6 +60,36 @@ client.trackMetric("custom metric", 3);
...
 
+### Telemetry Processor
+
+```javascript
+public addTelemetryProcessor(telemetryProcessor: (envelope: ContractsModule.Contracts.Envelope) => boolean)
+```
+
+Adds telemetry processor to the collection. Telemetry processors will be called one by one before telemetry item is pushed for sending and in the order they were added. 
+If one of telemetry processors returns false then telemetry item will not be sent. 
+If one of telemetry processors throws an error then telemetry item will not be sent.
```
dnduffy: "...one of the telementry..."

README.md
```diff
@@ -60,6 +60,36 @@ client.trackMetric("custom metric", 3);
 client.trackTrace("trace message");
 ```
 
+### Telemetry Processor
+
+```javascript
+public addTelemetryProcessor(telemetryProcessor: (envelope: ContractsModule.Contracts.Envelope) => boolean)
+```
+
+Adds telemetry processor to the collection. Telemetry processors will be called one by one before telemetry item is pushed for sending and in the order they were added.
```
dnduffy: "Adds a telementry..." ... "...before the telemetry..."
I might also move the bit about "in the order they were added" to just after "one by one" so it reads "...one by one, in the order they were added, before the telemetry item is pushed...".

README.md
```diff
@@ -60,6 +60,36 @@ client.trackMetric("custom metric", 3);
...
+public addTelemetryProcessor(telemetryProcessor: (envelope: ContractsModule.Contracts.Envelope) => boolean)
+```
+
+Adds telemetry processor to the collection. Telemetry processors will be called one by one before telemetry item is pushed for sending and in the order they were added. 
+If one of telemetry processors returns false then telemetry item will not be sent. 
+If one of telemetry processors throws an error then telemetry item will not be sent.
+
+**Example**
+
+Add the below code before you send any telemetry, it will remove stack trace information from any Exception reported by the SDK.
```
dnduffy: Didn't you just add the below code to the SDK?  Why would anyone need to copy and paste it?

Tests/Library/Client.tests.ts
```diff
@@ -186,55 +192,55 @@ describe("Library/Client", () => {
...
         function parseDuration(duration: string): number {
             if (!duration) {
                 return 0;
             }
-            
-            var parts = duration.match("(\\d\\d):(\\d\\d):(\\d\\d).(\\d\\d\\d)");  
-            return parseInt(parts[1])*60*60*1000 + parseInt(parts[2])*60*1000 + parseInt(parts[3])*1000 + parseInt(parts[4]);    
+
+            var parts = duration.match("(\\d\\d):(\\d\\d):(\\d\\d).(\\d\\d\\d)");
+            return parseInt(parts[1]) * 60 * 60 * 1000 + parseInt(parts[2]) * 60 * 1000 + parseInt(parts[3]) * 1000 + parseInt(parts[4]);
```
dnduffy: This could be re-parenthesized so that the * 60's and * 1000's aren't repeated for each value.
i.e. ((((parseInt(parts[1]) * 60) + parseInt(parts[2])) * 60) + parseInt(parts[3])) * 1000 + parseInt(parts[4])

dnduffy:
:+1:

SergeyKanzhelev:
I thought telemetry initializers already behaving like processors. If you return false from initializer - it will stop processing of the item and will not send it. Why processors?

kamilszostak:
@SergeyKanzhelev telemetry initializes are implemented in JavaScript SDK and they are not available in Node.js SDK. 

I thought that "processor" is a batter name in this case. They process already populated telemetry, so they don't do any "initialization".

Library/Client.ts
kamilszostak: Good point, IMHO if a first processor throws, there is no need to run other processors because the data may already be corrupted. I'll move the try/catch inside the _for_ loop, and leave the _break_ statement.

Tests/Library/Client.tests.ts
kamilszostak: What do you think about this? 

```javascript
    return parseInt(parts[1]) * 60 * 60 * 1000 + 
               parseInt(parts[2]) * 60 * 1000 +
               parseInt(parts[3]) * 1000 + 
               parseInt(parts[4]);
```

README.md
kamilszostak: This code is just an example of how a consumer of the SDK can write a telemetry processor.

SergeyKanzhelev:
@kamilszostak , got it. Thanks for explanation. It makes sense

Tests/Library/Client.tests.ts
dnduffy: It is a lot nicer to read but it still does all the redundant math which was what I was commenting on.

README.md
dnduffy: I see, from the comment above I thought it was instructions for something they should do rather than an example of how to write a processor.
For clarity I would suggest changing the start of the comment to read something like "The below code is an example of how to write a telemetry processor which will remove...".

PR #115, opened 2016-09-02
Describe branching structure

kamilszostak:


SergeyKanzhelev:
Should we use the same structure in JS repo?

dnduffy:
This would be useful in other repositories also. :)
:+1:

PR #116, opened 2016-09-06
[Sender] make resend interval configurable

xscript:
When in offline mode, resend interval for cached events is hard coded as 60 seconds. This is too long for small application/tools which only run for several seconds. Application will hang in there, being prevented from exiting.
Here making the resend interval configurable in *setOfflineMode* API to give more flexibility to end users.

### DIFF
--- a/Library/Channel.ts
+++ b/Library/Channel.ts
@@ -24,8 +24,8 @@ class Channel {
     /**
      * Enable or disable offline mode
      */
-    public setOfflineMode(value: boolean) {
-        this._sender.setOfflineMode(value);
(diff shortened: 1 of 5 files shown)


kamilszostak:
Thank you, looks good. Could you please add unit tests?

xscript:
Sure. Will add it soon.

xscript:
@kamilszostak I've added some unit tests. Please help review them. Thanks!

dnduffy:
:+1:

kamilszostak:
Thanks a lot! :shipit:

Issue #73, opened 2016-03-21
Documentation says that the trackException method takes a third parameter named measurements, but it is not actually defined in the method definition


haocs:
https://github.com/Microsoft/ApplicationInsights-node.js/blob/master/Library/Client.ts#L88

```javascript
/**
     * Log an exception you have caught.
     * @param   exception   An Error from a catch clause, or the string error message.
     * @param   properties  map[string, string] - additional data used to filter events and metrics in the portal. Defaults to empty.
     * @param   measurements    map[string, number] - metrics associated with this event, displayed in Metrics Explorer on the portal. Defaults to empty.
     */
    public trackException(exception:Error, properties?:{ [key: string]: string; }) {
        if(!Util.isError(exception)) {
            exception = new Error(<any>exception);
        }

        var data = ExceptionTracking.getExceptionData(exception, true, properties)
        this.track(data);
    }

```

jeffwilcox:
+1 would love to use this API as documented

PR #119, opened 2016-10-01
Resolve issue #73 about the TrackException method

gzepeda:
- Changed the definition of the trackException method to have the additional parameter called measurements.
- Changed the definition of the getExceptionData method to consider the measurements parameter
- No need to change the Exception Data Contract since it already had a measurements property


Tests/Library/Client.tests.ts
```diff
@@ -124,13 +124,20 @@ describe("Library/Client", () => {
...
-            assert.ok(trackStub.calledTwice);
+            assert.ok(trackStub.calledThrice);
 
             var args = trackStub.args;
+            
             assert.equal(args[0][0].baseData.exceptions[0].message, name);
+
             assert.equal(args[1][0].baseData.exceptions[0].message, name);
             assert.deepEqual(args[1][0].baseData.properties, properties);
+
```
dnduffy: Should these tests also check that the arguments which were not provided are indeed absent?

gzepeda: I believe it is a good idea, but probably for another test or in the same test?

kamilszostak: I would add a new test to cover both : missing properties and missing measurements. Thanks!

kamilszostak:
I changed the [release process](https://github.com/Microsoft/ApplicationInsights-node.js#branches) and all new features should go to a [develop](https://github.com/Microsoft/ApplicationInsights-node.js/tree/develop) branch. 

Could you please send a pull request to develop branch? 

Thanks!

gzepeda:
Closing to open a new one for the Develop Branch.

PR #121, opened 2016-10-04
Merge pull request to resolve #73

### LINKED ISSUE #73: Documentation says that the trackException method takes a third parameter named measurements, but it is not actually defined in the method definition

gzepeda:
- Changed the definition of the trackException method to have the additional parameter
- Changed the definition of the getExceptionData method to consider the measurements parameter
- No need to change the Exception Data Contract since it already had a measurements property
- Adding tests to verify calls to track exception with properties and calls to track properties and measurements

**_Not sure how to exclude previous commits out there_**


gzepeda:
I included some commits that were out there, I did not find the option to exclude them from the pull request.

gzepeda:
This is for issue #73

kamilszostak:
:shipit:

kamilszostak:
Previous commits are showing up because they are only in a master branch and I forgot to merge them back to the develop branch. Let's merge them back to develop. 

Thanks!

PR #129, opened 2016-10-28
HTTP dependency request tracking

jasongin:
This change adds automatic collection of dependency HTTP request data.

Please check that all the `RemoteDependencyData` fields are filled in correctly in the `ClientRequestParser.getDependencyData()` method. I didn't know where to find good documentation for those fields so in some cases I had to infer the correct usage by looking at other code.

The GitHub web diff view unfortunately doesn't compare some files that were renamed and have only minor changes:
- `Requests.ts` -> `ServerRequests.ts`
- `RequestDataHelper.ts` -> `ServerRequestParser.ts`
- `RequestDataHelper.tests.ts` -> `ServerRequestParser.ts`

Unit tests all pass. I also created a small test Node.js service to verify that both request and dependency telemetry is successfully uploaded. The dependencies do show up in the Application Map!


Library/Client.ts
```diff
@@ -131,17 +132,22 @@ class Client {
...
 
-    public trackRequestSync(request: http.ServerRequest, response: http.ServerResponse, ellapsedMilliseconds?: number, properties?: { [key: string]: string; }, error?: any) {
-        RequestTracking.trackRequestSync(this, request, response, ellapsedMilliseconds, properties, error);
+    public trackServerRequestSync(request: http.ServerRequest, response: http.ServerResponse, ellapsedMilliseconds?: number, properties?: { [key: string]: string; }, error?: any) {
+        ServerRequestTracking.trackRequestSync(this, request, response, ellapsedMilliseconds, properties, error);
     }
 
-    public trackRequest(request: http.ServerRequest, response: http.ServerResponse, properties?: { [key: string]: string; }) {
-        RequestTracking.trackRequest(this, request, response, properties);
+    public trackServerRequest(request: http.ServerRequest, response: http.ServerResponse, properties?: { [key: string]: string; }) {
```
OsvaldoRosado: Is this a breaking change? Would it be better to continue calling this trackRequest and merging the new trackClientRequest with trackDependency for compatibility with previous versions and consistency with AI terms for what these are?

jasongin: Yes, it's a breaking change. I suppose renaming this method is not that important.

However merging `trackClientRequest` with `trackDependency` doesn't make sense to me: the parameters are completely different, even though they end up doing something similar.

kamilszostak:
Thanks for sending a pull request! This is a really good addition to the Node.js SDK. 

I agree with @OsvaldoRosado, we should keep the naming consistent  with other AppInsights SDKs. This is "Request" (for all requests that node.js is receiving) and "Dependency" (for all requests the node.js is sending).

Library/Client.ts
```diff
@@ -131,17 +132,22 @@ class Client {
...
+    }
+
+    public trackClientRequest(requestOptions: any, request: http.ClientRequest, properties?: { [key: string]: string; }) {
+        ClientRequestTracking.trackRequest(this, requestOptions, request, properties);
     }
 
     public trackDependency(
         name: string,
         commandName: string,
+        target: string,
```
jasongin: The addition of the `target` parameter here is a breaking change. To avoid that it will need to be moved to an optional parameter at the end, although I think the intention of the new contract is actually that it should be required.

OsvaldoRosado:
> This is a really good addition to the Node.js SDK.

I agree, this is awesome 👍

Library/Client.ts
OsvaldoRosado: I don't think our bond currently asks for this field to be required. It might not be so bad to move this to the end and mark optional. Anyone have strong thoughts on that?

Library/Client.ts
kamilszostak: We should support both methods `trackRequest` and `trackDependency`. See [API Summary](https://azure.microsoft.com/en-us/documentation/articles/app-insights-api-custom-events-metrics/)

AutoCollection/ClientRequests.ts
```diff
@@ -0,0 +1,78 @@
...
+        this._initializeHttpModule(https);
+    }
+
+    private _initializeHttpModule(httpModule: any) {
+        const originalRequest = httpModule.request;
+        httpModule.request = (options, ...requestArgs) => {
+            const request: http.ClientRequest = originalRequest.call(
+                    originalRequest, options, ...requestArgs);
+            if (request) {
+                AutoCollectClientRequests.trackRequest(this._client, options, request);
```
jasongin: Possible, but I think that would make the code less consistent overall.

This dependency request code is following the same pattern as the already existing server request code. That code does not call the [`Client.trackRequest`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/cb64f4a5048a660da0d21c5113e608c8e9083d7c/Library/Client.ts#L138) method either: [the helper constructs a RequestData object directly instead](https://github.com/Microsoft/ApplicationInsights-node.js/blob/cb64f4a5048a660da0d21c5113e608c8e9083d7c/AutoCollection/RequestDataHelper.ts#L69).

Library/Client.ts
jasongin: Yes, but what about the `trackClientRequest` method I have added here? Without that, it would be difficult for an application to *manually* track HTTP dependency requests, because it takes a few dozen non-obvious lines of code to convert a `http.ClientRequest` into appropriate parameters for `trackDependency`. (Perhaps it could be renamed `trackDependencyRequest` instead of `trackClientRequest`.)

Library/Client.ts
jasongin: I don't see what "data" field you are referring to. `RemoteDependencyData` doesn't have a "data" field.

The `target` could be auto-filled by parsing the hostname out of the supplied `commandName` argument if it is a valid URI, or the `name` argument if it is a HTTP method + URI.

AutoCollection/ClientRequests.ts
kamilszostak: OK, let's keep it as it is.

Library/Client.ts
OsvaldoRosado: Sorry, the data field is one we're not using in this SDK yet, but is what commandName is migrating to eventually, as we deprecated it in the schema. I was just noting, as you did, that we can get the hostname out of commandName, based off of what the schema said the field should contain.

With this in mind, we should move target to the end and mark optional. Like @kamilszostak mentions below, this avoids breaking existing users.

Library/Client.ts
jasongin: I corrected the naming of `Client` methods. Now there is:
 * `trackRequest` - unchanged
 * `trackDependency` - added optional `target` parameter at the end
 * `trackDependencyRequest` - new, alternative to `trackDependency` that is appropriate to use if you have a `http.ClientRequest` object that you want to track manually (when autocollection is disabled).

Library/Client.ts
```diff
@@ -132,29 +133,33 @@ class Client {
...
         dependencyTypeName?: string,
         properties = {},
         dependencyKind = ContractsModule.Contracts.DependencyKind.Other,
         async = false,
-        dependencySource = ContractsModule.Contracts.DependencySourceType.Undefined) {
+        dependencySource = ContractsModule.Contracts.DependencySourceType.Undefined,
+        target: string = null) {
+
+        if (!target && commandName) {
+            target = url.parse(commandName).host;
```
OsvaldoRosado: Does url.parse always return an object?

Library/Client.ts
jasongin: OK, I moved `target` to the end. If not supplied it will attempt to auto-fill by parsing the `commandName`.

Library/Client.ts
jasongin: Yes. AFAIK it never throws unless you pass it a non-string argument. An incomplete URI is just parsed as a relative path, and any invalid characters get encoded.

AutoCollection/ClientRequestParser.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * Gets a dependency data contract object for a completed ClientRequest.
+     */
+    public getDependencyData(): ContractsModule.Contracts.Data<ContractsModule.Contracts.RemoteDependencyData> {
+        let urlObject = url.parse(this.url);
+        urlObject.search = undefined;
+        urlObject.hash = undefined;
+        let dependencyName = this.method + " " + url.format(urlObject);
```
jasongin: Fixed.

Library/Client.ts
jasongin: Done.

Tests/Library/Client.tests.ts
```diff
@@ -306,7 +315,7 @@ describe("Library/Client", () => {
                 assert.ok(trackStub.calledOnce);
                 var args = trackStub.args;
                 assert.equal(args[0][0].baseType, "Microsoft.ApplicationInsights.RequestData");
-                assert.equal(args[0][0].baseData.responseCode, 200);
```
kamilszostak: Is `responseCode` gone?

jasongin: It's not gone, but it is not 200 anymore. This test is simulating a failed request, so the `responseCode` never should have been 200. My refactor of the `RequestDataHelper`/`RequestParser` class fixed this bug.

kamilszostak:
I added a small comment about responseCode in `trackRequest()` unit test. Everything else looks fine. 

 :shipit:

jasongin:
Should this include a minor version bump in package.json, since it is a significant new feature?

kamilszostak:
No, no need to change the version number. I usually change it when I merge from a develop to master branch.

jasongin:
Closing this, because I have resubmitted as a new PR https://github.com/Microsoft/ApplicationInsights-node.js/pull/130, to target the develop branch.

PR #130, opened 2016-11-02
HTTP dependency request tracking

jasongin:
I'm resubmitting this as a new PR to target the develop branch instead of the master branch. See review comments on the original PR at https://github.com/Microsoft/ApplicationInsights-node.js/pull/129. There are no changes here compared to that one, aside from rebasing on top of develop.


kamilszostak:
:shipit:

AutoCollection/ClientRequestParser.ts
```diff
@@ -0,0 +1,119 @@
...
+        let remoteDependency = new ContractsModule.Contracts.RemoteDependencyData();
+        remoteDependency.target = urlObject.hostname;
+        remoteDependency.name = dependencyName;
+        remoteDependency.commandName = this.url;
+        remoteDependency.value = this.duration;
+        remoteDependency.success = this._isSuccess();
+        remoteDependency.async = true;
+        remoteDependency.dependencyTypeName =
+            ContractsModule.Contracts.DependencyKind[ContractsModule.Contracts.DependencyKind.Http];
+        remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
```
brahmnes: I believe a new dependency type, something called AI dependency, is being introduced, and that would be used instead of HTTP for communications between AI components.

Check with Nizar on the details.

jasongin: Yes, I'll include that in the follow-up change that adds component correlation. But for now this code doesn't know when a dependency is an AI component or not.

AutoCollection/ServerRequests.ts
```diff
@@ -0,0 +1,125 @@
...
+            return;
+        }
+
+        // store data about the request
+        var requestParser = new ServerRequestParser(request);
+
+        // response listeners
+        if (response && response.once) {
+            response.once("finish", () => {
+                AutoCollectServerRequests.endRequest(client, requestParser, response, null, properties, null);
```
brahmnes: I just noticed that in both the Server and Client auto collectors, it's currently not parsing or generating a hashed ikey and on the HTTP headers. This will be needed for stitching different AI components in the map. Perhaps that's the next step?

jasongin: Yes, that will be in a separate PR. I'm testing that now.

PR #131, opened 2016-11-03
Add source & target ikey headers for correlation

jasongin:
* Add source field to the `RequestData` contract
 * Add ikey hash to incoming request response headers and dependency request headers
 * Check for ikey hash headers on incoming requests and returned dependency responses, and record in the respective telemetry `source` and `target` fields
 * Set dependency type to ApplicationInsights if an ikey header was received in a dependency response
 * Fix some bugs related to slightly different `https.request` behavior on v0.10.x

### DIFF
--- a/AutoCollection/ClientRequestParser.ts
+++ b/AutoCollection/ClientRequestParser.ts
@@ -7,12 +7,15 @@ import ContractsModule = require("../Library/Contracts");
 import Client = require("../Library/Client");
 import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
+import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import RequestParser = require("./RequestParser");
 
 /**
  * Helper class to read data from the requst/response objects and convert them into the telemetry contract
  */
 class ClientRequestParser extends RequestParser {
+    private targetIKeyHash: string;
+
     constructor(requestOptions: string | Object, request: http.ClientRequest) {
         super();
         if (request && requestOptions) {
@@ -36,6 +39,8 @@ class ClientRequestParser extends RequestParser {
      */
     public onResponse(response: http.ClientResponse, properties?: { [key: string]: string }) {
         this._setStatus(response.statusCode, undefined, properties);
+        this.targetIKeyHash =
+            response.headers && response.headers[RequestResponseHeaders.targetInstrumentationKeyHeader];
     }
 
     /**
@@ -48,15 +53,26 @@ class ClientRequestParser extends RequestParser {
         let dependencyName = this.method.toUpperCase() + " " + url.format(urlObject);
 
         let remoteDependency = new ContractsModule.Contracts.RemoteDependencyData();
-        remoteDependency.target = urlObject.hostname;
+
+        if (this.targetIKeyHash) {
+            // Currently the service supports a dependency type name of ApplicationInsights,
+            // but not a dependency kind. So the dependency kind here is still just HTTP.
+            remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
+            remoteDependency.dependencyTypeName = ContractsModule.Contracts.DependencyKind
+                [ContractsModule.Contracts.DependencyKind.ApplicationInsights];
+            remoteDependency.target = urlObject.hostname + " | " + this.targetIKeyHash;
+        } else {
+            remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
+            remoteDependency.dependencyTypeName =
+                ContractsModule.Contracts.DependencyKind[ContractsModule.Contracts.DependencyKind.Http];
+            remoteDependency.target = urlObject.hostname;
+        }
+
         remoteDependency.name = dependencyName;
         remoteDependency.commandName = this.url;
         remoteDependency.value = this.duration;
         remoteDependency.success = this._isSuccess();
         remoteDependency.async = true;
-        remoteDependency.dependencyTypeName =
-            ContractsModule.Contracts.DependencyKind[ContractsModule.Contracts.DependencyKind.Http];
-        remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
         remoteDependency.dependencySource = ContractsModule.Contracts.DependencySourceType.Undefined;
         remoteDependency.properties = this.properties || {};
 
--- a/AutoCollection/ClientRequestParser.ts
+++ b/AutoCollection/ClientRequestParser.ts
@@ -53,18 +53,14 @@ class ClientRequestParser extends RequestParser {
         let dependencyName = this.method.toUpperCase() + " " + url.format(urlObject);
 
         let remoteDependency = new ContractsModule.Contracts.RemoteDependencyData();
+        remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
 
         if (this.targetIKeyHash) {
-            // Currently the service supports a dependency type name of ApplicationInsights,
-            // but not a dependency kind. So the dependency kind here is still just HTTP.
-            remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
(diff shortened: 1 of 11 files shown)


AutoCollection/ClientRequestParser.ts
```diff
@@ -48,15 +53,26 @@ class ClientRequestParser extends RequestParser {
         let dependencyName = this.method.toUpperCase() + " " + url.format(urlObject);
 
         let remoteDependency = new ContractsModule.Contracts.RemoteDependencyData();
-        remoteDependency.target = urlObject.hostname;
+
+        if (this.targetIKeyHash) {
+            // Currently the service supports a dependency type name of ApplicationInsights,
+            // but not a dependency kind. So the dependency kind here is still just HTTP.
+            remoteDependency.dependencyKind = ContractsModule.Contracts.DependencyKind.Http;
+            remoteDependency.dependencyTypeName = ContractsModule.Contracts.DependencyKind
```
nizarq: We should not be setting dependencyKind and dependencyTypeName. Both of them are deprecated. We should instead use the field 'type'

AutoCollection/ClientRequestParser.ts
```diff
@@ -48,15 +53,26 @@ class ClientRequestParser extends RequestParser {
...
+                [ContractsModule.Contracts.DependencyKind.ApplicationInsights];
```
nizarq: Instead of the string value 'ApplicationInsights', we have updated it to 'Application Insights'. This is more in line with the rest of the known types like "Azure Blobs" etc. 
See: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/nizarq/Src/DependencyCollector/Shared/Implementation/RemoteDependencyConstants.cs

jasongin: The space is unfortunate, because it means the allowed values can no longer be represented as a TypeScript enum. So we'll have to deprecate the whole DependencyKind enum.

And then will that be a breaking change to the [`trackDependency`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Client.ts#L148) API?

AutoCollection/ClientRequestParser.ts
```diff
@@ -36,6 +39,8 @@ class ClientRequestParser extends RequestParser {
      */
     public onResponse(response: http.ClientResponse, properties?: { [key: string]: string }) {
         this._setStatus(response.statusCode, undefined, properties);
+        this.targetIKeyHash =
+            response.headers && response.headers[RequestResponseHeaders.targetInstrumentationKeyHeader];
```
nizarq: Will the result of the expression with '&&' operator be string? Sorry not very fluent in node, so just doing due diligence.

jasongin: Yes. This is a super-common JavaScript shorthand, sort of an alternative to C#'s `?.` operator. If the left side is truthy, then the result of the expression is the value of the right side. To put it another way:

`A && B` is logically equivalent to `(A ? B : A)`
`A || B` is logically equivalent to `(A ? A : B)`

AutoCollection/ServerRequests.ts
```diff
@@ -93,18 +101,25 @@ class AutoCollectServerRequests {
             return;
         }
 
+        // Add the target ikey hash to the response headers, if not already provided.
+        if (client.config && client.config.instrumentationKeyHash &&
+            response.setHeader && !response.getHeader(RequestResponseHeaders.targetInstrumentationKeyHeader)) {
+                response.setHeader(RequestResponseHeaders.targetInstrumentationKeyHeader,
+                    client.config.instrumentationKeyHash);
+        }
```
nizarq: niggle: Exact same code as above. Worth making a private function to avoid duplication?

Library/RequestResponseHeaders.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * Standard parent Id header.
+     */
+    standardParentIdHeader: "x-ms-request-id",
+
+    /**
+     * Standard root id header.
+     */
+    standardRootIdHeader: "x-ms-request-root-id",
```
jasongin: I'm not, yet. I copied/translated all these constants from the C# code. I assumed these additional constants might be used later when we add per-request correlation.

Library/Contracts.ts
```diff
@@ -10,6 +10,7 @@ export module Contracts {
         SQL = 0,
         Http = 1,
         Other = 2,
+        ApplicationInsights = 3,
```
nizarq: To add to what I mentioned above. This class can be removed completely. We shouldn't depend on this anymore.

jasongin: Isn't this part of the public API, that is defining the valid values for the `dependencyKind` parameter to [`trackDependency`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Client.ts#L148)?

AutoCollection/ClientRequestParser.ts
nizarq: Actually no. Per the latest schema DependencyKind is deprecated. We shouldn't be sending enumerated values anymore. We should be using the type field which takes a string value.

AutoCollection/ClientRequestParser.ts
nizarq: got it thanks!

jasongin:
Oops, one of the tests is failing in the CI, due to the last change I made related to DependencyKind. I'll have to change usage of DependencyKind anyway based on comments, so I'll make sure the test is fixed then.

Library/RequestResponseHeaders.ts
jasongin: OK

Library/Contracts.ts
nizarq: Hmmm ... yes ... I think we will need some help from the AI team here. 
For .net it used to be the same. But the public API has since been updated. I am not sure how they handle breaking the public API. 

Here's is the current state of the .net API for reference: 
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/Core/Managed/Shared/TelemetryClient.cs#L262-L305

Anyhow, we should raise the discrepancy with them. If you don't want to get into removing the 'DependencyKind' contract. I suggest for the new 'Application Insights' dependency type, send 
DependencyKind = Http and DependencyTypeName= 'Application Insights'. 

The ingestion code is well equipped to ignore the Kind attribute and use the DependencyTypeName instead.

jasongin:
I pushed a commit that addresses all feedback above.

I have not removed the `DepdencyKind` enumeration yet, pending guidance about how that should affect the SDK's `trackDependency` API.

AutoCollection/ClientRequests.ts
```diff
@@ -1,15 +1,18 @@
 ///<reference path="..\Declarations\node\node.d.ts" />
 
 import http = require("http");
+import https = require("http");
```
kamilszostak: Is the same module imported with two different names on purpose?

AutoCollection/ServerRequests.ts
```diff
@@ -93,24 +96,38 @@ class AutoCollectServerRequests {
             return;
         }
 
+        AutoCollectServerRequests.addResponseIKeyHeader(client, response);
```
kamilszostak: Why is this called twice (lines 99 and 82)?

AutoCollection/ClientRequests.ts
jasongin: No, that's a mistake. I'm not sure why I didn't catch it in testing; the https module is only used here on node versions < 0.12, but I specifically tested for that. I'll investigate.

AutoCollection/ServerRequests.ts
jasongin: `trackRequest` and `trackRequestSync` are two methods for doing the same thing.

AutoCollection/ClientRequests.ts
jasongin: Fixed in another commit.

PR #132, opened 2016-11-04
remove IDE config files

joshgav:
The IDE config files are unnecessary. This PR removes them and adds them to the local .gitignore for good measure.


kamilszostak:
:shipit:

PR #146, opened 2016-12-07
Fix bug that causes custom context tag settings to be overwritten

lmillott:
Currently if you set a custom user Id or session Id it is overwritten by ServerRequestParser.ts. If there is already a setting for that context key, it should not be overwritten.

Additionally fixes an issue caused by the recent change to set the ikeyhash as a header on the response. It throws if the response has already been sent. Bug was introduced in this changeset: https://github.com/Microsoft/ApplicationInsights-node.js/commit/0a63c3fd3a959c88ba6751d53e633a14fd769f9a

Also fixed a couple build errors on windows using ts 2.0.3


Tests/AutoCollection/ServerRequestParser.tests.ts
```diff
@@ -95,4 +95,40 @@ describe("AutoCollection/ServerRequestParser", () => {
             assert.equal(requestData.baseData.url, "http://bing.com/");
         });
     });
+// newTags[ServerRequestParser.keys.locationIp] = tags[ServerRequestParser.keys.locationIp] || this._getIp();
```
kamilszostak: remove commented out code

Library/Util.ts
```diff
@@ -13,7 +13,7 @@ class Util {
             var cookieName = name + "=";
             var cookies = cookie.split(";");
             for (var i = 0; i < cookies.length; i++) {
-                var cookie = cookies[i];
+                var cookie: any = cookies[i];
```
kamilszostak: why do you need to cast it to `any`?

Tests/AutoCollection/ServerRequestParser.tests.ts
lmillott: thanks

Library/Util.ts
lmillott: it's causing a build error for me. Probably different typescript version. I've removed it

AutoCollection/RequestParser.ts
```diff
@@ -30,7 +30,7 @@ abstract class RequestParser {
             } else if (error instanceof Error) {
                 properties["error"] = error.message;
             } else if (typeof error === "object") {
-                for (var key in error) {
+                for (var key in <any>error) {
```
lmillott: it causes a build error for me. Probably different typescript version. I've removed it

PR #150, opened 2016-12-12
Release v0.17.1 (merge to master)

kamilszostak:


typings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "globalDependencies": {
+    "mocha": "registry:dt/mocha#2.2.5+20161028141524",
```
dnduffy: The version numbers here don't match those in the file above (e.g. mocha 3.1.2 vs. 2.2.5 and sinon 1.17.6 vs. 1.16.0...)

kamilszostak: Yes, they don't have type definitions for the latest versions of those libraries. I'll update typings when new versions are available. 

`typings info dt~sinon --versions` -> version 1.16.1
`typings info dt~mocha --versions` -> version 2.2.5

PR #156, opened 2016-12-13
Do not add correlation headers for excluded domains

### LINKED ISSUE #143: azure-storage and applicationinsights aren't playing nicely
AppInsights is conflicting with my azure-storage calls. After setting up fiddler, I found appinsights is adding this header to my http requests:
**x-ms-request-source-ikey**

that causes the request to return with this error: 

> HTTP/1.1 403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.

If I remove that header and re-play the same request, it succeeds.

Trivial repro:
```js
// (package.json/Dependencies)
{
    "applicationinsights": "^0.17.0",
    "azure-storage": "^1.3.2"
}
```
```js
// After setting proper env-vars...
var azure = require('azure-storage')
require('applicationinsights').setup().start()

azure.createBlobService().createContainerIfNotExists("somecontainer", err => console.error(err))
```
Comment out the 2nd line and the code magically works

kamilszostak:
Fixes #143


kamilszostak:
@jasongin could you review?

Library/Util.ts
```diff
@@ -151,5 +154,29 @@ class Util {
...
+     * and if it is safe to add correlation headers (x-ms-request-source-ikey, x-ms-request-target-ikey)
+     */
+    public static canIncludeCorrelationHeader(client: Client, options: http.RequestOptions | http.ServerResponse) {
+        let excludedDomains = client && client.config && client.config.correlationHeaderExcludedDomains;
+        if (!excludedDomains || excludedDomains.length == 0 || !options) {
+            return true;
+        }
+
+        let url = (<http.RequestOptions>options).host ||
+                ((<http.ServerResponse>options).getHeader && (<http.ServerResponse>options).getHeader("host"));
```
kamilszostak: It's to disable the `x-ms-request-target-ikey` header - [see](https://github.com/Microsoft/ApplicationInsights-node.js/pull/156/files#diff-641655496f5e68858e3a1056f5072f06). I was wondering if I should only disabled `x-ms-request-source-ikey` header, but I've decided to disable all correlation headers for excluded domains.

AutoCollection/ClientRequests.ts
```diff
@@ -73,8 +73,8 @@ class AutoCollectClientRequests {
      * Tracks an outgoing request. Because it may set headers this method must be called before
      * writing content to or ending the request.
      */
-    public static trackRequest(client: Client, requestOptions: any, request: http.ClientRequest,
-            properties?: { [key: string]: string }) {
+    public static trackRequest(client: Client, requestOptions: http.RequestOptions, request: http.ClientRequest,
```
jasongin: The `requestOptions` parameter may actually be a string. (See the Node.js documentation for `http.request()`.) And that is very common usage.

So that means if it is a string you'll need to parse it in `canIncludeCorrelationHeader`, since that code currently assumes it is an object.

Library/Util.ts
jasongin: I see where it is called from, but I don't see how it's going to work because there is no such `"host"` header for a server response as far as I know. (Or if there is, I guess it would refer to the server host, which doesn't make sense to filter on since it is always the same.)

AutoCollection/ClientRequests.ts
jasongin: Consider adding a host property to the `ClientRequestParser` (and move the call to it up before line 88) so that you can re-use its request options parsing instead of duplicating it. Or something like that.

Library/Util.ts
```diff
@@ -151,5 +154,29 @@ class Util {
...
+
+        for (let i = 0; i < excludedDomains.length; i++) {
+            let matches = url ? url.match(excludedDomains[i]) : null;
```
jasongin: I'm concerned that it's very easy to add a string to the `config.correlationHeaderExcludedDomains` array. (Unit tests actually do that.) Then a string gets interpreted directly as a regular expression, which is not necessarily what would be expected, for example a period in the string will match *any* character.

I'd suggest either (or a combination) of the following approaches:
 1. Check if the exclusion item is an instance of a `RegExp`, if not just do a plain string `indexOf` test. 
 2. Use glob matching instead of regular expressions, e.g. `"*.blob.core.windows.net"`. To avoid adding a dependency to a separate glob package, you can implement very simple glob syntax like: `new RegExp(str.replace(/\*/g, '.*')`

Library/Client.ts
```diff
@@ -142,7 +142,7 @@ class Client {
         ServerRequestTracking.trackRequest(this, request, response, properties);
     }
 
-    public trackDependencyRequest(requestOptions: any, request: http.ClientRequest, properties?: { [key: string]: string; }) {
+    public trackDependencyRequest(requestOptions: http.RequestOptions, request: http.ClientRequest, properties?: { [key: string]: string; }) {
```
jasongin: To be technically correct, I think this needs to be `string | http.RequestOptions | https.RequestOptions`. And same in `ClientRequests.trackRequest()`.

The currently checked-in typings define separate `RequestOptions` classes [for `http`](https://github.com/Microsoft/ApplicationInsights-node.js) and [for `https`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/typings/globals/node/index.d.ts#L567). While they [don't allow the `http.request()` options parameter to be a string](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/typings/globals/node/index.d.ts#L425), I consider that a bug in the typings since that API is documented to allow a string and is widely used that way.

AutoCollection/ClientRequests.ts
kamilszostak: I refactored the code to use request parser to get the URL.

Library/Client.ts
kamilszostak: Done.

Library/Util.ts
```diff
@@ -151,5 +154,27 @@ class Util {
...
+     * and if it is safe to add correlation headers (x-ms-request-source-ikey, x-ms-request-target-ikey)
+     */
+    public static canIncludeCorrelationHeader(client: Client, url: string) {
+        let excludedDomains = client && client.config && client.config.correlationHeaderExcludedDomains;
+        if (!excludedDomains || excludedDomains.length == 0 || !url) {
+            return true;
+        }
+
+        for (let i = 0; i < excludedDomains.length; i++) {
+            let regex = new RegExp(excludedDomains[i].replace(/\./g,"\.").replace(/\*/g,".*"));
```
jasongin: Is it OK that this can match more than just the host portion of the URL? For example, if the url's query parameters includes another url, that could be matched here.

Library/Util.ts
```diff
@@ -151,5 +154,27 @@ class Util {
...
+            let matches = url.match(regex);
```
jasongin: It would be a little more efficient to use the `RegExp.test()` method here, since that just returns a boolean result instead of creating an array of match objects.

Library/Util.ts
jasongin: To ask another way, should this method parse out the host portion of the URL before matching?

jasongin:
This is almost ready; I just added a couple comments above about regex usage.

Does anyone know why the CI is failing? I can't repro the error there:
```
Library/Contracts.ts(264,27): error TS1056: Accessors are only available when targeting ECMAScript 5 and higher.
```

Library/Util.ts
kamilszostak: Good point, I'll check only the hostname. I updated unit tests to cover this scenario.

kamilszostak:
I think this #164 will fix the CI build error.

SergeyKanzhelev:
@kamilszostak , yes sorry for breaking it. I just merged

jasongin:
LGTM

giza182:
Is there an way to extend the list of excluded domains without touching the source code? I have to exclude the azure queue and I see no obvious way to exclude that domain without changing the source code of the library.

sellomkantjwa:
Nvm. 
``` javascript
appInsights.setup(config.azureInstrumentationKey);
appInsights.client.config.correlationHeaderExcludedDomains.push("*.queue.core.windows.")
appInsights.start();
```
Works on 0.19.0. 
Does not work on 0.18.0 because there was a bug that only looked at the first excluded domain on the list in that version.

PR #160, opened 2016-12-30
read transaction root and parent ID

SergeyKanzhelev:
Are there any end-to-end tests I can modify to test it better?


kamilszostak:
@SergeyKanzhelev end to end tests are in [EndtoEnd.tests.ts](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Tests/EndToEnd.tests.ts). They are end-to-end from the SDK perspective, and they use a fake server to mock AI backend.

PR #187, opened 2017-02-09
Add automatic correlation between dependencies and requests

OsvaldoRosado:
This adds a "Correlation Context" shared between all code handling a given request (using the Zone.js library). This correlation context is used to share an operation id with requests and all dependencies. It is off by default but can be disabled.

This is best effort. A "wrap" method is supplied for end-users to manually fix any issues with automatic correlation if necessary. Even if automatic correlation fails, dependencies will still be recorded. These dependencies will either be missing a request association or will be associated to a different request.

A portion of the correlation context is provided for end user use and is passed in to the context object given to telemetry processors.

Automatic correlation is supported in Node > 4.0 or io.js > 3.3. Our unit tests indicate errors from timer patching in versions less than this.

### DIFF
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -1,0 +1,94 @@
+import http = require("http");
+
+interface CorrelationContext {
+    id: string;
+    req: http.ServerRequest;
+    res: http.ServerResponse;
+    metadata?: {};
+}
+
+class CorrelationContextManager {
+    private contexts: {[uid: number]: CorrelationContext} = [];
+    private currentContext: CorrelationContext = null;
+    private enabled: boolean = false;
+    private hasEverEnabled: boolean = false;
+
+    /** 
+     *  Provides the current Context.
+     *  The context is the most recent one entered into for the current
+     *  logical chain of execution, including across asynchronous calls.
+     */
+    public getCurrentContext(): CorrelationContext {
+        if (!this.enabled) {
+            return null;
+        }
+        return this.currentContext;
+    }
+
+    /** 
+     *  Enters a new Context.
+     *  All logical children of the execution path that entered this Context
+     *  will receive this Context object on calls to GetCurrentContext.
+     */
+    public enterNewContext(context: CorrelationContext) {
+        this.currentContext = context;
+    }
+
+    /**
+     *  Enables the CorrelationContextManager. This uses Node's async_wrap
+     *  API which is not available in older versions of Node (< 0.12.1).
+     *  This method will detect these old versions and do nothing.
+     */
+    public enable() {
+        if (/^0\.([0-9]\.)|(10\.)|(11\.)|(12\.0$)/.test(process.versions.node)) {
+            return;
+        }
+
+        this.enabled = true;
+        
+        if (!this.hasEverEnabled) {
+            this.hasEverEnabled = true;
+
+            // Suppress unstable warning from async-hook
+            process.env["NODE_ASYNC_HOOK_NO_WARNING"] = true;
+
+            // Load in async-hook and enable it
+            var asyncHook = require("async-hook");
+            asyncHook.addHooks({
+                init: this.onAsyncInit,
+                pre: this.onAsyncPre,
+                post: this.onAsyncPost,
+                destroy: this.onAsyncDestroy});
+            asyncHook.enable();
+        }
+    }
+
+    public disable() {
+        this.enabled = false;
+    }
+
+    public isEnabled() {
+        return this.enabled;
+    }
+
+
+    // Callbacks for async-hook
+    private onAsyncInit(uid: number) {
+        this.contexts[uid] = this.currentContext;
+    }
+
+    private onAsyncPre(uid: number) {
+        this.currentContext = this.contexts[uid];
+    }
+
+    private onAsyncPost(uid: number) {
+        this.currentContext = null;
+    }
+
+    private onAsyncDestroy(uid: number) {
+        delete this.contexts[uid];
+    }
+}
+
+// Force singleton
+export = new CorrelationContextManager();
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -2,8 +2,6 @@ import http = require("http");
 
 interface CorrelationContext {
     operationId: string;
-    req: http.ServerRequest;
-    res: http.ServerResponse;
 }
 
 class CorrelationContextManager {
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -1,8 +1,4 @@
 /// <reference path="..\node_modules\zone.js\dist\zone.js.d.ts" />
-var nodeVer = process.versions.node.split(".");
-if (parseInt(nodeVer[0]) > 3 || (parseInt(nodeVer[0]) > 2 && parseInt(nodeVer[1]) > 2)) { // Unit tests warn of errors < 3.3 from timer patching. All versions before 4 were 0.x
-    require("zone.js"); // Keep this first
-}
 import http = require("http");
 import Util = require("../Library/Util");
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -64,13 +65,16 @@ client.trackTrace("trace message");
 ### Telemetry Processor
 
 ```javascript
-public addTelemetryProcessor(telemetryProcessor: (envelope: ContractsModule.Contracts.Envelope) => boolean)
+public addTelemetryProcessor(telemetryProcessor: (envelope: ContractsModule.Contracts.Envelope, context: { http.RequestOptions, http.ClientRequest, http.ClientResponse, userProperties }) => boolean)
```
OsvaldoRosado: The original context variable for telemetry processors was added by @SergeyKanzhelev in a recent PR. I think it might be cleaner to not pass the correlation context as userProperties if automatic correlation is enabled, but rather as the entire object (removing these http.Request related fields and leaving it to the user to add these to the correlation context if they're interested in them). These http related fields could be provided as the context if automatic correlation is disabled.

Thoughts?

SergeyKanzhelev: That context was passed to telemetry processor so customer can extract certain values from http request and associate them with the telemetry collected by default. 

Like if you want to get the header value and add it as a custom property to the `RequestTelemetry`. 

So I'd suggest to revert it back and extend with other context if needed

AutoCollection/ServerRequestParser.ts
```diff
@@ -91,11 +91,15 @@ class ServerRequestParser extends RequestParser {
...
         newTags[ServerRequestParser.keys.operationName] = tags[ServerRequestParser.keys.operationName] || this.method + " " + url.parse(this.url).pathname;
         newTags[ServerRequestParser.keys.operationParentId] = tags[ServerRequestParser.keys.operationParentId] || this.parentId;
-        newTags[ServerRequestParser.keys.operationId] = tags[ServerRequestParser.keys.operationId] || this.operationId;
+        newTags[ServerRequestParser.keys.operationId] = this.getOperationId(tags);
 
         return newTags;
     }
 
+    public getOperationId(tags:{[key: string]:string}) {
+        return tags[ServerRequestParser.keys.operationId] || this.operationId;
```
SergeyKanzhelev: You'd need to pass operation ID and request ID. Every telemetry should have `.Context.Operation.Id = global ID` and `.Context.Operation.ParentId = current request ID`

AutoCollection/ServerRequests.ts
```diff
@@ -18,6 +19,7 @@ class AutoCollectServerRequests {
     private _client:Client;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
+    private _isAutoCorrelating: boolean;
```
SergeyKanzhelev: What's the default value? Did you try it on Breeze? How fast is it?

package.json
```diff
@@ -48,7 +48,10 @@
     "mocha": "3.1.2",
     "node-mocks-http": "1.2.3",
     "sinon": "1.17.6",
-    "typescript": "1.8.9",
+    "typescript": "2.0.10",
```
SergeyKanzhelev: cool!

AutoCollection/ServerRequests.ts
OsvaldoRosado: Its on by default. Works correctly on Breeze. I don't have numbers for perf hit (you would need quite a bit of synthetic traffic to really measure I would presume). We can try in INT.

Library/Client.ts
```diff
@@ -277,8 +278,12 @@ class Client {
 
         if (telemetryProcessorsCount === 0) {
             return accepted;
-        }
+        }     
 
+        var correlationContext = CorrelationContextManager.getCurrentContext();
+        var userCorrelationContext = correlationContext && correlationContext.userProperties;
+        contextObjects = contextObjects || {};
+        contextObjects['userProperties'] = userCorrelationContext;
```
KamilSzostak: Correlation context has two fields: 'operationId' and 'userProperties'. `userCorrelationContext` can be either a whole `correlationContext` object or just `userProperties` (like 284). If you assign it to `userProperties' here, wouldn't it duplicate the `operationId'?

OsvaldoRosado:
Note: I've updated this to be off by default (following some offline discussion). After testing of it on in some of our own services, we can release a new version with it on by default

AutoCollection/ServerRequestParser.ts
OsvaldoRosado: I've updated the PR to address this

package.json
```diff
@@ -48,7 +48,10 @@
...
     "typings": "2.0.0"
+  },
+  "dependencies": {
+    "zone.js": "0.7.6"
```
SergeyKanzhelev: have you cleared this from license perspective?

debug.log
```diff
@@ -0,0 +1,2 @@
+[0203/150119:ERROR:tcp_listen_socket.cc(76)] Could not bind socket to 127.0.0.1:6004
```
SergeyKanzhelev: remove this file

AutoCollection/CorrelationContextManager.ts
```diff
@@ -0,0 +1,152 @@
...
+        var req = null;
+        try {
+            req = require(module);
+        } catch (e) {
+            return null;
+        }
+        return req;
+    }
+
+    private static patchRedis() {
```
SergeyKanzhelev: write comment on what you are doing here. Also it would be great to share the docs (like `README.md` or API reference) on how other third party libraries can be patched if needed.

AutoCollection/ServerRequestParser.ts
```diff
@@ -89,13 +91,29 @@ class ServerRequestParser extends RequestParser {
...
+    }
+
+    public getOperationParentId(tags:{[key: string]:string}) {
+        return tags[ServerRequestParser.keys.operationParentId] || this.parentId || this.getOperationId(tags);
+    }
+
+    public getOperationName(tags:{[key: string]:string}) {
+        return tags[ServerRequestParser.keys.operationName] || this.method + " " + url.parse(this.url).pathname;
```
SergeyKanzhelev: was this ` url.parse(this.url).pathname` already parsed out of the requestTelemetry? Maybe just use `requestData.name`?

package.json
OsvaldoRosado: Yep!

debug.log
OsvaldoRosado: Good catch!

AutoCollection/ServerRequestParser.ts
OsvaldoRosado: `requestData.name` isn't available until getRequestData() is run. Would you prefer this moved to the constructor and replaced with class variables?

SergeyKanzhelev: It may be a separate improvement to allow customers to override this value.

SergeyKanzhelev:
cc: @lmolkova FYI in case you have any feedback

AutoCollection/ServerRequestParser.ts
OsvaldoRosado: Customers today can provide a tag override (tags will prevail over any precalculated value) or change it in a telemetry processor.

AutoCollection/CorrelationContextManager.ts
```diff
@@ -149,4 +158,68 @@ export class CorrelationContextManager {
         }
     }
 
+    // Zone.js breaks concatenation of timer return values.
+    // This fixes that.
+    private static patchTimers(methodNames: string[]) {
```
KamilSzostak: Consider sending this fix to zone.js repository.

AutoCollection/CorrelationContextManager.ts
```diff
@@ -0,0 +1,225 @@
...
+                }
+
+                return orig.apply(this, args);
+            };
+        }
+    }
+
+    // Zone.js breaks concatenation of timer return values.
+    // This fixes that.
+    private static patchTimers(methodNames: string[]) {
```
KamilSzostak: Consider sending this patch to zone.js repository.

Issue #182, opened 2017-02-01
properties with number 0 as value are not being sent to the server correctly


anliakho:
The bug is  https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Util.ts#L143 where 
`if (property && typeof property.toString === "function") `
should instead read 
`if (property !== undefined && typeof property.toString === "function") {`
As it is the value 0 results in  "invalid property type: number" being sent up

PR #191, opened 2017-02-14
Fix validateStringMap function

KamilSzostak:
Fix for #182. 

With this change, the `validateStringMap` function will correctly process properties with number 0 as value. It will also drop invalid values, instead of changing them to `invalid property type: ...`.

### DIFF
--- a/Library/Util.ts
+++ b/Library/Util.ts
@@ -140,10 +140,11 @@ class Util {
                 var property = obj[field];
                 var propertyType = typeof property;
                 if (propertyType !== "string") {
-                    if (property && typeof property.toString === "function") {
+                    if (property != null && typeof property.toString === "function") {
                         property = property.toString();
                     } else {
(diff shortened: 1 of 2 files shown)


Library/Util.ts
```diff
@@ -140,10 +140,11 @@ class Util {
                 var property = obj[field];
                 var propertyType = typeof property;
                 if (propertyType !== "string") {
-                    if (property && typeof property.toString === "function") {
+                    if (property != null && typeof property.toString === "function") {
```
Dmitry-Matveev: [TypeScript wiki](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) seems to want us use undefined.
Was it changed recently or we only care about null here?

KamilSzostak: `p != null` is the same as `p !== null && p !== undefined`. I wanted to make sure that it's safe to call  `toString`.

PR #194, opened 2017-03-01
Fixed correlationHeaderExcludedDomains only cares about first value

thomasacrosby:
currently unable to exclude multiple domains in correlationHeader check


KamilSzostak:
@thomasacrosby thanks for the contribution!

PR #195, opened 2017-03-03
Copying all descriptors of error properties

MSLaguana:
Without this change, attempting to enable async context tracking before `require("morgan")` would result in errors when `depd` attempted to set some properties on `Error` which were implicitly non-writable.

With this change, the methods that are just copied across will have exactly the same descriptor as the original method, so the replacement should be transparent.


KamilSzostak:
@OsvaldoRosado can you help review this PR?

PR #198, opened 2017-03-06
Restrict customProperties to kv pair of strings and warnings discouraging sensitive info storage

OsvaldoRosado:
These changes will allow us to more easily implement compatibility with https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md#http-protocol-proposal


OsvaldoRosado:
Adding @lmolkova for FYI

PR #199, opened 2017-03-06
Remove obsolete fields from envelope

### LINKED ISSUE #189: Remove os and osVer fields from envelope
`os` and `osVer` fields are not part of the envelope in the public schema:

https://github.com/Microsoft/ApplicationInsights-node.js/blob/c3241540e6ae0129fbf448c484f932fd241cd78e/Library/Client.ts#L222-L224

We should remove collection of those. However before dropping them - check with VSCode team to make sure they are not relying on these field. So they wouldn't be surprised.

KamilSzostak:
Fix #189. 

The latest `Envelope` scheme is [here](https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/Envelope.bond).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,5 @@
         "**/*.js.map": true,
         "**/*.d.ts": true
      },
-     "typescript.tsdk": "./node_modules/typescript/lib"
+    "typescript.tsdk": "./node_modules/typescript/lib"
 }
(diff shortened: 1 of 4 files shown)


SergeyKanzhelev:
Have you checked with VSCode that they do not rely on those fields?

Issue #209, opened 2017-03-21
Track* methods should accept tagOverrides and contextObjects


christopheranderson:
Using the custom `track*` methods from a Node.js application with lots of async code can be a headache since I currently have to set the operation id on the client context which might be called out of order. I can use the  `track` method to do this per track call, but then I lose the easy of use of the other `track*` commands. 

I'm working on a PR for this.

KamilSzostak:
👍

PR #210, opened 2017-03-21
Add tagOverrides and contextObjects

christopheranderson:
Addresses #209 

Also updated documentation.

Didn't add any additional tests since this was a lightweight addition and it didn't look like we had any specific tests for the existing `track` method.

### DIFF
--- a/Library/Client.ts
+++ b/Library/Client.ts
@@ -41,10 +41,11 @@ class Client {
 
     /**
      * Log a user action or other occurrence.
-     * @param   name    A string to identify this event in the portal.
-     * @param   properties  map[string, string] - additional data used to filter events and metrics in the portal. Defaults to empty.
-     * @param   measurements    map[string, number] - metrics associated with this event, displayed in Metrics Explorer on the portal. Defaults to empty.
-     * @param   tagOverrides the context tags to use for this telemetry which overwrite default context values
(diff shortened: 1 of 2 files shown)


Library/Client.ts
```diff
@@ -44,8 +44,11 @@ class Client {
      * @param   name    A string to identify this event in the portal.
      * @param   properties  map[string, string] - additional data used to filter events and metrics in the portal. Defaults to empty.
      * @param   measurements    map[string, number] - metrics associated with this event, displayed in Metrics Explorer on the portal. Defaults to empty.
+     * @param   tagOverrides the context tags to use for this telemetry which overwrite default context values
```
OsvaldoRosado: This should probably have a listing for contextObjects as well

christopheranderson: It probably should. Any thoughts on the right language there? I didn't find any good descriptions of what that contextObjects collection does, though I can take a crack at it, based on my understanding.

OsvaldoRosado: Perhaps something like "An event-specific context that will be passed to telemetry processors handling this event before it is sent. For a context spanning your entire operation, consider appInsights.getCorrelationContext"

christopheranderson: Done

Issue #61, opened 2016-01-07
Sampling


nickdima:
Is it possibile to enable sampling?

kamilszostak:
Currently Node.js SDK doesn't support sampling. We are looking into adding sampling when auto-collection is enabled, but we don't have a timeline for it yet. 

When auto-collection is disabled it should be easy to add some wrapper for `track*` methods which will do the sampling.

PR #217, opened 2017-03-30
Add sampling support

OsvaldoRosado:
Addresses #61 

This also places some of the groundwork for refactoring SDK functions into isolated telemetry processors


TelemetryProcessors/SamplingTelemetryProcessor.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export function samplingTelemetryProcessor(envelope: ContractsModule.Contracts.Envelope, contextObjects: { correlationContext: CorrelationContext }): boolean {
+    var samplingPercentage = envelope.sampleRate; // Set for us in Client.getEnvelope
+    var isSampledIn = false;
+
+    if (samplingPercentage === null || samplingPercentage === undefined || samplingPercentage >= 100) {
+        return true;
+    } else if (contextObjects.correlationContext && contextObjects.correlationContext.operation) {
+        // If we're using dependency correlation, sampling should retain all telemetry from a given request 
+        isSampledIn = getSamplingHashCode(contextObjects.correlationContext.operation.id) < samplingPercentage;
```
SergeyKanzhelev: in .NET we use User_ID if available and if it is not available - we fall back to operation ID. Will it be hard to replicate that logic?

TelemetryProcessors/SamplingTelemetryProcessor.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ *  A telemetry processor that handles sampling.
+ */
+export function samplingTelemetryProcessor(envelope: ContractsModule.Contracts.Envelope, contextObjects: { correlationContext: CorrelationContext }): boolean {
+    var samplingPercentage = envelope.sampleRate; // Set for us in Client.getEnvelope
+    var isSampledIn = false;
+
+    if (samplingPercentage === null || samplingPercentage === undefined || samplingPercentage >= 100) {
+        return true;
```
SergeyKanzhelev: I do not understand. If it's `null` we want to calculate it, correct?

SergeyKanzhelev: We want to skip if it was already calculated so you can chain multiple sampling processors

TelemetryProcessors/SamplingTelemetryProcessor.ts
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+        // If we're not using dependency correlation, sampling should use a random distribution on each item
+        isSampledIn = (Math.random() * 100) < samplingPercentage;
+    }
+
+    return isSampledIn;
+}
+
+/** Ported from AI .NET SDK */
+export function getSamplingHashCode(input: string): number {
```
SergeyKanzhelev: Why not use this class [HashCodeScoreGenerator ](https://github.com/Microsoft/ApplicationInsights-JS/blob/178b7d66d8d7a741f9154012f8d19c3fbc52d503/JavaScript/JavaScriptSDK/HashCodeScoreGenerator.ts#L2) and [SamplingScoreGenerator ](https://github.com/Microsoft/ApplicationInsights-JS/blob/178b7d66d8d7a741f9154012f8d19c3fbc52d503/JavaScript/JavaScriptSDK/SamplingScoreGenerator.ts#L5)

SergeyKanzhelev: More and more I feel we need to share the npm module with "Base" SDK for JavaScript that will define interface like trackTrace as well as telemetryprocessor interface and such

TelemetryProcessors/SamplingTelemetryProcessor.ts
OsvaldoRosado: The envelope will be prepopulated with the default level of sampling before any telemetry processors. Other telemetry processors can change the envelope sample rate field at will. This telemetry processor will always run last, reading the sampleRate field off of the envelope and deciding to keep or reject the data based off of it

TelemetryProcessors/SamplingTelemetryProcessor.ts
OsvaldoRosado: I don't believe we currently keep user id within the correlation context, so that will work for requests but no dependencies. We could update the correlation logic but that's probably a separate PR

TelemetryProcessors/SamplingTelemetryProcessor.ts
OsvaldoRosado: It's not used for basically historical reasons :) I had already written this together for Breeze and simply moved it across. When I wrote it the first time I didn't know JS had hashing logic at all. I believe they're basically the same however. A base SDK would seem useful for this...

MSLaguana:
Is it possible to query the SDK to work out whether an event would be filtered? I'm thinking of a scenario where an application would have to perform extra work to construct a telemetry event (getting appropriate data for it) that could be avoided if we know that the event will be ignored anyway.

OsvaldoRosado:
@MSLaguana it's not possible to know this ahead of time because of the possible random component to sampling. It is also by design that sampling is the last action to happen to an envelope before it is sent for sending. This allows customers to add their own telemetry processors that adjust the sample rate per item. For example, if you detect an item that you know is important and rare, you have a chance to adjust it's individual samplerate to 100 before it reaches sampling so that it is kept for sending.

KamilSzostak:
Could you add a sample code with a short description to the README?

OsvaldoRosado:
@KamilSzostak Good idea. Updated!

Tests/TelemetryProcessors/SamplingTelemetryProcessor.tests.ts
```diff
@@ -0,0 +1,147 @@
...
+            var iterations = 1000;
+            var accepted = 0;
+            mockData.sampleRate = 50;
+
+            for (var i=0; i<iterations; i++) {
+                var result = Sampling.samplingTelemetryProcessor(mockData, {correlationContext: null});
+                if (result) accepted++;
+            }
+
+            assert.ok(accepted > (iterations * 0.40), "data should pass more than 40% of the time");
```
KamilSzostak: This test will fail randomly once it a ~2.72846×10^(-8) runs... I guess we can live with it :)

OsvaldoRosado: :laughing:

TelemetryProcessors/SamplingTelemetryProcessor.ts
SergeyKanzhelev: So the logic is - if telemetry processor set `samplingRate` to `undefined` - this telemetry processor will do nothing? This looks strange

OsvaldoRosado:
The sampleRate field is pre-set when the envelope is constructed. User processors can change the field as desired.

OsvaldoRosado:
I don't think we have a changelog file. Instead all PR #'s are attached to the release info on new releases.

Issue #214, opened 2017-03-26
Need to Document trackRequest and trackRequestSync


SteveALee:
These 2 methods are not documented in the readme but trackRequestSync is used in [example](https://github.com/christopheranderson/azure-functions-app-insights-sample/blob/master/item/index.js#L26)

PR #218, opened 2017-04-03
Add missing docs

OsvaldoRosado:
Addresses #214

### DIFF
--- a/Library/Client.ts
+++ b/Library/Client.ts
@@ -145,14 +145,44 @@ class Client {
         this.track(data, tagOverrides, contextObjects);
     }
 
+    /**
+     * Log an incoming request. This method will synchronously record the event, rather than waiting for response send.
+     * Use this when you need to perform custom logic to record the duration of the request.
+     * 
+     * This call will also add outgoing headers to the supplied response object for correlating telemetry across different services.
+     * @param request              http.ServerRequest - the request object to monitor
+     * @param response             http.ServerResponse - the response object to monitor
+     * @param elapsedMilliseconds  number - the elapsed time taken to handle this request in milliseconds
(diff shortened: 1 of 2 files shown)


Library/Client.ts
```diff
@@ -145,14 +145,44 @@ class Client {
...
+     * Use this when you need to perform custom logic to record the duration of the request.
+     * 
+     * This call will also add outgoing headers to the supplied response object for correlating telemetry across different services.
+     * @param request              http.ServerRequest - the request object to monitor
+     * @param response             http.ServerResponse - the response object to monitor
+     * @param elapsedMilliseconds  number - the elapsed time taken to handle this request in milliseconds
+     * @param properties           map[string, string] - additional data used for filtering in the portal. Defaults to empty.
+     * @param error                any - an object indicating the request was unsuccessful. This object will be recorded with the request telemetry.
+     */
     public trackRequestSync(request: http.ServerRequest, response: http.ServerResponse, ellapsedMilliseconds?: number, properties?: { [key: string]: string; }, error?: any) {
```
SergeyKanzhelev: Should we have `startRequest` `stopRequest` instead? `Sync` is really misleading in this context.

OsvaldoRosado: I agree, this name is confusing. I'm not sure what start/stop would gain over the non-"sync" version of this call however

KamilSzostak: I had the same comment when the `*sync` methods were added - #75. According to the author, this follows the Node.js naming convention, but I'm open to renaming them in the future.

SteveALee: Agree, To me the sync implied it waits for the response, the exact opposite of actual behaviour. But once you grok it it's OK

PR #221, opened 2017-04-06
migrate from typings to @types

joshgav:
Per https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/, we should now use @types packages instead of typings for TS 2+.

@OsvaldoRosado the @types/zone.js@0.0.27 package doesn't seem to work with the version of zone.js in use at the moment, so I kept the `/// <reference />` line at the top of `AutoCollection/CorrelationContextManager.ts`. I also had a problem with `captureStackTrace`, which is defined explicitly in recent `node.d.ts` files, and had to cast to \<any\> to workaround, please see the end of that file. Is there a better way?

Thanks!


AutoCollection/CorrelationContextManager.ts
```diff
@@ -1,4 +1,5 @@
-/// <reference path="..\node_modules\zone.js\dist\zone.js.d.ts" />
+/// <reference path="../node_modules/zone.js/dist/zone.js.d.ts" />
```
OsvaldoRosado: I think this is required not because of the version of zone but the strange way I require it without importing so that it is not loaded on incompatible systems. Maybe a tsconfig can change this?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -215,12 +216,15 @@ export class CorrelationContextManager {
...
             var propertyName = props[i];
             if (!AppInsightsAsyncCorrelatedErrorWrapper[propertyName]) {
                 Object.defineProperty(AppInsightsAsyncCorrelatedErrorWrapper, propertyName, Object.getOwnPropertyDescriptor(orig, propertyName));
             }
         }
-        global.Error = AppInsightsAsyncCorrelatedErrorWrapper;
+        
+        // explicit cast to <any> required to avoid type error for captureStackTrace
+        // with latest node.d.ts (despite workaround above)
+        global.Error = <any>AppInsightsAsyncCorrelatedErrorWrapper;
```
OsvaldoRosado: Maybe we could avoid this with the error wrapper being a class that inherits from global.Error in explicit TS rather than the manual prototype assignment used here. However, I suspect that will cause some issues however. I'm okay with the cast to any

package.json
```diff
@@ -50,7 +50,9 @@
     "node-mocks-http": "1.2.3",
     "sinon": "1.17.6",
     "typescript": "2.0.10",
-    "typings": "2.0.0"
+    "@types/mocha": "^2.2.40",
+    "@types/node": "^7.0.12",
```
MSLaguana: Do you want to use the latest typings, or the typings for the earliest node that you support? By using the latest, you might not notice if you use a newer language feature (e.g. `array.find`).

MSLaguana: In particular, the previous node typings was for node 0.10, but it looks like the earliest that the `@types/node` package has is for node 4.X

joshgav:
As you can see I forgot to run `npm test`, sorry! Have to figure out why `Tests/Library/Client.tests.ts` isn't picking up the Sinon namespace. Will pick up again soon.

package.json
joshgav: Hmmm, good point. Node < 4.x isn't supported anymore, but customers might still be using it. @beckylino can we check what versions of Node an ikey is reporting from? I couldn't find that in our metrics.

In any case makes sense to bump the @types/node version back to 4.x, done.

joshgav:
Okay tests now pass! Thanks for the reviews.

package.json
OsvaldoRosado: I don't believe we record metrics for which versions of Node are in use

OsvaldoRosado: Going down to 4.x should be fine if that's the earliest available. Our CI builds will verify things still run on the older versions

joshgav: apropos this, looks like v0 was added on march 24: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node/v0, so we can update here if we want.

might be best to drop support for <4 soon and make it easier on ourselves at this point.

PR #222, opened 2017-04-07
update README and rename LICENSE

joshgav:


OsvaldoRosado:
@joshgav Looks like this has conflicts with the addition of sampling

README.md
```diff
@@ -1,219 +1,268 @@
...
-    .start();
-```
+The appInsights object provides a number of configuration methods. They are
+listed in the following snippet with their default values.
 
-> Be sure to call `require("applicationinsights")` before your other imports. This allows the SDK to do patching necessary for tracking correlation state before other libraries use patched methods. If you encounter conflicts with other libraries doing similar patching, place this import below those libraries.
-
-### Disabling automatic collection
+To correlate events in a request across callbacks and dependencies, you'll need
+to call `.setAutoDependencyCorrelation(true)`.
```
AlexBulankou: Instead of calling setAutoDependencyCorrelatio out separately, I would include a comment for each entry below

README.md
```diff
@@ -1,219 +1,268 @@
...
 ```javascript
-import appInsights = require("applicationinsights");
+let appInsights = require("applicationinsights");
 appInsights.setup("<instrumentation_key>")
-    .setAutoCollectRequests(false)
-    .setAutoCollectPerformance(false)
-    .setAutoCollectExceptions(false)
-    .setAutoCollectDependencies(false)
-    // no telemetry will be sent until .start() is called
+    .setAutoDependencyCorrelation(false)
```
AlexBulankou: Why is false default (without explanation)

README.md
```diff
@@ -1,219 +1,268 @@
...
+    ```javascript
+    appInsights.client.commonProperties = {
+      environment: process.env.SOME_ENV_VARIABLE
+    };
+    ```
 
-    res.writeHead(200, { "Content-Type": "text/plain" });
-    res.end("Hello World\n");
-}).listen(port);
+* Manually track all HTTP GET requests
```
AlexBulankou: Probably mention that in this case you want to set `` .setAutoCollectRequests(true)``?

README.md
joshgav: @OsvaldoRosado set it up that way originally, discussion [here](https://github.com/Microsoft/ApplicationInsights-node.js/pull/235#issuecomment-299582120) on defaulting to true beginning in 0.20. I'll update this based on outcome there, in the meantime will note that this is experimental.

README.md
joshgav: Do you mean `.setAutoCollectRequests(false)`? I added that now.

joshgav:
@OsvaldoRosado @AlexBulankou updated this to address your feedback, PTAL. I'll do another pass in a few days, I think this can be merged now though if you'd like. Thanks!

KamilSzostak:
:shipit:

PR #223, opened 2017-04-07
edits to comments in applicationinsights.ts

joshgav:
Mostly edits to comments, but also renamed `DisableConsoleLogging` to `DisableVerboseLogging` to match `EnableVerboseLogging`.


applicationinsights.ts
```diff
@@ -225,9 +241,10 @@ class ApplicationInsights {
     }
 
     /**
-     * Disables verbose debug and warning logging
+     * Disables debug and warning logging for AppInsights itself.
+     * @returns {ApplicationInsights} this class
      */
-    public static disableConsoleLogging() {
+    public static disableVerboseLogging() {
```
OsvaldoRosado: Should we still provide the old method as a deprecated copy that merely calls the new one?

joshgav: yes, thanks for catching that! will update.

PR #224, opened 2017-04-08
update package install and build scripts

joshgav:
* Adds a prepare script to replace the prepublish script for `npm install` in npm@5+, see https://github.com/npm/npm/issues/10074.
* Built artifacts from tsc are aggregated and stored in ./out for easier workspace management (e.g. `npm run clean`).
* New test case added which isn't transpiled by tsc to test some scenarios. Includes a test for loading the transpiled module, since it's now in ./out.

The actual steps required to install for development and production stay the same as now documented in the README:
  * use `npm install` to install dependencies and for local usage, e.g. with `npm link`,
  * use `npm test` to install, build, and test as before

When installed in production environments, the "prepare" script is not called. We'd continue to rely on prepublish to precompile for dependants.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,9 @@
 # Ignore build files
-obj
-*.d.ts
-*.tgz
-package-version
+out/
+obj/
 
 # Except imported Declarations and Typings
 !/Declarations/**
-!/typings/**
 
 # Ignore Node files
 node_modules/**
@@ -27,6 +24,3 @@ node_modules/**
 # Ignore log files
 npm-debug.log
 
-
-*.js.map
-*.js
--- a/Tests/js/endToEnd.js
+++ b/Tests/js/endToEnd.js
@@ -1,0 +1,11 @@
+const assert = require('assert');
+
+describe('module', function () {
+    describe('#require', function () {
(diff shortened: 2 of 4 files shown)


OsvaldoRosado:
Why prepare vs prepublishOnly?

joshgav:
@OsvaldoRosado I think we need `prepare` for our dev-time workflow, so that if someone `npm install`'s and `npm link`'s the library it fully works. Without `prepare`, `npm install` will install dependencies but not transpile TS, which I don't think is our current expectation.

And in case you ask ;), I didn't choose `postinstall` for this cause `postinstall` is run even when we are ourselves a dependency (which is usually the case in users' environments). In that case they should have pretranspiled files from the `prepublish` step (which we ourselves run). OTOH `prepare` is *only* run when the module is itself the top-level project (i.e. not a dependency).

OsvaldoRosado:
It looks like this is failing our tests <= node 0.12. I think its the use of arrow function in endtoend.js

joshgav:
@OsvaldoRosado 

> I think its the use of arrow function in endtoend.js

Thanks! Fixed.

Issue #225, opened 2017-04-09
Better differentiate Client and Server requests


joshgav:
In a nutshell: ClientRequests are AppInsights Dependency items and should be more differentiated from ServerRequests which are AppInsights Request items.

`AutoCollection/ClientRequests.ts` to a significant degree conflates outgoing HTTP requests, which are AppInsights RemoteDependency items, with incoming HTTP requests, which are AppInsights Request items. In particular, both derive from `RequestParser` and both are named `*Request`.

It would be clearer if ClientRequests was renamed to ClientDependencies or HttpDependencies and  was derived from Dependency-related base classes rather than RequestParser. This would more clearly differentiate these Dependency items from Request items.

This is particularly helpful when reasoning about correlation. Incoming HTTP requests may be coming from another AppInsights-instrumented service and so our local service looks for several headers in the incoming request to add to its new Request item for correlation. Outgoing HTTP requests are RemoteDependency items and can include headers from the local client and root Request which the remote receiver can use for further correlation in the Request item it will create.

That is, a local Request is a RemoteDependency of the source, calling service; and a local RemoteDependency is a Request in the target, called service.

I've prepared a PR implementing the above suggested changes and will submit shortly.

AlexBulankou:
To summarize the file renaming that is proposed to happen as part of this item:
1. AutoCollection/ClientRequestParser->AutoCollection/HttpDependencyParser
2. AutoCollection/ClientRequests -> AutoCollection/HttpDependencies
3. AutoCollection/ServerRequestParser -> AutoCollection/HttpRequestParser
4. AutoCollection/ServerRequests -> AutoCollection/HttpRequests

@OsvaldoRosado, @joshgav  Is that all? Rather than trying to merge Josh's original PR, might be easier to start again on the latest version and limit the changes to just this renaming.

PR #226, opened 2017-04-09
ClientRequests -> OutgoingHttpDependencies

joshgav:
see https://github.com/Microsoft/ApplicationInsights-node.js/issues/225 for background.


joshgav:
cc @jasongin

package.json
```diff
@@ -40,9 +40,9 @@
     }
   ],
   "scripts": {
-    "build": "tsc --module commonjs --declaration applicationinsights.ts",
-    "prepublish": "tsc --module commonjs --declaration applicationinsights.ts",
-    "pretest": "find Tests -type f -name \"*.ts\" | xargs tsc --module commonjs",
+    "build": "tsc --module commonjs --target es5 --declaration applicationinsights.ts",
```
OsvaldoRosado: Does this have any compat issues to be aware of in old node?

OsvaldoRosado:
While we're renaming this, why not just HttpDependencies?

Also, for some historical context there's some conversation about this naming at #129 .  Function names were changed to be in line with request/dependency as used in our other SDKs, but these class namings are from the original structure. I think your PR would provide a good cleanup.

joshgav:
@OsvaldoRosado 

> While we're renaming this, why not just HttpDependencies?

Indeed 😄 , I actually had that as an intermediate class at some point along the way, but then realized that incoming HTTP requests will always be Requests in the App Insights sense and didn't want outgoing requests, which will always be RemoteDependencies in the App Insights sense, to be confused. For the same reason I kept the actual class as OutgoingHttpDependencies so that no one makes the mistake of thinking this relates to incoming HTTP requests. I still lean to keeping the longer name cause I think it would be best to be as clear as possible in this area, since we have to reason about it a lot for correlation.

Will check out that other issue too, thanks!

package.json
joshgav: According to CI the only problem is with use of Object.keys() deep in a helper method. I'll check that out, but it seems to indicate the rest is okay?

joshgav:
All green 🎉 ! PTAL.

AutoCollection/OutgoingHttpDependencies.ts
```diff
@@ -0,0 +1,128 @@
...
+                !request['getHeader'](HttpHeaders.RemoteDependency.myIkey)) {
+            request['setHeader'](HttpHeaders.RemoteDependency.myIkey, outgoingHttpDependencyParser.myIkeyHash);
+        }
+
+        // Collect dependency telemetry about the request when it finishes.
+        if (request.on) {
+            request.on('response', (response: http.IncomingMessage) => {
+                outgoingHttpDependencyParser.onResponse(response, properties);
+                // TODO(joshgav): Is it okay to change ClientResponse -> IncomingMessage?
+                var context : { [name: string]: any; } = { "http.RequestOptions": requestOptions, "http.ClientRequest": request, "http.ClientResponse": response };
```
OsvaldoRosado: Do you know when Node switched from ClientResponse to IncomingMessage? I see ClientResponse isn't documented anymore

joshgav: not completely sure but it's in the index.d.ts file: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v0/index.d.ts#L450-L453

the `connection` property is also gone, have been updating that around the codebase too

OsvaldoRosado:
Closing as this was addressed in #309

Issue #192, opened 2017-02-14
Duplicated Metrics and Performance Counters


debugthings:
With the new billing model being rolled out we should remove the collection of both the performance counters and the custom metrics in the [`AutoCollectPerformance`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts) class.

Namely we should pick one or the other. Either remove [`_trackLegacyPerformance()`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts#L116-L220) from the child methods in [`trackPerformance()`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts#L110-L114) or we should remove all instances of trackMetric from these methods [`_trackCpu()`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts#L208-L212), [`_trackMemory()`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts#L228-L232), [`_trackNetwork()`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/Performance.ts#L257-L261).

debugthings:
🎉🎈🎉🎈🎉🎈!!!

Issue #162, opened 2017-01-04
Types generated from the public schema should contain comments.


dnduffy:
The Bond schema files contain comments and other annotations which should be output as comments in the generated type files.

Issue #122, opened 2016-10-04
Update bond schemas


KamilSzostak:
New schemas: https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/ContextTagKeys.bond

PR #228, opened 2017-04-15
Use autogenerated schemas

OsvaldoRosado:
This PR will:
- Update BOND files to latest version
- Regenerate contracts from the BOND using a new autogeneration format that has comments and uses `require`
- Remove all code reliant on older schemas (old perfcollector reporting for example)
- Update all code to use the autogenerated contracts rather than a hand modified file

Addresses issues:
#122 
#162 
#192


Library/Client.ts
```diff
@@ -240,30 +240,30 @@ class Client {
...
-                data.baseData.properties = this.commonProperties;
-            } else {
-                // otherwise, check each of the common ones
-                for (var name in this.commonProperties) {
-                    // only override if the property `name` has not been set on this item
-                    if (!data.baseData.properties[name]) {
-                        data.baseData.properties[name] = this.commonProperties[name];
+        data: Contracts.Data<Contracts.Domain>,
+        tagOverrides?: { [key: string]: string; }): Contracts.Envelope {
+        if (Contracts.domainSupportsProperties(data.baseData)) { // Do instanceof check. TS will automatically cast and allow the properties property
```
OsvaldoRosado: This is slightly different logic, but I think it is more correct (plus type safe!). The difference is the previous logic would dump a default properties property on any supplied baseData. While all of our data types support this key, it is not on the base Domain class, so new ones may intentionally not have it

AutoCollection/ServerRequestParser.ts
```diff
@@ -88,7 +88,6 @@ class ServerRequestParser extends RequestParser {
         newTags[ServerRequestParser.keys.locationIp] = tags[ServerRequestParser.keys.locationIp] || this._getIp();
         newTags[ServerRequestParser.keys.sessionId] = tags[ServerRequestParser.keys.sessionId] || this._getId("ai_session");
         newTags[ServerRequestParser.keys.userId] = tags[ServerRequestParser.keys.userId] || this._getId("ai_user");
-        newTags[ServerRequestParser.keys.userAgent] = tags[ServerRequestParser.keys.userAgent] || this.userAgent;
```
OsvaldoRosado: This field was dropped from the schema - so its logic is removed here accordingly

AutoCollection/Performance.ts
```diff
@@ -208,10 +163,6 @@ class AutoCollectPerformance {
             this._client.trackMetric(combinedName + "nice", totalNice / combinedTotal);
```
OsvaldoRosado: These metric names have to be verified for correctness as the legacy perfcounters are removed with these changes.

package.json
```diff
@@ -27,6 +27,10 @@
       "email": "kszostak@microsoft.com"
     },
     {
+      "name": "osrosado",
```
SergeyKanzhelev: let's replace individual names to the AppInsightsSDK account

OsvaldoRosado: Done

AutoCollection/Performance.ts
OsvaldoRosado: Had offline discussion. Will keep the clean trackMetric names and ensure compatibility is kept on backend for legacy name conversion

Tests/Library/Client.tests.ts
```diff
@@ -626,7 +626,7 @@ describe("Library/Client", () => {
             client1.commonProperties = commonproperties;
             client1.config.samplingPercentage = 99;
             mockData.baseData.properties = JSON.parse(JSON.stringify(properties));
-            var env = client1.getEnvelope(mockData);
+            var env = <any>client1.getEnvelope(mockData);
```
AlexBulankou: Why do we need <any>, isn't sample rate specified on Contracts.Envelope?

OsvaldoRosado: IIRC it's not casted because of sample rate, but the basedata assertions. The base type in our schema doesn't have the properties object.

AutoCollection/Performance.ts
```diff
@@ -200,18 +155,15 @@ class AutoCollectPerformance {
...
-            this._client.trackMetric(combinedName + "user", totalUser / combinedTotal);
-            this._client.trackMetric(combinedName + "sys", totalSys / combinedTotal);
-            this._client.trackMetric(combinedName + "nice", totalNice / combinedTotal);
-            this._client.trackMetric(combinedName + "idle", totalIdle / combinedTotal);
-            this._client.trackMetric(combinedName + "irq", totalIrq/ combinedTotal);
-
-            // todo: remove this legacy counter once the UI updates (~june 2015)
-            this._trackLegacyPerformance(PerfCounterType.ProcessorTime, totalUser / combinedTotal);
-            this._trackLegacyPerformance(PerfCounterType.PercentProcessorTime, (combinedTotal - totalIdle) / combinedTotal);
+            this._client.trackMetric("\\Processor(_Total)\\% Processor Time", (combinedTotal - totalIdle) / combinedTotal);
```
OsvaldoRosado: Following more offline discussions, common perf counters with .NET SDKs will be reported using the metric names they currently use until a new set is decided. There should be no user impact, as these are converted into appropriate values in the performanceCounter Analytics table by the backend

PR #231, opened 2017-04-28
Updating how cross-application correlations are tracked

MSLaguana:
Instead of using a hash of the instrumentation key we now use the appId, matching the .NET sdk.
We also use different headers to match the .NET sdk.

### DIFF
--- a/AutoCollection/ClientRequestParser.ts
+++ b/AutoCollection/ClientRequestParser.ts
@@ -6,7 +6,9 @@ import Contracts = require("../Declarations/Contracts");
 import Client = require("../Library/Client");
 import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
+import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import RequestParser = require("./RequestParser");
+import CorrelationIdManager = require("../Library/CorrelationIdManager");
 
 /**
  * Helper class to read data from the requst/response objects and convert them into the telemetry contract
@@ -37,7 +39,7 @@ class ClientRequestParser extends RequestParser {
      */
     public onResponse(response: http.ClientResponse, properties?: { [key: string]: string }) {
         this._setStatus(response.statusCode, undefined, properties);
-        this.correlationId = Util.getCorrelationContextTarget(response);
+        this.correlationId = Util.getCorrelationContextTarget(response, RequestResponseHeaders.requestContextTargetKey);
     }
 
     /**
@@ -52,12 +54,14 @@ class ClientRequestParser extends RequestParser {
         let remoteDependency = new Contracts.RemoteDependencyData();
         remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
 
+        remoteDependency.target = urlObject.hostname;
         if (this.correlationId) {
             remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_AI;
-            remoteDependency.target = urlObject.hostname + " | " + this.correlationId;
+            if (this.correlationId !== CorrelationIdManager.correlationIdPrefix) {
+                remoteDependency.target = urlObject.hostname + " | " + this.correlationId;
+            }
         } else {
             remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
-            remoteDependency.target = urlObject.hostname;
         }
 
         remoteDependency.name = dependencyName;
--- a/AutoCollection/ClientRequestParser.ts
+++ b/AutoCollection/ClientRequestParser.ts
@@ -6,7 +6,6 @@ import Contracts = require("../Declarations/Contracts");
 import Client = require("../Library/Client");
 import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
-import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import RequestParser = require("./RequestParser");
 
 /**
@@ -38,17 +37,7 @@ class ClientRequestParser extends RequestParser {
      */
     public onResponse(response: http.ClientResponse, properties?: { [key: string]: string }) {
         this._setStatus(response.statusCode, undefined, properties);
-        const contextHeaders = response.headers && response.headers[RequestResponseHeaders.requestContextHeader];
-        if (contextHeaders) {
-            const keyValues = contextHeaders.split(",");
-            for(let i = 0; i < keyValues.length; ++i) {
-                const keyValue = keyValues[i].split("=");
-                if (keyValue.length == 2 && keyValue[0] == RequestResponseHeaders.requestContextTargetKey) {
-                    this.correlationId = keyValue[1];
-                    break;
-                }
-            }
-        }
+        this.correlationId = Util.getCorrelationContextTarget(response);
     }
 
     /**
--- a/AutoCollection/ClientRequestParser.ts
+++ b/AutoCollection/ClientRequestParser.ts
@@ -13,7 +13,7 @@ import RequestParser = require("./RequestParser");
  * Helper class to read data from the requst/response objects and convert them into the telemetry contract
  */
 class ClientRequestParser extends RequestParser {
-    private targetIKeyHash: string;
+    private correlationId: string;
 
     constructor(requestOptions: string | http.RequestOptions | https.RequestOptions, request: http.ClientRequest) {
         super();
@@ -38,8 +38,17 @@ class ClientRequestParser extends RequestParser {
      */
     public onResponse(response: http.ClientResponse, properties?: { [key: string]: string }) {
         this._setStatus(response.statusCode, undefined, properties);
-        this.targetIKeyHash =
-            response.headers && response.headers[RequestResponseHeaders.targetInstrumentationKeyHeader];
+        const contextHeaders = response.headers && response.headers[RequestResponseHeaders.requestContextHeader];
+        if (contextHeaders) {
+            const keyValues = contextHeaders.split(",");
+            for(let i = 0; i < keyValues.length; ++i) {
+                const keyValue = keyValues[i].split("=");
+                if (keyValue.length == 2 && keyValue[0] == RequestResponseHeaders.requestContextTargetKey) {
+                    this.correlationId = keyValue[1];
+                    break;
+                }
+            }
+        }
     }
 
     /**
@@ -54,9 +63,9 @@ class ClientRequestParser extends RequestParser {
         let remoteDependency = new Contracts.RemoteDependencyData();
         remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
 
-        if (this.targetIKeyHash) {
+        if (this.correlationId) {
             remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_AI;
-            remoteDependency.target = urlObject.hostname + " | " + this.targetIKeyHash;
+            remoteDependency.target = urlObject.hostname + " | " + this.correlationId;
         } else {
             remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
             remoteDependency.target = urlObject.hostname;
--- a/AutoCollection/ClientRequests.ts
+++ b/AutoCollection/ClientRequests.ts
@@ -7,6 +7,7 @@ import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import ClientRequestParser = require("./ClientRequestParser");
+import { CorrelationContextManager, CorrelationContext } from "./CorrelationContextManager";
 
 class AutoCollectClientRequests {
     public static disableCollectionRequestOption = 'disableAppInsightsAutoCollection';
@@ -83,18 +84,25 @@ class AutoCollectClientRequests {
         // The getHeader/setHeader methods aren't available on very old Node versions, and
         // are not included in the v0.10 type declarations currently used. So check if the
         // methods exist before invoking them.
-        if (client.config && client.config.correlationId &&
-            Util.canIncludeCorrelationHeader(client, requestParser.getUrl()) &&
+        if (Util.canIncludeCorrelationHeader(client, requestParser.getUrl()) &&
             request['getHeader'] && request['setHeader']) {
-            const correlationHeader = request['getHeader'](RequestResponseHeaders.requestContextHeader);
-            if (correlationHeader) {
-                const components = correlationHeader.split(",");
-                const key = `${RequestResponseHeaders.requestContextSourceKey}=`;
-                if (!components.some((value) => value.substring(0,key.length) === key)) {
-                    request['setHeader'](RequestResponseHeaders.requestContextHeader, `${correlationHeader},${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
+            if (client.config && client.config.correlationId) {
+                const correlationHeader = request['getHeader'](RequestResponseHeaders.requestContextHeader);
+                if (correlationHeader) {
+                    const components = correlationHeader.split(",");
+                    const key = `${RequestResponseHeaders.requestContextSourceKey}=`;
+                    if (!components.some((value) => value.substring(0,key.length) === key)) {
+                        request['setHeader'](RequestResponseHeaders.requestContextHeader, `${correlationHeader},${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
+                    }
+                } else {
+                    request['setHeader'](RequestResponseHeaders.requestContextHeader, `${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
                 }
-            } else {
-                request['setHeader'](RequestResponseHeaders.requestContextHeader, `${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
+            }
+
+            const currentContext = CorrelationContextManager.getCurrentContext();
+            if (currentContext && currentContext.operation) {
+                request['setHeader'](RequestResponseHeaders.parentIdHeader, currentContext.operation.id);
+                request['setHeader'](RequestResponseHeaders.rootIdHeader, currentContext.operation.parentId);
             }
         }
 
--- a/AutoCollection/ClientRequests.ts
+++ b/AutoCollection/ClientRequests.ts
@@ -79,16 +79,23 @@ class AutoCollectClientRequests {
 
         let requestParser = new ClientRequestParser(requestOptions, request);
 
-        // Add the source ikey hash to the request headers, if a value was not already provided.
+        // Add the source correlationId to the request headers, if a value was not already provided.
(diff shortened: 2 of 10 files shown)


MSLaguana:
@nizarq Could you please take a look at this and make sure I didn't miss anything important?

MSLaguana:
This PR is based on the changes made to the .NET sdk. Regarding your points:

1. I queue up a single request for the `appId` when a `Config` is created, and retry only on non-4XX errors every `Config.correlationIdRetryInterval` (default 30 seconds). Until the request completes successfully, the `correlationId` will remain undefined, and not block execution at all.
2. I allow for different endpoints as much as the previous version did: https://github.com/Microsoft/ApplicationInsights-node.js/pull/231/files#diff-7b48e6fce63062fb2711022cca6afa31L28 seemed to hardcode the URL, and I use the same (configurable) base URL to fetch the appId.
3. I'm not sure what you mean here. Do you mean allowing users to disable cross component correlation?

AlexBulankou:
Thanks Jimmy! 
1. From my understanding of your code any time Config object is created you are initiating  a new network request without checking if request to resolve the same ikey has already completed or is in progress. We should not be doing this unnecessarily. 
2. Configurable base is not enough because users may be using one host name to send data (e.g. Vortex case) and another endpoint to resolve profile. We should expose this as separate configurable option, not derived from base.
3. To prevent uninitialized x-component dependencies (because of delay in resolution) that ultimately appear as extra nodes on app map, we should provide a way for a user to plug in a simple lookup table of ikey to appid mappings. For the entries provided in this lookup, calling profile endpoint will be unnecessary.

nizarq:
Since you made this change based on my original change. In the last couple of days I've sent these couple of updates to the original change. Please look at them and incorporate them as well. They address some of last remaining issues that Alex is pointing to.

https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/505
https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/526

nizarq:
@AlexBulankou 
> SDK should provide extensibility to specify instant resolution mechanism for customers who either wish to have no extra calls to profile endpoint or wish to always have them resolved, which wouldn't be the case because of requirement (1)

With 400s handled correctly, and ensuring a single call is made per ikey. I think above is a good to have at best. I would rather have customers ask us for this before we give it to them. For almost everyone out there in the wild, they will go against the public breeze endpoint and if breeze is up things work well. If not, we handle that gracefully.

MSLaguana:
I've updated it to only query once per ikey (per endpoint URL), and to default to a blank correlation id rather than lacking one, but it isn't clear to me exactly how that was supposed to work. @nizarq should we be sending the blank `cid-v1:` correlation ID on all requests (until we get the actual ID), or should we be checking for blank correlation IDs on inbound requests?

@AlexBulankou if you aren't content with having a single changable base endpoint, how do you want users to specify different endpoints for general telemetry and querying the appId?

AutoCollection/ClientRequestParser.ts
```diff
@@ -54,9 +63,9 @@ class ClientRequestParser extends RequestParser {
         let remoteDependency = new Contracts.RemoteDependencyData();
         remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
 
-        if (this.targetIKeyHash) {
+        if (this.correlationId) {
             remoteDependency.type = Contracts.RemoteDependencyDataConstants.TYPE_AI;
-            remoteDependency.target = urlObject.hostname + " | " + this.targetIKeyHash;
+            remoteDependency.target = urlObject.hostname + " | " + this.correlationId;
```
MSLaguana: @nizarq is it just this line that should check for a default `cid-v1:` correlationId? Still mark it as `TYPE_AI` since it specified something, but not include the default value in the `target`?

OsvaldoRosado:
@MSLaguana for the different base endpoint for the appid lookup, .NET uses the configuration option "ProfileQueryEndpoint". We can place this in our config object

MSLaguana:
@OsvaldoRosado How do you propose exposing it? As a parameter when constructing a Client / Config? As a parameter set immediately after constructing one, and we introduce a delay between constructing the Config and firing off the HTTP request? I know that the .NET sdk waits until the appId is needed before firing off a request, but I'd prefer to be more proactive to reduce the number of non-correlated requests if possible.

OsvaldoRosado:
@MSLaguana  I'm not a big fan of changing the constructors for this, especially since a constructor parameter doesn't exist for the telemetry endpoint, which is far more frequently changed (this endpoint is essentially only necessary to change for internal use). It would also need to be on the Config, Client, and the appInsights.setup method. Introducing a delay seems like a hack but might be the best option if you don't want to resolve until telemetry is generated. You can at least be proactive for changes to it with a custom setter on the property.

MSLaguana:
If we allow people to change the base URL after we have already started a request, what is the expected behavior? Cancel any pending requests and start again with the new URL? Use the new URL on the next retry after getting an error (including a 4XX error)?

I was thinking more like

```js
appInsights.setup("key").start();
appInsights.getClient().config.profileQueryEndpoint = "http://myCustomEndpoint.com"
```

and as long as that configuration was done synchronously with the creation of the client then it would be picked up in the request for the appId. We could also allow for environment variables, which might be easier.

OsvaldoRosado:
Note that you want `.client` instead of `getClient` in the above example, otherwise you now have two telemetry clients 😄 But yes, setting the property directly like that is what I have in mind.

I would expect telemetry generated after the URL has changed to no longer use the previously looked up appid. A new appid can be looked up immediately on set through a custom setter. Any pending requests (for the specific client at hand since you can have multiple) can be discarded

AutoCollection/ServerRequestParser.ts
```diff
@@ -34,8 +34,18 @@ class ServerRequestParser extends RequestParser {
             this.rawHeaders = request.headers || (<any>request).rawHeaders;
             this.socketRemoteAddress = (<any>request).socket && (<any>request).socket.remoteAddress;
             this.userAgent = request.headers && request.headers["user-agent"];
-            this.sourceIKeyHash =
-                request.headers && request.headers[RequestResponseHeaders.sourceInstrumentationKeyHeader];
+
+            if (request.headers && request.headers[RequestResponseHeaders.requestContextHeader]) {
+                const keyValues = request.headers[RequestResponseHeaders.requestContextHeader].split(",");
```
KamilSzostak: Consider moving this header parsing logic to `Utils` and reuse it in both `ClientRequestParser`  and `ServerRequestParser`.

MSLaguana:
@OsvaldoRosado See what you think of my changes to support profileQueryEndpoint. I don't currently support changing its value at an arbitrary time, but if that's a scenario that we want to support then I could.

@KamilSzostak Done.

OsvaldoRosado:
@MSLaguana I don't think we can get around having to support changing the value at arbitrary times. Anything that determines its configuration asynchronously will need to be able to do this (or refactor their telemetry logic to not import AppInsights until after configuration, since we automatically create a client on import)

MSLaguana:
@OsvaldoRosado How's this look then?

MSLaguana:
@nizarq @AlexBulankou What is the expected behavior if the network is out? e.g. we fail to resolve the endpoint? Should we retry in that scenario, or should we give up immediately?

MSLaguana:
It looks like the node.js SDK was missing the `x-ms-request-id` header on outbound requests. @nizarq can you also check to see that I've got the semantics of this right in https://github.com/Microsoft/ApplicationInsights-node.js/pull/231/commits/e1d91303a1254e1435db2764688578f3a3ae9a73 ?

AlexBulankou:
@nizarq @MSLaguana my proposed requirement (3) is closely related this PR: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/526 So far I'm not sure I understand how this addresses the problem of  initial couple of calls aggregated incorrectly in case they are not resolved, hence my last comment.

Library/Util.ts
```diff
@@ -178,5 +179,18 @@ class Util {
...
         return true;
     }
+
+    public static getCorrelationContextTarget(response: http.ClientResponse | http.ServerRequest) {
+        const contextHeaders = response.headers && response.headers[RequestResponseHeaders.requestContextHeader];
+        if (contextHeaders) {
+            const keyValues = contextHeaders.split(",");
+            for(let i = 0; i < keyValues.length; ++i) {
+                const keyValue = keyValues[i].split("=");
+                if (keyValue.length == 2 && keyValue[0] == RequestResponseHeaders.requestContextTargetKey) {
```
nizarq: In theory shouldn't you be using TargetKey for clientResponse and SourceKey for Server request. I know both values are the same so it doesn't really matter, but conceptually you have defined two different constants.

AutoCollection/ClientRequests.ts
```diff
@@ -79,16 +80,30 @@ class AutoCollectClientRequests {
...
+            if (client.config && client.config.correlationId) {
+                const correlationHeader = request['getHeader'](RequestResponseHeaders.requestContextHeader);
+                if (correlationHeader) {
+                    const components = correlationHeader.split(",");
+                    const key = `${RequestResponseHeaders.requestContextSourceKey}=`;
+                    if (!components.some((value) => value.substring(0,key.length) === key)) {
+                        request['setHeader'](RequestResponseHeaders.requestContextHeader, `${correlationHeader},${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
+                    }
+                } else {
+                    request['setHeader'](RequestResponseHeaders.requestContextHeader, `${RequestResponseHeaders.requestContextSourceKey}=${client.config.correlationId}`);
```
MSLaguana: It can be if a user explicitly overrides it. By default we set it to `"cid-v1:"` (the equivalent to .NET's `EmptyCorrelationId`)

AutoCollection/ServerRequests.ts
```diff
@@ -151,7 +151,7 @@ class AutoCollectServerRequests {
         var requestParser = _requestParser || new ServerRequestParser(request, correlationContext && correlationContext.operation.parentId || Util.newGuid());
 
         if (Util.canIncludeCorrelationHeader(client, requestParser.getUrl())) {
```
MSLaguana: I'm not sure, @KamilSzostak added in this check and may have more context. Perhaps it is because `trackRequestSync` is never called internally, so it must be called by the end-user and thus we want to track the call ignoring blacklists, but `trackRequest` here is also used internally?

Library/CorrelationIdManager.ts
```diff
@@ -0,0 +1,93 @@
+import https = require('https');
+import http = require('http');
+import url = require('url');
+
+class CorrelationIdManager {
+    public static correlationIdPrefix: "cid-v1:";
+
+    private static pendingLookups: {[key: string]: Function[]} = {};
+    private static completedLookups: {[key: string]: string} = {};
```
MSLaguana: These are used to implement the requirement that we only make a request once per session, even if someone creates multiple AI clients / configs.

Library/CorrelationIdManager.ts
```diff
@@ -0,0 +1,93 @@
...
+import http = require('http');
+import url = require('url');
+
+class CorrelationIdManager {
+    public static correlationIdPrefix: "cid-v1:";
+
+    private static pendingLookups: {[key: string]: Function[]} = {};
+    private static completedLookups: {[key: string]: string} = {};
+
+    public static queryCorrelationId(endpointBase: string, instrumentationKey: string, correlationIdRetryInterval: number, callback: (correlationId: string) => void) {
```
MSLaguana: It is possible for someone to construct multiple AI clients in the one node process/application with different instrumentation keys (or the same instrumentation key)

Library/Util.ts
MSLaguana: You are right, I missed that when I was refactoring the two instances of this

Library/Config.ts
```diff
@@ -45,16 +47,28 @@ class Config {
...
+        this.setCorrelationId = (correlationId) => this.correlationId = correlationId;
+
+        CorrelationIdManager.queryCorrelationId(
+            this.profileQueryEndpoint,
+            this.instrumentationKey,
+            this.correlationIdRetryInterval,
+            this.setCorrelationId);
     }
 
+    public setProfileQueryEndpoint(endpoint: string) {
```
OsvaldoRosado: I think the get/set syntax for accessors in TS would be preferable (see https://www.typescriptlang.org/docs/handbook/classes.html)/ Then `this.profileQueryEndpoint` can be set directly and still trigger the appid query

Library/Config.ts
```diff
@@ -8,37 +8,67 @@ class Config {
...
+        this.setCorrelationId = (correlationId) => this.correlationId = correlationId;
+
+        CorrelationIdManager.queryCorrelationId(
+            this.profileQueryEndpoint,
+            this.instrumentationKey,
+            this.correlationIdRetryIntervalMs,
+            this.setCorrelationId);
+    }
+
+    public setProfileQueryEndpoint(endpoint: string) {
```
OsvaldoRosado: I think the get/set syntax for accessors in TS would be preferable (see https://www.typescriptlang.org/docs/handbook/classes.html ) Then `this.profileQueryEndpoint` can be set directly and still trigger the appid query

MSLaguana: Done

Library/Config.ts
```diff
@@ -46,28 +45,26 @@ class Config {
...
-        this.correlationId = CorrelationIdManager.correlationIdPrefix; // Initialize with a blank correlation ID until we fetch the correct value.
-
         this.setCorrelationId = (correlationId) => this.correlationId = correlationId;
 
-        CorrelationIdManager.queryCorrelationId(
-            this.profileQueryEndpoint,
-            this.instrumentationKey,
-            this.correlationIdRetryIntervalMs,
-            this.setCorrelationId);
+        this.profileQueryEndpoint = process.env[Config.ENV_profileQueryEndpoint] || this.endpointBase;
```
OsvaldoRosado: Should this be setting the private property directly instead? Or does the public one delay the appid query as well?

MSLaguana: I've moved the delay into the `queryCorrelationId` call which is invoked in the setter, so by invoking the setter here we'll queue the fetching of the appId, but allow for it to be cancelled and restarted with a new endpoint if necessary.

OsvaldoRosado: 👍

AutoCollection/ServerRequests.ts
nizarq: Can you please ping @KamilSzostak  then and agree to what's the right thing to do. I think it's accidentally left out and should be fixed.

Library/CorrelationIdManager.ts
nizarq: Isn't the caller always just automatic collection code? Isn't there going to be a single config / client for automatic collection? I think you don't need static.

Library/CorrelationIdManager.ts
nizarq: Even when it comes to automatic collection? Which is where this is called from.

MSLaguana: `queryCorrelationId` is called when a `Config` is constructed, or when the `profileQueryEndpoint` on a `Config` is changed. A `Config` is constructed when a `Client` is constructed, and you can construct clients explicitly via `appInsights.getClient("key")` if you want.

Library/CorrelationIdManager.ts
OsvaldoRosado: We have manual methods for instrumenting HTTP calls that add headers, more than just the automatic collection client will use this

Library/CorrelationIdManager.ts
```diff
@@ -0,0 +1,93 @@
+import https = require('https');
+import http = require('http');
+import url = require('url');
+
+class CorrelationIdManager {
+    public static correlationIdPrefix: "cid-v1:";
+
+    private static pendingLookups: {[key: string]: Function[]} = {};
```
MSLaguana: Done

Library/Config.ts
```diff
@@ -8,37 +8,64 @@ class Config {
...
             "*.blob.core.cloudapi.de",
             "*.blob.core.usgovcloudapi.net"];
+        
+        this.setCorrelationId = (correlationId) => this.correlationId = correlationId;
+
+        this.profileQueryEndpoint = process.env[Config.ENV_profileQueryEndpoint] || this.endpointBase;
     }
 
+    public set profileQueryEndpoint(endpoint: string) {
+        CorrelationIdManager.cancelCorrelationIdQuery(this._profileQueryEndpoint, this.instrumentationKey, this.setCorrelationId);
```
AlexBulankou: Do we need unit test to verify that cancelCorrelationIdQuery is called and that it cancels pending queries?

Library/CorrelationIdManager.ts
```diff
@@ -0,0 +1,95 @@
...
+import http = require('http');
+import url = require('url');
+
+class CorrelationIdManager {
+    public static correlationIdPrefix: "cid-v1:";
+
+    // To avoid extraneous HTTP requests, we maintain a queue of callbacks waiting on a particular appId lookup,
+    // as well as a cache of completed lookups so future requests can be resolved immediately.
+    private static pendingLookups: {[key: string]: Function[]} = {};
+    private static completedLookups: {[key: string]: string} = {};
```
AlexBulankou: Do we need unit test to verify  pendingLookups and completedLookups state?

Library/CorrelationIdManager.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    public static queryCorrelationId(endpointBase: string, instrumentationKey: string, correlationIdRetryInterval: number, callback: (correlationId: string) => void) {
+        // GET request to `${this.endpointBase}/api/profiles/${this.instrumentationKey}/appId`
+        // If it 404s, the iKey is bad and we should give up
+        // If it fails otherwise, try again later
+        const appIdUrlString = `${endpointBase}/api/profiles/${instrumentationKey}/appId`;
```
OsvaldoRosado: Shouldn't this be profileQueryEndpoint?

Library/Config.ts
```diff
@@ -8,37 +8,64 @@ class Config {
...
     public static ENV_iKey = "APPINSIGHTS_INSTRUMENTATIONKEY";
     public static legacy_ENV_iKey = "APPINSIGHTS_INSTRUMENTATION_KEY";
+    public static ENV_profileQueryEndpoint = "APPINSIGHTS_PROFILE_QUERY_ENDPOINT";
 
     public instrumentationKey: string;
-    public instrumentationKeyHash: string;
+    public correlationId: string;
     public sessionRenewalMs: number;
     public sessionExpirationMs: number;
+    public endpointBase: string;
```
OsvaldoRosado: Should this be private? Only ever would internal users want to change both things that are computed based off this. An uninformed user changing this accidentally instead of endpointUrl will cause appId resolution to fail.

OsvaldoRosado: Internal users can easily set both endpointUrl and profileQueryEndpoint. IMO endpointBase can just be a private const used to avoid duplicating the default string

MSLaguana:
I've addressed @OsvaldoRosado's comments, and rebased to clean up the commit history. If you're happy with it, please complete this PR.

Issue #183, opened 2017-02-01
Loading pachage.json of hosting application to get the application version won;t work if System.js loader is used


anliakho:
Line in question uses `require(packageJsonPath || "../../../package.json")`
this will fail is System.js is used as package.json is not a dependency of ApplicationInsights

https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Context.ts#L29

PROPOSAL: provide an overload for ApplicationInsight client constructor that will also take application version and description in and if passed-in use that instead of trying to retrieve it.

OsvaldoRosado:
The version number could get put as a static field in `ApplicationInsights.ts` so as to not rely on package.json

anliakho:
@OsvaldoRosado just to be clear the package.json I was referring to was a package.json from the app hosting application insights, there is really no problem retrieving the application insights version from it's package file as it all lives in proper relative location, the issue arises when assumptions are made about the hosting app's package file lopcation

gzepeda:
Hello @OsvaldoRosado and @anliakho. 

Right now after retrieving the application version and description from the package.json file, it stores them here (code snippet from Context.ts)

```
        this.tags[this.keys.applicationVersion] = version;
        // TODO: consider sending it as a custom property
        //if(description) {
        //    this.tags[this.keys.applicationBuild] = description;
        //}
```

Should the values be saved on the same place? do we uncomment the above code?

Also, not only a new constructor needs to be created for the Client class, but also for the Context class;  so the app version and description can be inserted by whatever client is using application insights. Is this true?

Let us know if we are under the correct line of thought.

OsvaldoRosado:
I would leave that commented bit commented for now, so we can address it as a separate issue. 

A manually passed in version should indeed still end up in these context tags. 

I think this can be done without a change to constructors. The property is already publicly available within `client.context.tags`. Wanting to catch this before initialization shouldn't be necessary because the automatic insertion code is within a try catch ( @anliakho  can verify if this does or does not crash as is). 

This situation may be uncommon enough that telling users to set context directly like this is fine (it is publicly accessible for reasons just like this). But we can provide a helper method on the context, config, or client (I can think of arguments for each) if it would be more convenient.

PR #233, opened 2017-05-02
Set the host app version to the context tags. #183

gabsanchez:
#183 Sets the host app version to the context tags, so we don't rely on the package.json.
Team:
@gabsanchez
@ronaldour
@childmario95
@tohka53
@mariopg

Engineers in charge:
@oscargarciacolon
@gzepeda


Library/Client.ts
```diff
@@ -376,6 +377,14 @@ class Client {
 
         return newTags;
     }
+    /**
+     * Sets the client app version to the contest tags.
```
OsvaldoRosado: Nit: typo on contest > context

Library/Client.ts
```diff
@@ -376,6 +377,14 @@ class Client {
...
+     * @param version, takes the host app version.
+     */
+    public setVersion(version: string)
```
OsvaldoRosado: What do you think of `overrideApplicationVersion` ?

OsvaldoRosado:
Thanks for submitting this!

gabsanchez:
@OsvaldoRosado thank you for your feedback, please let me know if there's anything else I can do to complete this pull request.

Tests/Library/Client.tests.ts
```diff
@@ -805,4 +805,10 @@ describe("Library/Client", () => {
             assert.ok(!processorExecuted, "telemetry processor should NOT be executed");
         });
     });
+    describe("#setVersion()", () => {
```
OsvaldoRosado: The method name should be updated here as well (string and the method call below)

gzepeda:
Kudos to the team!

Issue #55, opened 2015-11-24
How to get severity level values


bragma:
Hi,
I'm working on a transport for Winston logger and I'd like to map Winston log levels to AI's trace levels. I've seen an enumeration of supported levels in the "Contracts" module, but I am not sure on how to extract those values for using in my code. What's the suggested way?
Thanks!

scsouthw:
Not sure I'm understanding the question, but you can set severity by passing it to the trackTrace method of the client: https://github.com/Microsoft/ApplicationInsights-node.js/blob/master/Library/Client.ts#L70-L71

I'm not sure what the best way to extract the enum for use in your code would be. It should be possible to import the contracts module after it is installed via npm but the reference path would be messy. I doubt these will change so you could redeclare the enum in your code as a workaround.

bragma:
@scsouthw  sorry, I've not described well what I'm trying to do. As part of my Winston transport, I have to map default Winston's log levels (strings) to App Insight's levels (numbers). 

```js
	var levels = {
...
		crit: 4,	// AI 'Critical' 
		error: 3,	// AI 'Error' 
		warning: 2,	// AI 'Warning' 
...
	};
```

Instead of inserting the numeric values corresponding to AI's levels, I'd like to reference the enum values from the Contracts file, something like this:

```js
	var levels = {
...
		crit: AppInsights.Contracts.SeverityLevel.Critical,
...
	};
```

As you suggest I could import the contracts module, but it's messy. What about making them accessible directly from the ApplicationInsights class? It seems to me that's the only exported symbol from the module.

PR #234, opened 2017-05-02
Provide access to contracts

ronaldour:
This will provide access to contracts by client class, thus this will provide access to severity levels #55 

@gabsanchez @ronaldour @childmario95 @tohka53 @mariopg @oscargarciacolon @gzepeda

### DIFF
--- a/applicationinsights.ts
+++ b/applicationinsights.ts
@@ -7,6 +7,7 @@ import AutoCollectServerRequests = require("./AutoCollection/ServerRequests");
 import Client = require("./Library/Client");
 import Config = require("./Library/Config");
 import Context = require("./Library/Context");
+import Contracts = require("./Declarations/Contracts");
 import Logging = require("./Library/Logging");
 import Util = require("./Library/Util");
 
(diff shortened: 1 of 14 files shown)


OsvaldoRosado:
Exposing this on `Client` seems like the wrong choice to me. Why not the root applicationinsights.ts?

ronaldour:
we thought about the two options but you are right it would be better on applicationinsights.ts

ronaldour:
Updated to access contracts by applicationinsights.ts  class

applicationinsights.ts
```diff
@@ -22,6 +23,7 @@ class ApplicationInsights {
     * The default client.
     */
     public static client: Client;
+    public static contracts = Contracts;
```
OsvaldoRosado: Do you think this should be capitalized? I know all of the other fields are lowercase, but this isn't a field so much as exposing a namespace. @KamilSzostak does JS SDK have any precedence on this to follow?

KamilSzostak: Not really. The only problem I had was with source maps, which is case-sensitive for file paths, but this is not the case here.

gzepeda:
Kudos to the team!

PR #235, opened 2017-05-05
Adding diagnostic-channel

MSLaguana:
By using diagnostic-channel and diagnostic-channel-publishers, we can support context tracking through third-party libraries as well as getting additional telemetry for those dependencies.


OsvaldoRosado:
I think we should turn dependency correlation on by default with this, especially now that we have sampling to address the previous concern over data rate. The fundamental correlation bits (zone.js) have been present since the last version with no complaints. Are there any concerns with this? @KamilSzostak  @AlexBulankou @SergeyKanzhelev @joshgav

joshgav:
@OsvaldoRosado enabling by default would make sense for users. Was the reason for not enabling by default originally only to gain confidence in the functionality? Is there a perf cost?

As an aside, we should be sure to distinguish between:
* context restoration across async calls within a process (e.g. with Zone.js)
* event correlation across service/process boundaries (e.g. with HTTP headers)

OsvaldoRosado:
@joshgav Distinguishing is important. Here I'm speaking specifically about context restoration across async calls. HTTP Header based cross-process correlation is already working by default IIRC.

This was indeed originally not enabled by default to gain confidence in the feature. I don't have hard numbers of perf cost since we don't have a robust performance measurement setup yet for this SDK like we do for .NET

KamilSzostak:
IMHO we should enable dependency correlation by default. It's beneficial to a user even if there is potentially a small perf cost. @OsvaldoRosado did you enable it for Breeze, have we noticed any perf impact?

KamilSzostak:
Could you also update a README file and describe the functionality, e.g. list all available diagnostic-channels, how to enable, disable. Thanks!

OsvaldoRosado:
@KamilSzostak I didn't get a chance to enable it for breeze outside of on my own box. It worked fine on my box :stuck_out_tongue:

PR #237, opened 2017-05-06
Fixing CorrelationId

MSLaguana:
There was a typo which lead to the `correlationIdPrefix` being null.


KamilSzostak:
I made a very similar mistake last Friday, I'm windering why TypeScript compiler doesn't give any warning. If you define a variable like this `let test: 'test';` the output javascript doesn't contain it.

Issue #227, opened 2017-04-13
Support Typings Definitions for latest versions


myunggeun-song:
We are using Application Insights with TypeScript and are using Typings as well. But I found there is only 0.15.12 version's definition in the registry of Typings. Would you add Typescript definitions of each version including latest to the registry?

* Registry: https://api.typings.org/entries/dt/applicationinsights/versions
* Application Insights 0.15.12 Definition: https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/253e456e3c0bf4bd34afaceb7dcbae282da14066/applicationinsights/index.d.ts

FYI, we are using 0.18.0 and the interfaces of 0.15.12 are not the same as it, so we are using it with workaround.

OsvaldoRosado:
I believe the latest source correctly generates ts types embedded in the npm package, making this just work (at least for TS 2.0+).

I'll verify with a generated package, but copying latest build output directly to my node_modules folder seems to show correct types in vscode.

Issue #158, opened 2016-12-16
Type declarations for App Insights available via @types/applicationinsights are out-of-sync


SergeyKanzhelev:

jbrantly:
It seems like the npm package should just ship with definitions since the project itself is written in TypeScript.

lostintangent:
I'd second @jbrantly's comment!

OsvaldoRosado:
Closing to merge with #227

PR #240, opened 2017-05-10
Include typescript definitions in NPM package

OsvaldoRosado:
Addresses #227 (and #158) by including types directly on the NPM package, removing the need for @types.

As a side effect, tsc will complain if this is used in a project without types for node (as our public SDK methods have input parameters of types from node, such as http.ServerRequest). This seems reasonable to expect to be available, as you won't get very far in a Node project using TS without node types available :smile:

Zone types are moved from a comment reference to included for compile in tsconfig. This prevents the reference from being included in our exported types (avoiding a build error for anyone missing Zone types)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -13,8 +13,8 @@ Tests
 
 # Don't publish source files that aren't needed in package
 *.ts
+!*.d.ts
 *.js.map
-*.d.ts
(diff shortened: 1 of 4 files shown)


joshgav:
Awesome! Thanks!

PR #248, opened 2017-05-12
Fix SDK version reports as node:unknown

OsvaldoRosado:
This code was out of date after #224, leading to the package.json being not found


KamilSzostak:
Do you want to release it in 0.20.1 or 0.21? Please assign a milestone.

OsvaldoRosado:
Let's do 0.20.1. This is worth including in a point release if we make one. If we end up not having one, I'll move to 0.21

PR #254, opened 2017-05-19
Fix SDK crash when Error.prepareStackTrace returns a non-string

OsvaldoRosado:
This turns off Zone stack-rewriting which can't handle non-default stack trace formats. This resolves an incompatibility with "caller", which when used in conjunction with this SDK would crash upon use.

### DIFF
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -170,11 +170,11 @@ export class CorrelationContextManager {
 
             // Is this object set to rewrite the stack?
             // If so, we should turn off some Zone stuff that is prone to break
-            var stackRewrite = orig['stackRewrite'];
-            if (orig['prepareStackTrace']) {
-                orig['stackRewrite'] = false;
-                var stackTrace = orig['prepareStackTrace'];
-                orig['prepareStackTrace'] = (e, s) => {
+            var stackRewrite = (<any>orig).stackRewrite;
+            if ((<any>orig).prepareStackTrace) {
+                (<any>orig).stackRewrite= false;
+                var stackTrace = (<any>orig).prepareStackTrace;
+                (<any>orig).prepareStackTrace = (e, s) => {
                     // Remove some AI and Zone methods from the stack trace
                     // Otherwise we leave side-effects
                     s.splice(0, 3);
@@ -186,7 +186,7 @@ export class CorrelationContextManager {
             orig.apply(this, arguments);
 
             // Restore Zone stack rewriting settings
-            orig['stackRewrite'] = stackRewrite;
+            (<any>orig).stackRewrite = stackRewrite;
             
             // getOwnPropertyNames should be a superset of Object.keys...
             // This appears to not always be the case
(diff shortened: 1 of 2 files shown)


AutoCollection/CorrelationContextManager.ts
```diff
@@ -168,7 +168,25 @@ export class CorrelationContextManager {
                 return AppInsightsAsyncCorrelatedErrorWrapper.apply(Object.create(AppInsightsAsyncCorrelatedErrorWrapper.prototype), arguments);
             }
 
+            // Is this object set to rewrite the stack?
+            // If so, we should turn off some Zone stuff that is prone to break
+            var stackRewrite = orig['stackRewrite'];
```
AlexBulankou: Make "strackRewrite" and "prepare" constants?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -168,7 +168,25 @@ export class CorrelationContextManager {
...
+            if (orig['prepareStackTrace']) {
+                orig['stackRewrite'] = false;
+                var stackTrace = orig['prepareStackTrace'];
+                orig['prepareStackTrace'] = (e, s) => {
+                    // Remove some AI and Zone methods from the stack trace
+                    // Otherwise we leave side-effects
+                    s.splice(0, 3);
```
AlexBulankou: what is 3? Trying to see how hard it will be for someone unfamiliar with zone to support this later

KamilSzostak: Does `s` always have three methods? Is it possible that we are removing too much?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -168,7 +168,25 @@ export class CorrelationContextManager {
...
+                    return stackTrace(e, s);
+                }
+            }
+
+            // Apply the error constructor
             orig.apply(this, arguments);
+
+            // Restore Zone stack rewriting settings
+            orig['stackRewrite'] = stackRewrite;
```
AlexBulankou: do we want to wrap it into try...finally to restore orig always?

AutoCollection/CorrelationContextManager.ts
OsvaldoRosado: I don't expect the first 3 methods to change without a code change on our side, since they're all basically passing around our error handlers and zone error handlers. (In other words, the whole top of the stack is just AI SDK and Zone) If we ever change how our error handling works we'll have to adjust this number.

OsvaldoRosado: 3 here is the number of methods we add to the stack trace. The library that we crash on will still fail if we don't remove ourselves from the trace

AutoCollection/CorrelationContextManager.ts
OsvaldoRosado: Good idea!

AutoCollection/CorrelationContextManager.ts
OsvaldoRosado: I think that might be confusing for this use case. Perhaps I should use dot notation instead and cast to any. The only reason I'm using this bracket notation is to avoid typescript errors on these hidden properties and methods

AutoCollection/CorrelationContextManager.ts
OsvaldoRosado: I've added a UT to check if this number is correct. If we ever add more methods that make this magic number incorrect, the UT will fail, making it obvious this needs to change.

OsvaldoRosado: I've added a UT to check if this number is correct

PR #265, opened 2017-06-05
Updating headers to use new scheme

MSLaguana:
Request IDs are expected to follow https://github.com/lmolkova/correlation/blob/master/hierarchical_request_id.md
and https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md

This change defaults to using the new hierarchical ID format, but continues
to interoperate with the legacy format as well.

### DIFF
--- a/AutoCollection/ClientRequests.ts
+++ b/AutoCollection/ClientRequests.ts
@@ -20,6 +20,8 @@ class AutoCollectClientRequests {
 
     public static INSTANCE: AutoCollectClientRequests;
 
+    private static requestNumber = 1;
+
     private _client: Client;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
@@ -110,8 +112,15 @@ class AutoCollectClientRequests {
 
             const currentContext = CorrelationContextManager.getCurrentContext();
             if (currentContext && currentContext.operation) {
+                const uniqueParentId = currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.';
+                request['setHeader'](RequestResponseHeaders.requestIdHeader, uniqueParentId);
+                // Also set legacy headers
                 request['setHeader'](RequestResponseHeaders.parentIdHeader, currentContext.operation.id);
-                request['setHeader'](RequestResponseHeaders.rootIdHeader, currentContext.operation.parentId);
+                request['setHeader'](RequestResponseHeaders.rootIdHeader, uniqueParentId);
+
+                if (currentContext.operation.correlationContextHeader) {
+                    request['setHeader'](RequestResponseHeaders.correlationContextHeader, currentContext.operation.correlationContextHeader);
+                }
             }
         }
 
--- a/AutoCollection/ClientRequests.ts
+++ b/AutoCollection/ClientRequests.ts
@@ -91,6 +91,9 @@ class AutoCollectClientRequests {
 
         let requestParser = new ClientRequestParser(requestOptions, request);
 
+        const currentContext = CorrelationContextManager.getCurrentContext();
+        const uniqueParentId = currentContext && currentContext.operation && (currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.');
+
         // Add the source correlationId to the request headers, if a value was not already provided.
         // The getHeader/setHeader methods aren't available on very old Node versions, and
         // are not included in the v0.10 type declarations currently used. So check if the
@@ -110,16 +113,15 @@ class AutoCollectClientRequests {
                 }
             }
 
-            const currentContext = CorrelationContextManager.getCurrentContext();
             if (currentContext && currentContext.operation) {
-                const uniqueParentId = currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.';
                 request['setHeader'](RequestResponseHeaders.requestIdHeader, uniqueParentId);
                 // Also set legacy headers
                 request['setHeader'](RequestResponseHeaders.parentIdHeader, currentContext.operation.id);
                 request['setHeader'](RequestResponseHeaders.rootIdHeader, uniqueParentId);
 
-                if (currentContext.operation.correlationContextHeader) {
-                    request['setHeader'](RequestResponseHeaders.correlationContextHeader, currentContext.operation.correlationContextHeader);
+                const correlationContextHeader = currentContext.customProperties.serializeToHeader();
+                if (correlationContextHeader) {
+                    request['setHeader'](RequestResponseHeaders.correlationContextHeader, correlationContextHeader);
                 }
             }
         }
@@ -129,12 +131,12 @@ class AutoCollectClientRequests {
             request.on('response', (response: http.ClientResponse) => {
                 requestParser.onResponse(response, properties);
                 var context : { [name: string]: any; } = { "http.RequestOptions": requestOptions, "http.ClientRequest": request, "http.ClientResponse": response };
-                client.track(requestParser.getDependencyData(), null, context);
+                client.track(requestParser.getDependencyData(uniqueParentId), null, context);
             });
             request.on('error', (e: Error) => {
                 requestParser.onError(e, properties);
                 var context : { [name: string]: any; } = { "http.RequestOptions": requestOptions, "http.ClientRequest": request, "Error": e };
-                client.track(requestParser.getDependencyData(), null, context);
+                client.track(requestParser.getDependencyData(uniqueParentId), null, context);
             });
         }
     }
--- a/AutoCollection/ClientRequests.ts
+++ b/AutoCollection/ClientRequests.ts
@@ -7,7 +7,7 @@ import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import ClientRequestParser = require("./ClientRequestParser");
-import { CorrelationContextManager, CorrelationContext } from "./CorrelationContextManager";
+import { CorrelationContextManager, CorrelationContext, PrivateCustomProperties } from "./CorrelationContextManager";
 
 import {enable as enableMongodb} from "./diagnostic-channel/mongodb.sub";
 import {enable as enableMysql} from "./diagnostic-channel/mysql.sub";
@@ -92,7 +92,7 @@ class AutoCollectClientRequests {
         let requestParser = new ClientRequestParser(requestOptions, request);
 
         const currentContext = CorrelationContextManager.getCurrentContext();
-        const uniqueParentId = currentContext && currentContext.operation && (currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.');
+        const uniqueRequestId = currentContext && currentContext.operation && (currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.');
 
         // Add the source correlationId to the request headers, if a value was not already provided.
         // The getHeader/setHeader methods aren't available on very old Node versions, and
@@ -114,12 +114,12 @@ class AutoCollectClientRequests {
             }
 
             if (currentContext && currentContext.operation) {
-                request['setHeader'](RequestResponseHeaders.requestIdHeader, uniqueParentId);
+                request['setHeader'](RequestResponseHeaders.requestIdHeader, uniqueRequestId);
                 // Also set legacy headers
                 request['setHeader'](RequestResponseHeaders.parentIdHeader, currentContext.operation.id);
-                request['setHeader'](RequestResponseHeaders.rootIdHeader, uniqueParentId);
+                request['setHeader'](RequestResponseHeaders.rootIdHeader, uniqueRequestId);
 
-                const correlationContextHeader = currentContext.customProperties.serializeToHeader();
+                const correlationContextHeader = (<PrivateCustomProperties>currentContext.customProperties).serializeToHeader();
                 if (correlationContextHeader) {
                     request['setHeader'](RequestResponseHeaders.correlationContextHeader, correlationContextHeader);
                 }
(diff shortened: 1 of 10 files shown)


AutoCollection/ClientRequests.ts
```diff
@@ -20,6 +20,8 @@ class AutoCollectClientRequests {
 
     public static INSTANCE: AutoCollectClientRequests;
 
+    private static requestNumber = 1;
```
KamilSzostak: Do you ever reset this counter? It is possible that `uniqueParentId` will not be unique. For instance, if Node server restarts and the counter goes to 1, or if there are more than one Node servers and each has a different requestNumber. Will it cause any problems?

MSLaguana: The parentId is a hierarchical Id of the form `|root.suffix.` or `|freshRoot.`, and `suffix` and `freshRoot` are randomly generated for this particular request. Assuming that `parentId` is unique for this particular incoming request (which is reasonable, a collision of suffix or newly created GUID is unlikely) the `uniqueParentId` will be unique for each outgoing request unless `requestNumber` is ever reset. If the node server restarts or there are multiple servers, they should all be generating different `parentId`s

AutoCollection/CorrelationContextManager.ts
```diff
@@ -35,16 +36,16 @@ export class CorrelationContextManager {
...
         parentId = parentId || operationId;
         
         if (this.enabled) {
             return {
                 operation: {
                     name: operationName,
                     id: operationId,
-                    parentId: parentId
+                    parentId: parentId,
+                    correlationContextHeader: correlationContextHeader
```
OsvaldoRosado: Why is this necessary here? If possible, I would like to keep this operation object free of properties not present on https://github.com/Microsoft/ApplicationInsights-dotnet/blob/de488729c28b721a758e0024ea2d4f79909ff797/src/Core/Managed/Shared/Extensibility/Implementation/OperationContext.cs so that the API surface is somewhat consistent

AutoCollection/ServerRequestParser.ts
```diff
@@ -169,6 +169,29 @@ class ServerRequestParser extends RequestParser {
...
+        this.rawHeaders = request.headers || (<any>request).rawHeaders;
+        this.userAgent = request.headers && request.headers["user-agent"];
+        this.sourceCorrelationId = Util.getCorrelationContextTarget(request, RequestResponseHeaders.requestContextSourceKey);
+
+        if (request.headers) {
+            this.correlationContext = request.headers[RequestResponseHeaders.correlationContextHeader];
+
+            if (request.headers[RequestResponseHeaders.requestIdHeader]) {
+                this.parentId = request.headers[RequestResponseHeaders.requestIdHeader];
+                this.requestId = CorrelationIdManager.generateRequestId(this.parentId);
```
OsvaldoRosado: Why is the manually supplied parentId ignored in the new format?

OsvaldoRosado: I'm also not sure this approach is correct. I would think our request ids for putting into AI telemetry envelopes do not change format but only the id sent out in the header does. This replaces the id for the envelope with the hierarchical id if I'm reading correctly. 

@SergeyKanzhelev What's the correct approach here?

AutoCollection/ServerRequestParser.ts
```diff
@@ -110,6 +106,10 @@ class ServerRequestParser extends RequestParser {
         return this.requestId;
     }
 
+    public getCorrelationContext() {
+        return this.correlationContext;
```
OsvaldoRosado: This is fairly confusing naming, given that this is not a `CorrelationContext` object

AutoCollection/CorrelationContextManager.ts
MSLaguana: My understanding is that if an incoming request has a `Correlation-Context` header, then we must forward that value on in any outgoing requests. Where would you suggest that we store the value, if not here with the other values that have the same (whole-request) lifetime?

OsvaldoRosado: Is that the property bag header? It should fill `CorrelationContext.customProperties`. That was the placeholder object intended for this. The idea is to deserialize it into a dictionary so that its user-editable, and then re-serialize it when outgoing

OsvaldoRosado: We'll have to update the comment there. Today it says
```
Properties here can be exposed in a future SDK release via outgoing HTTP headers for correlating data cross-component.
```

AutoCollection/ServerRequestParser.ts
MSLaguana: I'm not sure that I understand the purpose of the code-supplied parentId. From what I understand of the spec, if an incoming request specifies a `Request-Id`, then this request is part of a larger tracked operation, and we should take that Id and extend it appropriately, which is what I do here. What I don't understand is how to combine an incoming `Request-Id` with an ambient/passed-in `Request-Id`. If the incoming request does not specify a `Request-Id` then we will go into the `else` case which checks for legacy headers, and if they are not found it will use the passed-in parentId.

AutoCollection/ServerRequestParser.ts
MSLaguana: Agreed. I'm open to suggestions; perhaps `getCorrelationContextHeader`? It is "whatever value is specified in the `correlation-context` header, if any"

AutoCollection/CorrelationContextManager.ts
MSLaguana: According to my reading of https://github.com/lmolkova/correlation/blob/master/http_protocol_proposal_v1.md#correlation-context the `Correlation-Context` header should only be writable in the originating service, and should be read-only in all others. If that's the case, does deserializing into a dictionary still make sense? Although I see at the end of that section it does also talk about adding to it, which is a little confusing.

OsvaldoRosado: It has to be considered fully read-write in at least one case (where you are the originating service). Even when only read deserializing makes sense to actually use the values (like for the suggested A-B testing scenario) I also suspect the read-only applies to existing keys but does not prevent adding new ones. I think this is what is meant by `MUST support adding properties to Correlation-Context.` @lmolkova could verify the intent here.

AutoCollection/ServerRequestParser.ts
OsvaldoRosado: This could get removed entirely since it should be CorrelationContext.customProperties

MSLaguana: Right now this getter is how we get access to the value to put it in the custom properties; all the header parsing happens in this class.

OsvaldoRosado: Ah yes, sorry I thought this class had access to the correlation context and could fill it directly. It would make sense since that is part of the header parsing

AutoCollection/ServerRequestParser.ts
OsvaldoRosado: The manually passed in parent id was to allow filling requestId from `correlationContext.operation.parentId` when manually calling trackRequestSync. TrackRequestSync is too late to get the value from headers (since it should be the last call in the request handler). If you do, you won't be correlated with any of the dependency telemetry that used a different parentId.

OsvaldoRosado: I'm also noticing the parameter is renamed here from requestid to parentid. It's a bit confusing to do that as the parameter should set requestid, we just happen to set the requestid with the value of parentid (to resolve the situation above).

AutoCollection/ClientRequests.ts
lmolkova: you should not assume that parentId is indeed unique. It could not be unique because of
- bug on the sender side
- redirects (headers are copied as it on redirects)
- some proxy that retries requests with the same headers

The [spec suggests to append](https://github.com/lmolkova/correlation/blob/master/hierarchical_request_id.md#incoming-request) small e.g. 32-bits random integer (hex encoded to reduce length) for the parent id

AutoCollection/CorrelationContextManager.ts
lmolkova: The spec says that the very first service may add properties to the Correlation-Context.
Any further service MUST consider properties as read only i.e. MUST NOT modify existing properties. And SHOULD NOT add new properties.
This is a guidance and AppInsights SDK does not know which service is the first one, so you should allow adding properties to the Correlation-Context.

If the requirement to keep properties came from upstream as read-only is too strict and hard to implement; and you would rather use simple dictionary - go ahead, make them mutable.

AutoCollection/ClientRequests.ts
lmolkova: Just realized that the implementation does exactly what spec suggests: for the incoming requests it appends random int suffix, for the outgoing requests, it appends monotonically incremented number.  So the outgoing request will have unique Id.

AutoCollection/ServerRequestParser.ts
```diff
@@ -169,6 +169,29 @@ class ServerRequestParser extends RequestParser {
...
+                this.requestId = CorrelationIdManager.generateRequestId(this.parentId);
+                this.correlationContext = request.headers[RequestResponseHeaders.correlationContextHeader];
+            } else {
+                // Legacy fallback
+                const rootId = request.headers[RequestResponseHeaders.rootIdHeader];
+                this.parentId = request.headers[RequestResponseHeaders.parentIdHeader];
+                this.requestId = CorrelationIdManager.generateRequestId(rootId || this.parentId || parentId);
+                this.correlationContext = null;
+            }
+            this.operationId = CorrelationIdManager.getRootId(this.requestId);
```
lmolkova: what if there were nor Request-Id nor legacy-headers?

MSLaguana: If no headers are present, then `this.parentId` will be null, but `this.requestId` will be `CorrelationIdManager.generateRequestId(parentId)` which will generate a hierarchical ID from the passed-in identifier (if any) or from nothing, so `this.requestId` should always be set.

lmolkova: right, I thought there is the else-if branch for legacy headers. never mind

AutoCollection/CorrelationContextManager.ts
MSLaguana: I think we can't deserialize the header into a map, since the spec explicitly says that keys may be duplicated, and we must not aggregate or deduplicate values. I'd prefer to keep this as an opaque string (that we potentially add `, key=value` to the end of), but I'd also be reasonably happy with converting it to an array of `key=value` strings, splitting on `', '` and potentially appending more in future.

OsvaldoRosado: There's no way to reconcile the API surface here while keeping it an opaque string, unless we ignore the spec requirement to support adding properties and make the entire functionality private/internal to the SDK (which is maybe fine? but it would be great to get this working if the only issue is on how the end user interacts with it).

We have a key-value store already in the public API surface designated for this use, so if we keep this as an opaque string and publicly make it available on the correlation context we now have an easy to use storage that says its for correlation but doesn't work, and a complicated to use one that does work but you have to handle formatting yourself and know you cant use any illegal characters for headers.

We can replace the already designated object perhaps with one with custom get/setters. Avoiding the edge cases about duplication shouldn't be too difficult, this was already handled in .NET https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/c64fc1d28773c229b4351e6acf7dc6788229dd31/Src/Common/HeadersUtilities.cs

MSLaguana: I don't see how .NET is handling duplicates in that code; they essentially aggregate all duplicates together and ignore all but the first value (and when they write a value, they erase all previous instances of the key and write the new key/value to the end). That doesn't seem like it matches the spec at all.

OsvaldoRosado: That is strange isn't it.. The get returning only the first value seems perfectly fine (our SDK doesn't allow you to create duplicate keys, so if you have them you must have special code to read them. I don't think this is really an aggregation, nothing is being combined) That set seems totally broken though.

To strictly follow the spec (which as @lmolkova mentioned isn't totally necessary if it is too difficult) I think setting would instead work on a different data set that only you can initialize (and cannot conflict with any values that were prepopulated into the context), and before sending the header this different data set is appended on to the header string

MSLaguana:
I believe I have addressed these issues now.

MSLaguana:
Looks like one of the tests failed, but only on one of the node versions. It seems that `client.sendPendingData` invoked its callback twice? Nothing I touched should impact that though.

AutoCollection/ClientRequests.ts
```diff
@@ -91,6 +91,9 @@ class AutoCollectClientRequests {
 
         let requestParser = new ClientRequestParser(requestOptions, request);
 
+        const currentContext = CorrelationContextManager.getCurrentContext();
+        const uniqueParentId = currentContext && currentContext.operation && (currentContext.operation.parentId + AutoCollectClientRequests.requestNumber++ + '.');
```
OsvaldoRosado: Nit: Isn't this a request id not a parent id?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -8,13 +8,17 @@ export interface CorrelationContext {
...
 
     /** Do not store sensitive information here. 
      *  Properties here can be exposed in a future SDK release via outgoing HTTP headers for correlating data cross-component.
      */
-    customProperties: { [id: string]: string };
+    customProperties: { 
+        addHeaderData(header: string): void;
+        getProperty(prop: string): string;
+        setProperty(prop: string, val: string): void;
+        serializeToHeader(): string;
```
OsvaldoRosado: I don't think anything other than get and set should be part of the public API

AutoCollection/CorrelationContextManager.ts
```diff
@@ -8,13 +8,17 @@ export interface CorrelationContext {
         name: string;
         id: string;
         parentId: string; // Always used for dependencies, may be ignored in favor of incoming headers for requests
-        correlationContextHeader?: string;
     };
 
     /** Do not store sensitive information here. 
      *  Properties here can be exposed in a future SDK release via outgoing HTTP headers for correlating data cross-component.
```
OsvaldoRosado: This comment should be updated since we now support this :smile:

OsvaldoRosado:
@MSLaguana Looks like a re-run of the unit tests removed the failure. Might have already been a flaky test

AutoCollection/ClientRequests.ts
MSLaguana: One small note here: I've used a single counter in the whole process, while the .NET sdk seems to use one counter for each request. This means that the IDs could get large in a long-lived process.

Issue #264, opened 2017-06-05
Invalid method signature leads to implicit 'any' type


rstropek:
I found three method signatures that seem to be incorrect:

* [Channel.ts](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Channel.ts#L89)
* [Client.ts](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Client.ts#L239)
* [Sender.ts](https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Sender.ts#L42)

In all three locations, a parameter `callback` is used with the following type declaration: `(string) => void`. It looks like the author wants the callback to receive a parameter of type `string`. However, this TypeScript syntax leads to a parameter called `string` with type `any`. In my understanding, the callback's type should be something like `(msg: string) => void`.

The reason why this is important is TypeScript's TS7006 error: *Parameter '...' implicitly has an 'any' type'*. I want to compile my Node.js application with the TypeScript compiler setting `noImplicitAny`. In that case, the method signature mentioned above leads to an error and I cannot use `noImplicitAny`.

OsvaldoRosado:
Great catch! Marking this for the next release.

Thanks!

OsvaldoRosado:
Fixed with #272

PR #272, opened 2017-06-12
Clear all implicit any usages

OsvaldoRosado:
Addresses #264 and ensures we don't regress by enforcing this at compile time

@MSLaguana Part of this is a restructure of the CustomProperties object you wrote. There should be no actual code change just placing it into a real class. Could you give a quick look to make sure this is still the same? The previous "pseudo-class" structure wasn't working well with the enforcement of no implicit any added here.

### DIFF
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -251,7 +251,7 @@ export class CorrelationContextManager {
     }
 }
 
-class CustomPropertiesImpl implements CustomProperties {
+class CustomPropertiesImpl implements PrivateCustomProperties {
     private static bannedCharacters = /[,=]/;
     private props: {key: string, value:string}[] = [];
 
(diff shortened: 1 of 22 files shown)


AutoCollection/CorrelationContextManager.ts
```diff
@@ -299,3 +250,53 @@ export class CorrelationContextManager {
         global.Error = <any>AppInsightsAsyncCorrelatedErrorWrapper;
     }
 }
+
+class CustomPropertiesImpl implements CustomProperties {
```
MSLaguana: This should implement `PrivateCustomProperties`, or that interface should be removed in favor of this class (which would require some minor changes in the other classes which reference it currently)

OsvaldoRosado: Didn't realize that interface was there. Good catch!

Issue #270, opened 2017-06-09
Fractional millisecond values are not sanitized or serialized properly on trackDependency


itilitymaarten:
```
context.log(`Dependency tracked: ${dependencyName} ${prop} ${duration[0] * 1000 + duration[1] / (1000 * 1000)} false`);
insightsClient.trackDependency(dependencyName, prop, duration[0] * 1000 + duration[1] / (1000 * 1000), false);
```

I use the code above to log a dependency call. The duration array is a result from `process.hrtime()` (the first element contains seconds, the second element contains nanoseconds which couldn't fit into seconds precision). I know the function is called, because the context.log above it is logged (with proper arguments). However, the dependency call isn't showing up in Application insights, while the tracing and everything else I'm logging through the SDK is.

Any suggestions on how to debug this problem or what might be causing it?

OsvaldoRosado:
I can reproduce this problem. The following error is returned from the backend:
`Field 'duration' on type 'RemoteDependencyData' is not a valid duration string. Expected: dd.hh:mm:ss.fffffff format not exceeding 1000 days, Actual: 21.17:03:44.372.2452609539032`

You can workaround this issue by not giving non-integer millisecond values to trackDependency. For example, the following works:
```
var duration = process.hrtime();
client.trackDependency('testDep', 'prop', Math.round(duration[0] * 1000 + duration[1] / (1000 * 1000)), false);
```

OsvaldoRosado:
Fixed with #274

PR #274, opened 2017-06-13
Fix fractional millisecond conversion

OsvaldoRosado:
Addresses #270 

@KamilSzostak perhaps JS SDK should use this as well instead of rounding to integer milliseconds?

This change adds support for a variable amount of subsecond precision (min 3 digits, max 7).

### DIFF
--- a/Tests/Library/Util.tests.ts
+++ b/Tests/Library/Util.tests.ts
@@ -152,6 +152,7 @@ describe("Library/Util", () => {
             test(24 * 60 * 60 * 1000, "1.00:00:00.000", "hours overflow");
             test(11 * 3600000 + 11 * 60000 + 11111, "11:11:11.111", "all digits");
             test(5 * 86400000 + 13 * 3600000 + 9 * 60000 + 8 * 1000 + 789, "5.13:09:08.789", "all digits with days");
+            test(1001.505, "00:00:01.001505", "fractional milliseconds");
         });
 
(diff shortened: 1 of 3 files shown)


KamilSzostak:
JS SDK is using the `ms % 1000` approach. The code looks almost the same as in Node.js - https://github.com/Microsoft/ApplicationInsights-JS/blob/master/JavaScript/JavaScriptSDK/Util.ts#L447.

Let's make the same fix in JS SDK.

Tests/Library/Util.tests.ts
```diff
@@ -152,6 +152,7 @@ describe("Library/Util", () => {
             test(24 * 60 * 60 * 1000, "1.00:00:00.000", "hours overflow");
             test(11 * 3600000 + 11 * 60000 + 11111, "11:11:11.111", "all digits");
             test(5 * 86400000 + 13 * 3600000 + 9 * 60000 + 8 * 1000 + 789, "5.13:09:08.789", "all digits with days");
+            test(1001.505, "00:00:01.001505", "fractional milliseconds");
```
KamilSzostak: can you add extra unit tests to check `1001.5`, `1001.55` (not all precision digits) and `1001.5059` (too many digits, round up?)

OsvaldoRosado: Good idea. Added!

Issue #63, opened 2016-01-13
Exceptions are stored on disk even if offline mode is off and are never sent


bragma:
I've noticed this odd (probably wrong) behavior associated to the AutoCollection of Exceptions. In a brief: if "enableOfflineMode" if off, unhandled exceptions are not sent and pile up on disk.

Details follow:
If AutoCollection is enabled, all uncaught exceptions are trapped by "uncaughtException" event on the process. This causes the exception data to be sent to "handleCrash" method on Channel object with "isNodeCrashing" set to true. "handleCrash" sends the exception data to "saveOnCrash" and this method does not check if "_enableOfflineMode" is true. It just stores on disk.

If "_enableOfflineMode" if false, 2 (bad) things happen: Files are stored on disk even if this functionality is off. It may still make sense, since it's the only way to make unhandled exception collection work. But since "_enableOfflineMode" is false, files stored on disk are never sent.

So, ultimately:
1. Files will pile up forever on disk
2. Unhandled Exceptions will not be sent

Maybe it would be better to:
1. Not store exceptions to disk, since they will not be sent. It may happen that on first run offline mode is off, then switched on later. But its an odd scenario. If this is not the case, Exception files will be piled up forever.
2. Try to send exceptions (only) on next start even if offline mode is off.

What do you think?

southwood:
Hi @bragma, thanks for writing this up. From what you're saying and looking at the code it looks like you're assessment and this was probably a regression during one of the offlineMode changes. It should be noted that the crash handling is has been impacted by several of the last version changes of node so it's also possible that this is a version-specific issue.

I like your proposal. Separating crashes from the 'offlineMode' concept seems like the right thing to do. Feel free to make a pull request or ask @BogdanBe to triage.

bragma:
@southwood sadly I'm really busy right now. I'd happily pick up the PR but I think this problem deserves more care I can afford right now. :(

jsheely:
This should address this issue. https://github.com/Microsoft/ApplicationInsights-node.js/pull/83

Though I believe more thought will need to be put into how offlineMode works

PR #300, opened 2017-08-22
Checking if offline mode is enabled before reading from or writing to disk

AlexBulankou:
Fixing #63

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -54,7 +54,7 @@ class Sender {
             path: parsedUrl.pathname,
             method: "POST",
             withCredentials: false,
-            headers: <{[key: string] : string}>{
+            headers: <{ [key: string]: string }>{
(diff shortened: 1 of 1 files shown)


Library/Sender.ts
```diff
@@ -137,30 +137,32 @@ class Sender {
     }
 
     public saveOnCrash(payload: string) {
-        this._storeToDiskSync(payload);
+        if (this._enableOfflineMode) {
```
AlexBulankou: This is the only change, everything else is formatting/typos

Dmitry-Matveev: Do you need similar check on reading from the storage or it was already there? As per PR title, it looks like we wanted to add both.

AlexBulankou: Yes, we are already checking on L97 if enableOfflineMode is true, before starting to transmit stored files.

Issue #302, opened 2017-08-23
Refactor track* API


AlexBulankou:
Before we ship stable SDK I would like to refactor track API:

1. To make track* functions more consistent between themselves. 
2. To allow specifying timestamp
3. To address some reported bugs such as inability to specify result code for dependencies #286 and not being able to track requests without request object #148 
4. To make names consistent with property names in Portal and Analytics
5. To make API consistent with [.NET API](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/f3f43d0f2159ac34023af8e8c7b5afe2da782400/src/Microsoft.ApplicationInsights/Managed/Shared/TelemetryClient.cs)
6. To reduce the number of positional argument to the ones required, while keeping the others in TrackOptions bag.

This is the proposal:
```JavaScript

interface TrackOptions
{
    timestamp?: Date;
    properties?: { [key: string]: string; }
    measurements?:{ [key: string]: number; }
    contextObjects?: {[name: string]: any;}
    tagOverrides?: { [key: string]: string; }
}

void trackDependency(
     dependencyTypeName: string,
     target: string,
     dependencyName: string,
     data: string,
     duration: number,
     resultCode: string
     success: bool
     trackOptions: TrackOptions);

void trackEvent(
     eventName: string,
     trackOptions: TrackOptions);

void trackTrace(
     message: string,
     severityLevel: SeverityLevel
     trackOptions: TrackOptions);

void trackMetric(
     metricName: string,
     metricValue: number
     trackOptions: TrackOptions);

void trackException(
     exception: Error,
     trackOptions: TrackOptions);

void trackRequest(
     name: string,
     duration: number,
     resultCode: string
     success: bool
     trackOptions: TrackOptions);
```

@OsvaldoRosado @SergeyKanzhelev @KamilSzostak

OsvaldoRosado:
What does `trackRequest` do in this proposal? Currently our `trackRequest` doesn't allow creation of a manual request - only autocreation from `req` and `res` objects. 

I think I would like to see `trackRequest` be purely manual, and we can have a separate extensions/helpers API that can perform the current actions of `trackRequest`/`trackRequestSync`/`trackDependencyRequest`.

Keeping these helpers is important as they enable cross-component correlation to work even in manually instrumented applications

OsvaldoRosado:
Not part of track*, but we should also refactor the root API methods. For example, `getClient` is endlessly confusing.

[...]

PR #304, opened 2017-08-26
Refactor track* API PR

AlexBulankou:
As discussed in #302.
I'll be adding a few more unit tests to cover trackRequest function.


AutoCollection/diagnostic-channel/mysql.sub.ts
```diff
@@ -17,11 +17,16 @@ export const subscriber = (event: IStandardEvent<mysql.IMysqlData>) => {
...
-                dbName,
-                sqlString,
-                event.data.duration | 0,
-                success,
-                "mysql");
+            {
+                target: dbName,
+                data: sqlString,
+                name: sqlString,
+                duration: event.data.duration | 0,
```
OsvaldoRosado: Is this guard necessary? `trackDependency` should defend against NaN duration

AutoCollection/Exceptions.ts
```diff
@@ -36,9 +36,8 @@ class AutoCollectExceptions {
             var self = this;
             if (!this._exceptionListenerHandle) {
                 var handle = (reThrow: boolean, error: Error) => {
-                    var data = AutoCollectExceptions.getExceptionData(error, false);
-                    var envelope = this._client.getEnvelope(data);
-                    this._client.channel.handleCrash(envelope);
+                    this._client.trackException({exception: error});
+                    this._client.flush();
```
OsvaldoRosado: `flush` isn't equal here, right? It's not using the `isNodeCrashing` parameter to `channel.triggerSend`

Library/Client.ts
```diff
@@ -38,269 +45,99 @@ class Client {
...
      * Generic track method for all telemetry types
      * @param data the telemetry to send
-     * @param tagOverrides the context tags to use for this telemetry which overwrite default context values
+     * @param telemetryType specify the type of telemetry you are tracking from the list of Contracts.DataTypes
      */
-    public track(
-        data: Contracts.Data<Contracts.Domain>,
-        tagOverrides?: { [key: string]: string; },
-        contextObjects?: { [name: string]: any; }) {
+    public track(telemetry: Telemetry, telemetryType: string) {
```
OsvaldoRosado: Do we have a better type to restrict here than `string`?

Library/DependencyTelemetry.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * True if remote call was successful, false otherwise
+     */
+    success: boolean;
+
+    /**
+     * Unique dependency identifier used for correlation. Leave it unspecified when calling trackDependency manually
+     */
+    dependencyId?: string;
```
OsvaldoRosado: If this is to be unused, should we remove it from the public type and internally use a type that extends this?

Library/EnvelopeFactory.ts
```diff
@@ -0,0 +1,329 @@
...
+     * @param config Client configuration
+     */
+    public static createEnvelope(
+        telemetry: Telemetry,
+        telemetryType: string,
+        commonProperties?: { [key: string]: string; },
+        context?: Context,
+        config?: Config): Contracts.Envelope {
+
+        var data:
```
OsvaldoRosado: Is it necessary to enumerate all event types here?

Library/EventTelemetry.ts
```diff
@@ -0,0 +1,19 @@
+import Telemetry = require("./Telemetry")
```
OsvaldoRosado: What do you think of moving the Telemetry* classes into their own folder (either within Library or somewhere else)?

Library/NodeClient.ts
```diff
@@ -0,0 +1,60 @@
...
+            Logging.warn("trackNodeHttpRequestSync requires NodeHttpRequestTelemetry object with request, response and duration specified.");
+        }
+    }
+
+    /**
+     * Log RequestTelemetry from HTTP request and response. This method will `follow` the request to completion.
+     * Use trackNodeHttpRequestSync function to log telemetry immediately without waiting for request completion
+     * @param telemetry Object encapsulating incoming request and response information
+     */
+    public trackNodeHttpRequest(telemetry: NodeHttpRequestTelemetry) {
```
OsvaldoRosado: What's your opinion on simply overriding `trackRequest` here to overload? If input is not of type `NodeHttpRequestTelemetry` call into the base class. This allows `trackRequest` to be used for both manual and the preferred one of the two node shortcut methods. Same question for the dependency shortcut as well

applicationinsights.ts
```diff
@@ -48,7 +48,7 @@ class ApplicationInsights {
      * @returns {ApplicationInsights/Client} a new client
      */
     public static getClient(instrumentationKey?: string) {
```
OsvaldoRosado: It would be really great to rename this :smile:

Library/Client.ts
OsvaldoRosado: Would it help to make `Contracts.DataTypes` an enum?

AutoCollection/diagnostic-channel/mysql.sub.ts
AlexBulankou: Removed the guard. Thinking about it, it is not just unnecessary, but incorrect, because duration could be missing  in case of incomplete dependency that is being logged and we shouldn't be treating every empty value as 0. We are doing it now (for all SDKs) but should address it separately.

AutoCollection/Exceptions.ts
AlexBulankou: Addressed, changed flush signature to accommodate the scenario.

SergeyKanzhelev:
LGTM

Library/Client.ts
AlexBulankou: Done

Library/EnvelopeFactory.ts
AlexBulankou: Shortened.

Library/EventTelemetry.ts
AlexBulankou: Done.

Library/NodeClient.ts
AlexBulankou: I am not sure I like the ambiguity it creates. If I know about AppInsights, track* is very simply to understand and I can see that extra methods are node-specific. If we use overloads, then I have no option but to learn about these additional overloads which adds complexity, so it is not that simple anymore.
Also I don't like the idea of checking the types for presence of specific fields, since these are interfaces so simple instanceof wouldn't work.

applicationinsights.ts
AlexBulankou: Ashamed to admit, but I don't get the instinctive gag reflex when typing `getClient` :) However I would happy to rename, any suggestions? Semantically `createClient` makes more sense, but what did you have in mind? Just `.client`?

Library/NodeClient.ts
OsvaldoRosado: Fair enough. I'm okay with having these separate methods.

applicationinsights.ts
OsvaldoRosado: We have `.client` already, which does what `getClient` should do - return the current client. `createClient` or `newClient` would be a lot better. Alternatively we can just expose the `Client` class and users can use `new Client()`

AlexBulankou: Got it, will change.

Library/TelemetryTypes/TelemetryType.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Telemetry type enumeration
+ */
+enum TelemetryType {
```
OsvaldoRosado: FYI - TypeScript 2.4+ allows string-based enums. That could avoid some of the map-type syntax you use in the envelope factory.

Library/DependencyTelemetry.ts
AlexBulankou: Done.

OsvaldoRosado:
Quick heads up. Just noticed this makes the README sections on manual instrumentation out of date

Library/TelemetryTypes/TelemetryType.ts
AlexBulankou: Updated and changed.

PR #305, opened 2017-08-28
Adding ikey as part of storage folder path to fix #202

### LINKED ISSUE #202: Offline mode does not persist endpoint with file, leading to data being sent to wrong endpoints
Currently offline mode saves all files into %temp%\appInsights-node folder.

We have two clients each configured with a different URL endpoint.  The URL endpoint is not saved in the offline file.

When AppInsights goes back online, the client sends the first file in the offline folder to whichever URL endpoint the client happened to be configured with. This can be different than the original endpoint intended for the data.

If this different endpoint rejects the data, then AI will re-save the file and try later.  However, if the endpoint accepts the data, then AI will delete this file and fail to deliver the offline packet to the correct endpoint.

We also see a problem with all Node.JS AppInsights apps sharing the same folder.  If an endpoint in an active client rejects an offline packet stored by a different app, then this can stop the chain of re-sending packets from this client, event though that client's packets would successfully be sent.

.NET AppInsights in contrast saves entire HTTP request in an applicationKey-specific folder when it is offline.  This solves the problem both about URL and about delivery of offline packets.

AlexBulankou:
* Fixing #202

### DIFF
--- a/Library/Client.ts
+++ b/Library/Client.ts
@@ -43,7 +43,7 @@ class Client {
         this.context = new Context();
         this.commonProperties = {};
 
-        var sender = new Sender(() => config.endpointUrl);
+        var sender = new Sender(this.config);
         this.channel = new Channel(() => config.disableAppInsights, () => config.maxBatchSize, () => config.maxBatchIntervalMs, sender);
     }
 
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -7,22 +7,24 @@ import url = require("url");
 import zlib = require("zlib");
 
 import Logging = require("./Logging");
+import Config = require("./Config")
 import AutoCollectClientRequests = require("../AutoCollection/ClientRequests");
 
 class Sender {
     private static TAG = "Sender";
     // the amount of time the SDK will wait between resending cached data, this buffer is to avoid any throtelling from the service side
     public static WAIT_BETWEEN_RESEND = 60 * 1000;
-    public static TEMPDIR: string = "appInsights-node";
+    public static TEMPDIR_PREFIX: string = "appInsights-node";
 
-    private _getUrl: () => string;
+    private _config: Config;
+    private _storageDirectory: string;
     private _onSuccess: (response: string) => void;
     private _onError: (error: Error) => void;
     private _enableOfflineMode: boolean;
     protected _resendInterval: number;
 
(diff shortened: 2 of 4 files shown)


Library/Client.ts
```diff
@@ -43,7 +43,7 @@ class Client {
         this.context = new Context();
         this.commonProperties = {};
 
-        var sender = new Sender(() => config.endpointUrl);
+        var sender = new Sender(this.config);
```
OsvaldoRosado: Are there side effects to removing the input function? I imagine the inputs to this and channel were all wrapped inside functions for some reason

AlexBulankou: I think they were wrapped, so that sender always has access to the the latest (possibly updated) endpointUrl and doesn't cache it as local variable. This delayed retrieval is still preserved as we are only getting url from inside `send()` method.

OsvaldoRosado: Couldn't ikey still change however?

AlexBulankou:
Yes, and I am not caching it, every function that reads or writes constructs the path. Or I misunderstood the question?

________________________________

OsvaldoRosado:
@AlexBulankou I think you understood the question right. I just wanted to make sure the config object reference can never go stale. If it’s all good then no concerns from me

Library/Sender.ts
```diff
@@ -160,7 +162,7 @@ class Sender {
     private _storeToDisk(payload: any) {
 
         //ensure directory is created
-        var directory = path.join(os.tmpdir(), Sender.TEMPDIR);
+        var directory = path.join(os.tmpdir(), Sender.TEMPDIR_PREFIX + this._config.instrumentationKey);
```
SergeyKanzhelev: make sure it's not an information disclosure. Multiple processes may have access to temp folder which is sometimes shared among multiple processes. Maybe take a hash of it?

AlexBulankou: To clarify, are you suggesting hashing the ikey in the folder name, so that other processes may not read it? We could do so, however even if we take the hash of the folder name, thus hiding the ikey, what's preventing the other process from parsing the data in the folder, where ikey is part of every envelope? - in fact they can do so today.

SergeyKanzhelev: ok. good point. Perhaps allow to override the path so one worried about privacy and tampering can configure it to ACL-ed directory.

AlexBulankou: Added #307

Issue #311, opened 2017-09-01
Set up of auto-correlation for http.createServer() does not cover all cases


0x7f:
I was investigating an issue where the correlation context got lost in an [hapi](https://hapijs.com/) server. After digging through the code, I first expected their [domain-wrapper](https://github.com/hapijs/hapi/blob/master/lib/protect.js), but actually this works as expected.

The real issue was the way hapi calls `http.createServer`. The auto-collection code [expects](https://github.com/Microsoft/ApplicationInsights-node.js/blob/v0.17.2/AutoCollection/ServerRequests.ts#L45-L70) it being called like this (at least in latest release v0.17.2):
```
const handler = (req, res) => {...}
const server = http.createServer(handler);
```
but hapi does something like [this](https://github.com/hapijs/hapi/blob/v16.5.2/lib/connection.js#L89-L90):
```
const handler = (req, res) => {...}
const server = http.createServer();
server.on("request", handler);
```
which was introduced by [this patch](https://github.com/hapijs/hapi/commit/3b75e4f269fb4821a0d8abdc16180e503f73af6e) 3 years ago.

The most straight forward way to fix this would be to also patch the `server.on` function and hook into every event handler registering for the `request` (and `checkContinue`) event. What do you think?

OsvaldoRosado:
Great catch! Thanks for spotting this!

OsvaldoRosado:
Resolved with #312

PR #312, opened 2017-09-01
Fix AutoCollection of http.createServer()

0x7f:
Possible fix for https://github.com/Microsoft/ApplicationInsights-node.js/issues/311

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -72,13 +72,17 @@ class AutoCollectHttpRequests {
     private _initialize() {
         this._isInitialized = true;
 
-        var originalHttpServer = http.createServer;
-        http.createServer = (onRequest) => {
-            // todo: get a pointer to the server so the IP address can be read from server.address
-            return originalHttpServer((request:http.ServerRequest, response:http.ServerResponse) => {
+        const wrapOnRequestHandler: Function = (onRequest?: Function) => {
+            if (!onRequest) {
+                return undefined;
+            }
+            if (typeof onRequest !== 'function') {
+                throw new Error('onRequest handler must be a function');
+            }
+            return (request:http.ServerRequest, response:http.ServerResponse) => {
                 // Set up correlation context
-                var requestParser = new HttpRequestParser(request);
-                var correlationContext = this._generateCorrelationContext(requestParser);
+                const requestParser = new HttpRequestParser(request);
+                const correlationContext = this._generateCorrelationContext(requestParser);
 
                 CorrelationContextManager.runWithContext(correlationContext, () => {
                     if (this._isEnabled) {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -52,6 +52,6 @@
   "dependencies": {
     "diagnostic-channel": "0.2.0",
     "diagnostic-channel-publishers": "0.2.1",
-    "zone.js": "0.7.6"
+    "zone.js": "^0.8.17"
```
OsvaldoRosado: I would be wary of updating `zone.js` without other changes to the `AppInsightsAsyncCorrelatedErrorWrapper`. Last time I tried this, various versions of node gave errors related to restricted properties. Once the type error is fixed to allow the unit tests to continue, I would expect you'll get the same errors reported in unit tests for quite a few node versions.

We also restrict our dependencies to a specific version for support reasons, so the `^` should be removed

AutoCollection/HttpRequests.ts
```diff
@@ -90,26 +94,35 @@ class AutoCollectHttpRequests {
...
-        https.createServer = (options, onRequest) => {
-            return originalHttpsServer(options, (request:http.ServerRequest, response:http.ServerResponse) => {
-                // Set up correlation context
-                var requestParser = new HttpRequestParser(request);
-                var correlationContext = this._generateCorrelationContext(requestParser);
+        function wrapServerEventHandler(server) {
+            const originalOn = server.on.bind(server);
+            server.on = (eventType: string, eventHandler) => {
+                switch (eventType) {
+                    case 'request':
```
OsvaldoRosado: Is the actual implementation case-sensitive? If not, we should compare against `eventType.toLowerCase()`

AutoCollection/HttpRequests.ts
```diff
@@ -90,26 +94,35 @@ class AutoCollectHttpRequests {
...
 
-        var originalHttpsServer = https.createServer;
-        https.createServer = (options, onRequest) => {
-            return originalHttpsServer(options, (request:http.ServerRequest, response:http.ServerResponse) => {
-                // Set up correlation context
-                var requestParser = new HttpRequestParser(request);
-                var correlationContext = this._generateCorrelationContext(requestParser);
+        function wrapServerEventHandler(server) {
+            const originalOn = server.on.bind(server);
+            server.on = (eventType: string, eventHandler) => {
```
OsvaldoRosado: Our unit tests are flagging type errors around these for having implicit `any`. Can you use their concrete types?

OsvaldoRosado:
Thanks for this fix!

package.json
0x7f: Good points. Removed that commit from the PR.

AutoCollection/HttpRequests.ts
0x7f: Sorry for not checking the tests before pushing. Hope I catched all cases now.

AutoCollection/HttpRequests.ts
0x7f: These are the string mentioned in the official Node.js docs (https://nodejs.org/api/http.html#http_event_request) and as far as I know, they are kept as is inside the EventEmitter class. See: https://github.com/nodejs/node/blob/v8.4.0/lib/events.js#L232

AutoCollection/HttpRequests.ts
```diff
@@ -72,13 +72,17 @@ class AutoCollectHttpRequests {
...
 
-        var originalHttpServer = http.createServer;
-        http.createServer = (onRequest) => {
-            // todo: get a pointer to the server so the IP address can be read from server.address
-            return originalHttpServer((request:http.ServerRequest, response:http.ServerResponse) => {
+        const wrapOnRequestHandler: Function = (onRequest?: Function) => {
+            if (!onRequest) {
+                return undefined;
+            }
+            if (typeof onRequest !== 'function') {
```
OsvaldoRosado: This is different behavior than the non-wrapped function. In node 6.11 at least, I still get an object returned for `http.createServer(null, "abc")`

OsvaldoRosado: Same for `http.createServer(null, undefined)`

0x7f: To be honest, I do not fully get your point. The function `http.createServer` does not accept a second parameter (or at least is not using it). Only the `https` version is. And the function is [always returning a server object](https://github.com/nodejs/node/blob/v6.11.2/lib/http.js#L43-L45), no matter what you pass to it. Digging deeper and having a look at the Server constructor, it will only attach the listener function if it is truthy: https://github.com/nodejs/node/blob/v6.11.2/lib/_http_server.js#L233-L235 and then in the EventEmitter's addListener, it will [throw an exception when the listener is not a function](https://github.com/nodejs/node/blob/v6.11.2/lib/events.js#L215-L216).

So I think the patch mimics the behavior of the unpatched version.

Actually the current appinsights' master is having a different behavior I think. It always passes a valid function to `createServer` even though the user-provided argument might not be one: https://github.com/Microsoft/ApplicationInsights-node.js/blob/0131e35d29c268c8c99368a7a3a6435828186569/AutoCollection/HttpRequests.ts#L78-L91

0x7f: While digging through the Node.js code, I found that just patching 'EventEmitter.on' isn't enough, because we also have to wrap 'addListener' (https://github.com/nodejs/node/blob/v6.11.2/lib/events.js#L275-L279). And since the createServer is calling addListener eventually anyways, we do not have to patch that. Will update the PR accordingly.

OsvaldoRosado: > Actually the current appinsights' master is having a different behavior I think. It always passes a valid function to createServer even though the user-provided argument might not be one

Great catch!

My point before was over `if (typeof onRequest !== 'function')` in `wrapOnRequestHandler` potentially throwing when it wasn't expected, but I missed the truthiness check right above it that returns `undefined`, avoiding this problem :smile:

0x7f: Yeah, hope it covers all cases. At least it should for the current implementation of node.

@OsvaldoRosado i fixed the `.addListener()` and `.on()` patching and added a comment explaining why it was done the way it is: https://github.com/0x7f/ApplicationInsights-node.js/blob/b069fc2e5ba99ec95240187fe645535ef9192513/AutoCollection/HttpRequests.ts#L100-L126

wdyt?

OsvaldoRosado: LGTM. Thanks for this!

Issue #289, opened 2017-07-19
Telemetry data stop daily even with small request and reduced sampling


mehmetilker:
I am using Asp.Net App insights as well and there is no problem with it even with %100 sampling but one of my services (Node.js) is using this library but even thought I reduced sampling to %10 still dail cap finish too early. I am not sanding any extra request. And I cancelled collect request with the following settings.

Is there any other setting I need to do or is it possible node.js library is  storing extra information ?

`
appInsights.setup('xxxx')

		.setAutoDependencyCorrelation(true)
		.setAutoCollectRequests(false)
		.setAutoCollectPerformance(true)
		.setAutoCollectExceptions(true)
		.setAutoCollectDependencies(true)
		.start();
`

OsvaldoRosado:
This SDK sends more performance counters than the one for .NET. If you don't use this data, you can try turning off the performance autocollector. Does this app simply generate more events? Take a look in analytics in the various tables there to see the data you're sending to Application Insights. That way you can see the difference between what is being collected

mehmetilker:
Performance counter is on in .Net Sdk as well. I assume both .Net and Node Sdk should be identical.
Another thing is that  I see a warning saying that "You have reached your daily data limit. Increase your daily cap to fix this." on Overview tab but does not show and data.

I will try with turning off performance auto collector.

OsvaldoRosado:
> Performance counter is on in .Net Sdk as well. I assume both .Net and Node Sdk should be identical.

Ideally this would be the case, but it's not today. The Node SDK sends a different and larger set of counters than .NET. It may also send them at a higher rate, but I haven't looked into that yet. If it does, that's probably causing this issue.

mehmetilker:
I have turned performance collector off and still getting warning for daily cap.
On Azure App insight it shows 73 daily request and it seems it is enough to finish the daily cap. 

Is it possible that my config is wrong ?

I put the app insights related code just before app.listen section.
And doing nothing else.
I set sampling to %1 on azure as well.

`
var appInsights = require("applicationinsights");

	appInsights.setup('vvvvvvv')
		.setAutoDependencyCorrelation(true)
		.setAutoCollectRequests(false)
		.setAutoCollectPerformance(false)
		.setAutoCollectExceptions(true)
		.setAutoCollectDependencies(true)
		.start();
}
var server = app.listen(app.get('port'), function() {
`

gclifford:
[...]

Issue #282, opened 2017-06-27
Enhance npm documentation


nicolehaugen79:
Consider the following improvements to the package documentation to help appeal to node.js devs:

1.) The "Getting Started" section points the dev to a general [purpose walk-through](https://azure.microsoft.com/documentation/articles/app-insights-create-new-resource/) for configuring AI; instead, there is a node.js specific one available that seems like it would be better to point devs to: https://docs.microsoft.com/en-us/azure/application-insights/app-insights-nodejs.

2.) The Node.js MSFT doc mentioned above does a really good job of emphasizing the value props of using AI in the first paragraph - the npm package documentation should do the same.  

In particular, since node.js devs looking at this package may likely have their app hosted using a different cloud platform, an important point that the npm package documentation should emphasize is that "AI can be used for Node.js services hosted anywhere...even other public clouds".  This helps to make it clear that our technologies are pluggable.

3.) To help with discoverability, the package tags should be revisited - for example, "tracing" and "telemetry" may be other keywords that would be beneficial to include.

OsvaldoRosado:
Great observations! Thanks for the feedback!

Issue #269, opened 2017-06-08
Rearrange exports for typing


OsvaldoRosado:
Today with built-in types it's not convenient to get a reference to the Client or ApplicationInsights classes as a "type", nor anything in the Contracts object. We should re-think how we export objects to make this easier

Issue #155, opened 2016-12-13
Is there an example on how to override config properties?


SergeyKanzhelev:
I want to control these properties: `maxBatchSize` `maxBatchIntervalMs`. How can I override them?

gzepeda:
Hello Guys, 

Looks like what is needed here is a code snippet on how to do it. Is that correct?

OsvaldoRosado:
I think so

OsvaldoRosado:
There's one example of this now in the readme, for configuring sampling. We should consider documenting the entire configuration object

OsvaldoRosado:
Documentation on all config options is now in the README.

PR #314, opened 2017-09-06
Address 0.22 issues

OsvaldoRosado:
Addresses #269, #282, #289, #155. This also renames "offline mode" to "disk retry caching", adds it to documentation, and makes it enabled by default for the default client instance.

Of particular note is the changes for #269. Our current SDK with integrated typings has an issue where it's difficult to do things like this:
```var client: Client = appInsights.client```
This is because you don't have a direct reference to the `Client` class. Instead you must do `var client: typeof appInsights.client = appInsights.client`.

Similar issues are also present when using the contracts objects.

To solve this, this PR restructures our main applicationinsights module entry point, removing the class and placing the methods into a plain module and directly exporting `Client` and `Contracts`. This allows for direct reference to `Client`, for example, in the following ways:
```
// Previously unsupported es6 syntax
import { Client } from 'applicationinsights';
var client: Client = new Client(ikey);
-- or --
// Require syntax
var appInsights = require('applicationinsights');
var client: appInsights.Client = new appInsights.Client(ikey);
var defaultClient: appInsights.Client = appInsights.defaultClient
```

This removes the `createClient` API method in favor of the above approach and renames the default client from `appInsights.client` to `appInsights.defaultClient` for additional clarity.

For technical reasons for compatibility with this restructuring, the configuration builder approach when initializing this SDK is now inside it's own class, `Configuration`. The `setup` method returns this class, so the following snippet still works as is:
```
appInsights.setup("<instrumentation_key>")
    .setAutoDependencyCorrelation(true)
    .setAutoCollectRequests(true)
    .setAutoCollectPerformance(true)
    .setAutoCollectExceptions(true)
    .setAutoCollectDependencies(true)
    .setAutoCollectConsole(true)
    .start();
```
However, if you wish to configure options outside of this builder scenario, you can no longer call `setX` methods directly on `appInsights`. Instead, they are available directly from `appInsights.Configuration.setX`

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -8,7 +8,7 @@ import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import HttpRequestParser = require("./HttpRequestParser");
 import { CorrelationContextManager, CorrelationContext, PrivateCustomProperties } from "./CorrelationContextManager";
-import AutoColllectPerformance = require("./Performance");
+import AutoCollectPerformance = require("./Performance");
 
 class AutoCollectHttpRequests {
 
@@ -34,7 +34,7 @@ class AutoCollectHttpRequests {
         // Autocorrelation requires automatic monitoring of incoming server requests
         // Disabling autocollection but enabling autocorrelation will still enable
         // request monitoring but will not produce request events
-        if ((this._isAutoCorrelating || this._isEnabled) && !this._isInitialized) {
+        if ((this._isAutoCorrelating || this._isEnabled || AutoCollectPerformance.isEnabled()) && !this._isInitialized) {
             this.useAutoCorrelation(this._isAutoCorrelating);
             this._initialize();
         }
@@ -85,11 +85,19 @@ class AutoCollectHttpRequests {
                 const requestParser = new HttpRequestParser(request);
                 const correlationContext = this._generateCorrelationContext(requestParser);
 
+                // Note: Check for if correlation is enabled happens within this method.
+                // If not enabled, function will directly call the callback.
                 CorrelationContextManager.runWithContext(correlationContext, () => {
                     if (this._isEnabled) {
                         // Auto collect request
                         AutoCollectHttpRequests.trackRequest(this._client, request, response, null, requestParser);
                     }
+
+                    // Add this request to the performance counter
+                    // Note: Check for if perf counters are enabled happens within this method.
+                    // TODO: Refactor common bits between trackRequest and countRequest so they can
+                    // be used together, even when perf counters are on, and request tracking is off
+                    AutoCollectPerformance.countRequest(request, response);
 
                     if (typeof onRequest === "function") {
                         onRequest(request, response);
@@ -176,9 +184,6 @@ class AutoCollectHttpRequests {
             Logging.info("AutoCollectHttpRequests.trackRequest was called with invalid parameters: ", !request, !response, !client);
             return;
         }
-
-        // Register performance counter monitor
-        AutoColllectPerformance.countRequest(request, response);
 
         // store data about the request
         var correlationContext = CorrelationContextManager.getCurrentContext();
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -8,6 +8,7 @@ import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import HttpRequestParser = require("./HttpRequestParser");
 import { CorrelationContextManager, CorrelationContext, PrivateCustomProperties } from "./CorrelationContextManager";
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: AutoCollection/HttpRequests.ts [hunk 1]
```diff
@@ -8,6 +8,7 @@ import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import HttpRequestParser = require("./HttpRequestParser");
 import { CorrelationContextManager, CorrelationContext, PrivateCustomProperties } from "./CorrelationContextManager";
+import AutoColllectPerformance = require("./Performance");
```
Dmitry-Matveev: Minor: Looks like extra 'l' is in the name.

AutoCollection/HttpRequests.ts
```diff
@@ -176,6 +177,9 @@ class AutoCollectHttpRequests {
             return;
         }
 
+        // Register performance counter monitor
+        AutoColllectPerformance.countRequest(request, response);
```
Dmitry-Matveev: It looks like `countRequest()` will just calculate `start` time and then schedule continuation on end request to calculate the overall request duration and assign it to `lastRequestExecutionTime`. I think that we have this time calculated already at the moment we execute `TrackRequest()` later on - can we reuse that time and avoid double calculation of the same/similar value?

AutoCollection/Performance.ts
```diff
@@ -151,18 +130,18 @@ class AutoCollectPerformance {
                 totalIrq += irq;
 
                 var total = (user + sys + nice + idle + irq) || 1; // don"t let this be 0 since it is a divisor
-                this._client.trackMetric({name: name + "user", value: (user / total) * 100});
+                // this._client.trackMetric({name: name + "user", value: (user / total) * 100});
             }
 
-            var combinedName = "% total cpu";
+            // var combinedName = "% total cpu";
```
Dmitry-Matveev: Minor: I think we can remove unused/commented code and keep it in history if we need to enable it back.

README.md
```diff
@@ -203,7 +214,7 @@ let appInsights = require("applicationinsights");
 appInsights.setup("_ikey-A_").start();
 
 // track some events manually under another ikey
-let otherClient = appInsights.createClient("_ikey-B_");
```
Dmitry-Matveev: I think there could've been a reason it was not exposed as a constructor. Maybe, [something along those lines](https://iron9light.wordpress.com/2008/07/24/effective-java-even-c-1-creating-and-destroying-objects/) (Item 1). What do you think?
This is public API change, so I'd like to make sure we've considered options here before changing the contract.

applicationinsights.ts
```diff
@@ -4,285 +4,261 @@ import AutoCollectExceptions = require("./AutoCollection/Exceptions");
...
+     * @param enableWarningLogging if true, enables warning logging
+     * @returns {Configuration} this class
      */
-    public static enableVerboseLogging(enableWarningLogging = true) {
-        Logging.enableDebug = true;
-        Logging.disableWarnings = !enableWarningLogging;
-        return ApplicationInsights;
+    public static setInternalLogging(enableDebugLogging = false, enableWarningLogging = true) {
+        Logging.enableDebug = enableDebugLogging;
+        Logging.disableWarnings = enableWarningLogging;
```
Dmitry-Matveev: Is this behavior change intentional (`disable = enable` vs. `disable = !enable`)? It's doing the opposite now and I do not see it in the bug list being addressed by this PR.

AutoCollection/HttpRequests.ts
OsvaldoRosado: Good catch!

AutoCollection/HttpRequests.ts
OsvaldoRosado: Great idea! I'll have to make sure that that code is reachable even when automatic request collection is disabled, but if so, that makes total sense

README.md
OsvaldoRosado: The reason was that prior to my removing the main SDK class and replacing with a plain module, it wasn't actually possible to export it directly. This is already a change regardless because in the current SDK release this is called `getClient` rather than `createClient`.

I like exporting the class as it gives a direct access to what will likely become something like the base SDK in .NET, and by modifying your import statement to import just Client, you can use that base SDK directly with no other magic from the full SDK.

applicationinsights.ts
OsvaldoRosado: Yes it's intentional as before this was strangely placed as separate enable and disable functions. This is very much inconsistent with every other toggle that is one method that takes an on/off boolean.

OsvaldoRosado: Actually I see you mean on `disableWarnings` rather than the change to the method name. Looks like a bug! Good catch

AutoCollection/Performance.ts
OsvaldoRosado: Fair point, will remove

PR #303, opened 2017-08-24
Updated diagnostic-channel, added subscribers for winston and pg

jackhorton:
Fixes #256

diagnostic-channel-publishers@0.2.0 added new publishers for pg@6.x, pg-pool@1.x, and winston@2.x, so I have added subscribers for those as well. Finally, I noticed that mongoCore was not being brought in and enabled by default, so I fixed that as well.

I checked that it built and passed all tests as well.

PR #319, opened 2017-09-11
Add calls to enable winston and postgres

jackhorton:
I realized that in my last pull request (https://github.com/Microsoft/ApplicationInsights-node.js/pull/303), I forgot/did not realize I had to make explicit calls to enable the subscribers for these two new publishers.


OsvaldoRosado:
It would be good to cover this in a unit test so that mismatches in the two areas are caught quicker in the future

jackhorton:
Should those unit tests happen in this pull or a future one? I am not sure when I would be able to get to it, as it stands right now -- probably not until later this week or next week.

OsvaldoRosado:
A future one is fine. I'm in my own testing effort in preparation for the 1.0 release, so I may get to this test myself.

PR #324, opened 2017-09-14
Add automated functional tests

OsvaldoRosado:
This adds automated functional tests to our CI (and available for local use as well). The tests included are not complete, more will be added in future PRs.

The test framework spins up required dependent services (such as mongo or redis) in docker containers, and cleans them up after the test run is complete.

The framework architecture is split between a *TestApp* and a *Runner* with the following responsibilities:

**TestApp**
- HTTP server that receives requests from Runner
- Has AI SDK configured to send telemetry to Runner
- All major SDK settings configurable via environment variables
- Initially has no test endpoints configured, but has a list of tasks that it knows how to perform
- Can accept a JSON POST request that describes what endpoints to listen to, and when receiving requests on them, what tasks to perform

**Runner**
- Includes a built-in ingestion service to receive telemetry from TestApp
- Sends a list of task sequences to TestApp to validate scenarios
- Has a list of expectations for what telemetry should be received for each task possible in TestApp
  - Eg. Runner has a validation for TrackDependency, not a validation for a complete test that calls TrackDependency 3 times. Validations for entire task sequences are generated at run time using these base validations as building blocks
- Sends requests to TestApp after registering endpoints on it, then validates it received telemetry according to its expectation rules, and that all telemetry is correlated correctly
- Performs a final run with many parallel requests to ensure correlation works outside of the trivial sequential request scenario

New test scenarios can be configured by editing `Runner\TestSequence.json`. New tasks for use in these scenarios can be added by adding the new task to `TestApp\Tasks\index.js` and the expected telemetry output of that task to `Runner\TaskExpectations.js`


Tests/FunctionalTests/Runner/Config.js
```diff
@@ -0,0 +1,16 @@
+var Config = {
+    RunnerPort: "9091",
+    TestAppAddress: "http://localhost:9099",
+    InstrumentationKey: "743e9ee1-1c7b-42c1-b41c-9cac8ea659e8",
```
KamilSzostak: is it a real instrumentation key? if not maybe change it to `imafake-instr-umen-tati-onkey1111111` :)

Tests/FunctionalTests/TestApp/Main.js
```diff
@@ -0,0 +1,80 @@
+// process.env['APPLICATION_INSIGHTS_NO_DIAGNOSTIC_CHANNEL'] = true;
```
KamilSzostak: what is this env var doing?

OsvaldoRosado: This was a helper to locally disable diagnostic channel. Removed!

Tests/FunctionalTests/Runner/Config.js
OsvaldoRosado: Actually not used anymore! Removed

PR #332, opened 2017-09-22
Add "perfrun" mode to functional test runner

OsvaldoRosado:
This is used to aid manual performance testing

### DIFF
--- a/Tests/FunctionalTests/RunFunctionalTests.js
+++ b/Tests/FunctionalTests/RunFunctionalTests.js
@@ -2,7 +2,7 @@
 const fs = require('fs');
 const path = require('path');
 const childProcess = require('child_process');
-let noRunner = false;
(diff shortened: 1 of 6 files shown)


Tests/FunctionalTests/RunFunctionalTests.js
```diff
@@ -99,7 +100,11 @@ function main() {
         if (path === "-h" || path === "--help") {
             return help();
         }
-    } else {
+        if (process.argv.indexOf("-perfmode") !== -1) {
```
KamilSzostak: Why is `--help` in `path` and `-perfMode` in `process.argv`?

OsvaldoRosado: Path is always the first argument, and help can only be the first argument, just a reuse of the variable

PR #336, opened 2017-10-02
Prepare 1.0 release

OsvaldoRosado:


Library/Sender.ts
```diff
@@ -101,9 +101,14 @@ class Sender {
                         if (res.statusCode === 200) {
                             setTimeout(() => this._sendFirstFileOnDisk(), this._resendInterval);
                             // store to disk in case of burst throttling
-                        } else if (res.statusCode === 206 ||
```
KamilSzostak: what happened with `206`?

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "applicationinsights",
   "license": "MIT",
   "bugs": "https://github.com/Microsoft/ApplicationInsights-node.js/issues",
-  "version": "0.22.0",
+  "version": "1.0.0",
```
KamilSzostak: YEAH!1!!! 🎆🍾💥🥇👍 🍺

Library/Sender.ts
KamilSzostak: Nevermind, I've just noticed the comment.

package.json
OsvaldoRosado: 🎉 🎉

PR #341, opened 2017-10-14
Fix typo in comment

### LINKED ISSUE #340: Typo in Config.d.ts
In `Config.d.ts` there is a typo in the comment for [`maxBatchIntervalMs`](https://github.com/Microsoft/ApplicationInsights-node.js/blob/1263246014e68d600b1cb25572312bfde06d0bda/Library/Config.ts#L21). The default value is given as `1500` when it should be `15000`.

I tried to fix this myself, but this comment appears to be auto-generated, so I couldn't find the source of the difference in order to correct it, and there didn't seem to be any point in fixing it in `master` as I assume that the build process from `develop` would overwrite it.

martincostello:
Fix incorrect default value in comment for `maxBatchIntervalMs`.

Resolves #340.

### DIFF
--- a/Library/Config.ts
+++ b/Library/Config.ts
@@ -18,7 +18,7 @@ class Config {
     public endpointUrl: string;
     /** The maximum number of telemetry items to include in a payload to the ingestion endpoint (Default 250) */
     public maxBatchSize: number;
-    /** The maximum amount of time to wait for a payload to reach maxBatchSize (Default 1500) */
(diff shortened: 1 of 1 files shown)


OsvaldoRosado:
Great catch! Thanks for the PR!

OsvaldoRosado:
It looks like something has changed in the test infrastructure, since functional tests are now failing totally unrelated to these changes. Unfortunately, that's now blocking this PR. Will fix to unblock.

Issue #352, opened 2017-11-23
Dependency calls over HTTPS are not detected on Node 8.x


benc-uk:
Seems like any dependency call that goes over HTTPS is not tracked. 

For example I have a demo app which makes two backend API calls to different REST APIs, only the HTTP one is appearing in App Insights. Anything over HTTP is not getting picked up. See screenshot

In addition any Cosmos DB calls are not detected either, which naturally go over HTTPS

This is with Node 8.9.1 running in a Azure App Service

![image](https://user-images.githubusercontent.com/14982936/33184020-800bb47a-d072-11e7-8ec3-52553237b184.png)
 
## EDIT: This seems to be a problem with Node 8.x. I've switched my app to run on Node 6.9.1 and the HTTPS dependencies started coming through! :)

OsvaldoRosado:
This is interesting. It sounds like there's something special going on in the `request` library.

The following short snippet shows dependencies generated for native https calls)
```
var appInsights = require("applicationinsights");
appInsights.setup("IKEY").start();
appInsights.defaultClient.channel.send = console.log; // Send telemetry to console rather than AI

var https = require("https");
https.get("https://google.com");
```

benc-uk:
Interesting. However I'm not sure it is related to the `request` library, as none of my Cosmos DB calls were detected until I switched to Node 6.9.1. Looking in the Cosmos DB client library source code, they don't use `request`, they use the native `https` method

OsvaldoRosado:
This was due to internal changes in node on 8.9. If you update your SDK to 1.0.1 you should be able to use Node 8.9 and still retain autocollection.

Please feel free to reopen if you're still experiencing this!

Issue #351, opened 2017-11-17
SampleRate should not be applied for metrics telemetry


OsvaldoRosado:
In https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/EnvelopeFactory.ts#L85
we apply the configured sample rate to all telemetry items.

In https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/TelemetryProcessors/SamplingTelemetryProcessor.ts#L13
we refuse to sample out data for metrics, but we don't reset the sample rate to 1.

We should either never apply the configured sample rate to metrics items, or reset it in the `SamplingTelemetryProcessor`. Never setting the configured sample rate within `EnvelopeFactory` sounds cleanest, as it doesn't override actions users might take themselves in their own telemetry processors.

Issue #349, opened 2017-11-09
Double dependency reporting for HTTP Requests


annanhan:
It looks like the remote dependencies are getting double logged for HTTP requests. I have two apps that are communicating with each other, so I can see the request getting logged once on the receiving end.

![doubledependency](https://user-images.githubusercontent.com/10620173/32618430-b6b1305a-c545-11e7-92bb-6a33e1aaf5a1.png)
It also seems like the C# library could be double logging dependencies as well.

Could this be related to: https://github.com/Microsoft/ApplicationInsights-node.js/issues/323?

Here is how I have AppInsights configured:
```
var appInsights = require('applicationinsights');
var winston = require('winston');
var expressWinston = require('express-winston');
require('le_node');

appInsights.setup(config.appInsightsKey)
      .setAutoCollectConsole(false)
      .start();
  //register common properties to global client (used for auto-collection)
  appInsights.defaultClient.commonProperties = log.commonAppInsightsProperties;
  appInsights.defaultClient.addTelemetryProcessor(telemetryProcessor);

var log = new winston.Logger({
  transports:[
    new winston.transports.Console({
      handleExceptions: true,
      prettyPrint: true,
      level: config.logLevel.console
    }),
    new winston.transports.Logentries({
      name: 'logentriesLogger',
      token: config.logentriesKey,
      handleExceptions: true,
      minLevel: config.logLevel.logentries,
      withStack: true
    }),
    new winston.transports.CustomLogger({
      level: config.logLevel.appInsights,
      handleExceptions: true
    })
  ],
  exitOnError: false
});
```

OsvaldoRosado:
This could be related to #323. How are you making your http requests? With the built in `http` library or something else?

annanhan:
With [request](https://www.npmjs.com/package/request).

OsvaldoRosado:
In this case, this is definitely one of the problems described in #323, that also found the same to be happening with `request`. My current guess is that we're double collecting because of how request works internally.

Thanks for reporting this! Until this is fixed, you may be able to workaround with a telemetry processor that keeps track of what dependency calls it has seen in the current correlation context, or disable automatic dependency collection and do manual `trackDependency` calls

OsvaldoRosado:
The double reporting is due to 300-level redirects. The request module follows these, and subsequently makes a second request. In 1.0.0 status codes were not reported correctly to showcase this difference, and that was fixed in the newly-released SDK v1.0.1.

[...]

Issue #348, opened 2017-11-08
getCorrelationContext() doesn't return correlation context


OsvaldoRosado:
Copied from: https://github.com/Microsoft/ApplicationInsights-node.js/issues/323#issuecomment-342670907

> There is a bug in getCorrelationContext()
https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/applicationinsights.ts#L106
if (this._isCorrelating) { return CorrelationContextManager.CorrelationContextManager.getCurrentContext(); }
this._isCorrelating is always undefined since in 1.0.0 you've moved to private scope variables rather than properties on the instance (this). The code should be if(_isCorrelating) { ...

PR #353, opened 2017-11-28
Fix bugs for 1.0.1

OsvaldoRosado:
Address #352, #351, #349, #348 and prepare a 1.0.1 release.

### DIFF
--- a/AutoCollection/HttpDependencies.ts
+++ b/AutoCollection/HttpDependencies.ts
@@ -23,6 +23,11 @@ class AutoCollectHttpDependencies {
 
     private static requestNumber = 1;
 
+    // The difference between this and the disable flag above is that we delete this flag before returning to user code.
+    // The idea is we create this flag inside patched code, and delete it before returning from it, so the user has no impact from it
+    // We use this flag to ensure we don't double-collect telemetry
+    private static alreadyAutoCollectedFlag = '_appInsightsAutoCollected';
+
     private _client: TelemetryClient;
     private _isEnabled: boolean;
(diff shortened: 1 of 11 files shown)


AutoCollection/HttpDependencies.ts
```diff
@@ -55,29 +56,43 @@ class AutoCollectHttpDependencies {
...
-                const request: http.ClientRequest = originalHttpsRequest.call(
-                    https, options, ...requestArgs);
-                if (request && options && !(<any>options)[AutoCollectHttpDependencies.disableCollectionRequestOption]) {
-                    AutoCollectHttpDependencies.trackRequest(this._client, options, request);
-                }
-                return request;
-            };
-        }
+        https.request = (options, ...requestArgs: any[]) => {
+            const request: http.ClientRequest = originalHttpsRequest.call(
```
KamilSzostak: isn't it exactly the same as for http.request? Would it make sense to move it to a method and just pass originalHttpsRequest/originalRequest object?

OsvaldoRosado: Nope, the base functions take different numbers of arguments 😄

OsvaldoRosado: I do really dislike the duplication though..

KamilSzostak: Ok, let's leave it as it is.

OsvaldoRosado: Actually scratch that, I think this can be cleanly de-duplicated. Will change

Issue #368, opened 2018-02-02
Custom .trackRequest not being submitted


jem-Solytica:
When using Azure Functions with the applicationinsights package (1.0.1), the custom ```client.trackRequest()``` events seem to not be submitted to App Insights. After having issues with custom code, I am unable to get any results when querying the "request" in application insights analytics, but see the other telemetry data (metrics, events, exceptions, etc.). I have also tried the example code at: https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring. But even a new function app with just the example code does not cause the custom request items to show up in analytics.

No difference in outcome whether I use the base integration (adding the ```APPINSIGHTS_INSTRUMENTATIONKEY``` app setting), or a custom setup chain within the function. I have tried including/excluding the ```.start()``` from the setup chain, and various true/false settings on the setup chain.

jem-Solytica:
It appears that either numerous examples are wrong, or there's a missing type cast in both the .trackRequest() and .trackDependency() module.  

The issue is that the .responseCode must be of type string in both instances, but the examples indicate using numeric values. 

When the response codes are changed to strings (encapsulated in double or single quotes) in the .trackDependency() and .trackRequest() calls, the errors go away, and the telemetry can be seen in the Application Insights queries.


Here is the test code:
```
/*
 * File: index.js
 * Author: Jason E. McCollough 
 * Copyright: (c) 2018, Solytica Corp.
 * Description: Test file to debug application insights issues with custom requests and custom dependencies not being reported
 * Version: 1802.1
 * License: MIT
 */

function doAI() {
  console.log('Test function starting!');
  let appInsights = require('applicationinsights');

  appInsights.setup("<add_your_key>")
      .setAutoDependencyCorrelation(false)
      .setAutoCollectRequests(false)
      .setAutoCollectPerformance(false)
      .setAutoCollectExceptions(false)
      .setAutoCollectDependencies(false)
      .setAutoCollectConsole(false)
      .setUseDiskRetryCaching(false)
      .setInternalLogging(true, true); //we don't call start because we don't want any auto-collection
  let aiClient = appInsights.defaultClient;

  aiClient.trackTrace({message: "TEST: add .setInternalLogging() to setup chain"});  //send a specific trace, so it's easier to manually correlate the items below

  //  console.log(JSON.stringify(aiClient,null,2));  //review the full client settings, etc.

[...]

Issue #365, opened 2018-01-11
trackNodeHttpRequest does not follow type contract


OsvaldoRosado:
`trackNodeHttpRequest` says it accepts all parameters of type `Telemetry`, when it actually only uses `request`, `response`, and `properties`, and ignores all others.

Issue #362, opened 2017-12-14
TypeError: Cannot read property 'indexOf' of null


digeff:
We get this exception sometimes which might be related to this line https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/AutoCollection/CorrelationContextManager.ts#L206  and to having stack traces without filenames.

    TypeError: Cannot read property 'indexOf' of null
        at Function.AppInsightsAsyncCorrelatedErrorWrapper.orig.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/applicationinsights/out/AutoCollection/CorrelationContextManager.js:155:47)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
        at Function.NativeError.prepareStackTrace (/opt/logpoints-agent/node_modules/logpoints-agent/node_modules/zone.js/dist/zone-node.js:900:30)
[...]

Issue #358, opened 2017-12-12
MAC signature error when reading files from Azure File Storage


rudfoss:
I get this error when using ApplicationInsights correlation tracking in combination with Azure File Storage on nodejs:

`The` MAC signature found in the HTTP request [MAC] is not the same as any computed signature.`

I expect it's the same issue as the one documented here: https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/416 and here: https://github.com/Azure/azure-sdk-for-net/issues/3460

I checked the default configuration in `correlationHeaderExcludedDomains` and found that it includes four domains already. The problem is that they are all for blob storage and not file storage:

```
"*.blob.core.windows.net"
"*.blob.core.chinacloudapi.cn"
"*.blob.core.cloudapi.de"
"*.blob.core.usgovcloudapi.net"
```

The fix for me was to include `'*.file.core.windows.net'` in `correlationHeaderExcludedDomains` as well, but shouldn't that also be part of the pre-configured domains?

OsvaldoRosado:
Great catch! Thanks for reporting this! This scenario should indeed work out-of-the-box.

It looks like the .NET SDK is less specific here, leaving out the `blob` part of the default excluded domain names. Doing the same here would avoid this problem

rudfoss:
No problem, happy to help. Thanks for the quick reply!

mwmw7:
error message :
express:azure   name: 'StorageError',
  express:azure   message: 'Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.\nRequestId:cc46450d-001a-0004-6549-9fd1ef000000\nTime:2018-02-06T12:52:42.2939570Z',
  express:azure   code: 'AuthenticationFailed',
[...]

PR #369, opened 2018-02-08
Fix bugs for 1.0.2

OsvaldoRosado:
These changes address the following issues: #362, #358, #346, #344, #368, #365 

These changes also improve internal logging coverage in a few areas, and fix a few stability issues with retry caching from disk.

### DIFF
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -193,7 +193,7 @@ export class CorrelationContextManager {
             if ((<any>orig).prepareStackTrace) {
                 (<any>orig).stackRewrite= false;
                 var stackTrace = (<any>orig).prepareStackTrace;
-                (<any>orig).prepareStackTrace = (e: any, s: any) => {
+                (<any>orig).prepareStackTrace = (e: any, s: any[]) => {
                     // Remove some AI and Zone methods from the stack trace
                     // Otherwise we leave side-effects
 
@@ -203,20 +203,25 @@ export class CorrelationContextManager {
                     //  Zone | Zone | CorrelationContextManager | CorrelationContextManager | User
                     var foundOne = false;
                     for (var i=0; i<s.length; i++) {
-                        if (s[i].getFileName().indexOf("AutoCollection/CorrelationContextManager") === -1 &&
-                            s[i].getFileName().indexOf("AutoCollection\\CorrelationContextManager") === -1) {
+                        let fileName = s[i] && typeof s[i].getFileName === 'function' && s[i].getFileName();
+                        if (fileName) {
+                            if (fileName.indexOf("AutoCollection/CorrelationContextManager") === -1 &&
+                                fileName.indexOf("AutoCollection\\CorrelationContextManager") === -1) {
 
-                            if (foundOne) {
-                                break;
+                                if (foundOne) {
+                                    break;
+                                }
+                            } else {
+                                foundOne = true;
                             }
-                        } else {
-                            foundOne = true;
                         }
                     }
                     // Loop above goes one extra step
                     i = Math.max(0, i - 1);
                     
-                    s.splice(0, i);
+                    if (foundOne) {
+                        s.splice(0, i);
+                    }
                     return stackTrace(e, s);
                 }
             }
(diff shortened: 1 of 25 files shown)


Declarations/Contracts/TelemetryTypes/DependencyTelemetry.ts
```diff
@@ -34,7 +34,7 @@ export interface DependencyTelemetry extends Telemetry {
     /**
      * Result code returned form the remote component. This is domain specific and can be HTTP status code or SQL result code
      */
-    resultCode: string;
+    resultCode: string | number;
```
Dmitry-Matveev: Will we serialize it to string for JSON purposes, so that endpoint accepts the item?

Library/Sender.ts
```diff
@@ -150,38 +156,112 @@ class Sender {
...
                 this._onErrorHelper(error);
                 return;
             }
 
-            //create file - file name for now is the timestamp, a better approach would be a UUID but that
-            //would require an external dependency
-            var fileName = new Date().getTime() + ".ai.json";
-            var fileFullPath = path.join(directory, fileName);
+            this._getShallowDirectorySize(directory, (err, size) => {
+                if (err || size < 0) {
```
Dmitry-Matveev: If the AI endpoint is inaccessible - will there be any perf issues in case we concurrently queue up a lot of storage operations such as calculate size of all files and then save a file? Or due to batching in SDK such calls will be rare and only happen couple times a minute, so no hit on storage throughput is expected?

Library/Sender.ts
```diff
@@ -150,38 +156,112 @@ class Sender {
...
+        return totalSize;
+    }
+
     /**
      * Stores the payload as a json file on disk in the temp directory
      */
     private _storeToDisk(payload: any) {
-
-        //ensure directory is created
+        // tmpdir is /tmp for *nix and USERDIR/AppData/Local/Temp for Windows
```
Dmitry-Matveev: Are we sure it's always USERDIR on Windows? I remember there were cases when %temp% defaults to "\Windows\Temp" where many users may have access. If it can end up in Windows\Temp - is there a way in node to give Windows permissions on directory such as ACL to protect it? If it can't - please disregard the comment.

Declarations/Contracts/TelemetryTypes/DependencyTelemetry.ts
OsvaldoRosado: Yep! While converting to an Envelope it's converted to string.

I changed the type here because many pre-existing docs incorrectly assumed this to take a number. I figured it would be most beneficial to just actually fully support it as a number (before this it would silently fail as we don't expect number and would not convert to string).

Library/Sender.ts
OsvaldoRosado: Batching is indeed the rate limit on this. On top of that, because this call is fully async (not blocking the event loop) I don't expect perf impact even if the batch size is configured to be very low.

Library/Sender.ts
OsvaldoRosado: It's not guaranteed to always be USERDIR, the system configuration is used to get the temp directory. Here's the logic: https://github.com/nodejs/node/blob/master/lib/os.js#L121
```
path = process.env.TEMP ||
       process.env.TMP ||
       (process.env.SystemRoot || process.env.windir) + '\\temp';
if (path.length > 1 && path.endsWith('\\') && !path.endsWith(':\\'))
  path = path.slice(0, -1);
```

So if the `TEMP` or `TMP` environment variable is changed, that new temp will be respected instead.

As for ACL: In node itself the chmod arguments we're already using is the API presented for changing permissions. Unfortunately the underlying libuv library that does this on node's behalf doesn't map chmod to ACL in a way that will set anything other than the readonly flag.

If we're interested in doing this beyond what node itself supports, we'll need native code(or a library that itself has native code). I found one library that does this, but it doesn't have the same amount of node version support that we have (and I'm not yet sure of the implications of taking a dependency on windows-only native code)

AutoCollection/CorrelationContextManager.ts
```diff
@@ -203,20 +203,25 @@ export class CorrelationContextManager {
                     //  Zone | Zone | CorrelationContextManager | CorrelationContextManager | User
                     var foundOne = false;
                     for (var i=0; i<s.length; i++) {
-                        if (s[i].getFileName().indexOf("AutoCollection/CorrelationContextManager") === -1 &&
-                            s[i].getFileName().indexOf("AutoCollection\\CorrelationContextManager") === -1) {
-
-                            if (foundOne) {
-                                break;
+                        let fileName = s[i] && typeof s[i].getFileName === 'function' && s[i].getFileName();
```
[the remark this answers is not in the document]
OsvaldoRosado: Good catch! Thanks!

Issue #246, opened 2017-05-12
Add support for tunneling data through a proxy


OsvaldoRosado:
We can do this by adding support for a configurable Agent (https://nodejs.org/api/http.html#http_class_http_agent) that the Sender uses

JasonJJFS:
Hi all any ETA on this issue? Or is there a mitigation for running behind a corporate proxy.

OsvaldoRosado:
No ETA for this yet. The workaround right now is to add exceptions to your firewall for the IPs listed at https://docs.microsoft.com/azure/application-insights/app-insights-ip-addresses

Depending on how your proxy works, you may also be able to manually forward to it by changing the `endpointUrl` address in the `Config` object.

JasonJJFS:
Yes we've done that and now works as expected. Many thanks.

PR #382, opened 2018-04-03
Feature/proxy support

spike008t:
Hello!

So here a contribution to support `http_proxy` and `https_proxy` env variable.
I've done that because I do really need it at the moment.

If you have any suggestion to improve it, I'm agree to update the code.

Maybe if I do have the time, I can investigate with Http.Agent class to handle it. 
But at the moment the following update works for me.

Can be an answer for #246

### DIFF
--- a/Library/Config.ts
+++ b/Library/Config.ts
@@ -31,8 +31,10 @@ class Config {
     public correlationIdRetryIntervalMs: number;
     /** A list of domains to exclude from cross-component header injection */
     public correlationHeaderExcludedDomains: string[];
-    /** The proxy URL to use if defined */
-    public proxyUrl: string;
+    /** The proxy for http URL to use if defined */
+    public proxyHttpUrl: string;
+    /** The proxy for https URL to use if defined */
+    public proxyHttpsUrl: string;
 
     private endpointBase: string = "https://dc.services.visualstudio.com";
     private setCorrelationId: (v: string) => void;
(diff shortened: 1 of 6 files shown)


OsvaldoRosado:
Thanks for the PR! These changes look great!

I did notice that the proxy isn't being respected in the web request done in https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/CorrelationIdManager.ts#L37

To avoid code duplication of needing to support proxy in both places, I think we should move these changes into a helper in Library/Util.ts that takes unproxied http/https requestOptions, does the necessary modifications to requestOptions to apply the proxy when necessary, and calls http/https.request itself (or maybe just returns modified requestOptions and consumers still do their own call to http/https.request?)

There should probably also be added unit tests for this. If we follow the above suggestion, it could be two short tests added in https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Tests/Library/Util.tests.ts to check that a proxy placed in config ends up changing the requestOptions and when no proxy is configured no modification happens.

spike008t:
Thank you for your review and feedback!

I didn't notice that request is also called in CorrelationIdManager.ts. 
I'm with you to move it as a helper. And I think that wrapping the http/https.request itself is a good option for my point of view.

I'll update my branch and let you know when that is ready.

spike008t:
@OsvaldoRosado So I've added the PR. But I don't see why the CI failed...

Issue #383, opened 2018-04-13
The envelope properties init order is not compatible with Xpert (Microsoft Inside)


XiaoliangHuang:
# Issue
we use this sdk to collect node.js logs, and want to sent this data to [xpert](https://xpert/), and then encountered some problems that the order or envelope properties are not supported.

# Root cause
The requirement is that 'data', 'tags' or 'ext' (part B/C) should appear before name, time, ikey... (part A).
the detailed rule can be find in the following section:

[Http Request Format using JSON](https://osgwiki.com/wiki/Vortex/ClientProtocol):
Although we accept JSON, Vortex requires an order to the fields in the JSON payload. See the [Vortex/EventQuality](https://osgwiki.com/wiki/Vortex/EventQuality) page for details on exactly what you have to do. the rule "FieldsOutOfOrder"

In other strongly typed language like  C#,  new a instance will allocate memories for all the properties as it defined, but in JS,  the memory allocation is based on the order fields added, then when it converted to string (Json.Stringify()), the order will different from its declaration.

# Fix
see the pull request https://github.com/Microsoft/ApplicationInsights-node.js/pull/384

hectorhdzg:
@XiaoliangHuang let us know if this is still relevant, xpert documentation doesn't include any details about JSON format anymore.

PR #384, opened 2018-04-13
Adjust the envelope properties initiate order to meet Xpert requirement

XiaoliangHuang:
fix bug #383 The envelope properties init order is not compatible with Xpert (Microsoft Internal).

there fix contains the following parts:

1. adjust the properties init step of creating envelope 
2. set the ikey part as lowercase to consistent with other application insights sdks

### DIFF
--- a/Declarations/Contracts/Generated/Envelope.ts
+++ b/Declarations/Contracts/Generated/Envelope.ts
@@ -48,11 +48,5 @@ import Base = require('./Base');
          */
         public data: Base;
         
-        constructor()
-        {
-            this.ver = 1;
-            this.sampleRate = 100.0;
-            this.tags = {};
-        }
     }
 export = Envelope;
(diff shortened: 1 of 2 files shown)


OsvaldoRosado:
Thanks for the PR!

Could you add a unit test to check that the properties you care about are still first after serialization by `JSON.stringify` in `Channel`? 

Ordering of properties in objects were not guaranteed pre-ES2015 (though many engines happened to implement objects this way), and even post-ES2015 order is unclear in V8 vs other engines for certain property names (namely property names that are or cast to numbers get placed in sorted order in V8). It would be great to verify the ordering expectation works on our current test suite and avoid any regressions in the future.

Declarations/Contracts/Generated/Envelope.ts
```diff
@@ -48,11 +48,5 @@ import Base = require('./Base');
          */
         public data: Base;
         
-        constructor()
```
OsvaldoRosado: Requiring this change is concerning, as it's a file generated from our BOND schema and will eventually get overwritten. Is there a better way to accomplish the reordering?

OsvaldoRosado:
@XiaoliangHuang are you still looking to iterate on this PR or should we close this for now?

OsvaldoRosado:
Closing due to lack of new PR iterations. Feel free to re-open if you're looking to work on this again!

Issue #180, opened 2017-01-27
Intermittent ApplicationInsights:Sender [ [Error: UNABLE_TO_VERIFY_LEAF_SIGNATURE] ]


annanhan:
While using applicationinsights@0.17.2, I noticed these errors in our logs. The node app is hosted on Heroku using an SSL endpoint addon. After doing some research, it looks like node isn't accepting the server's cert. The messages were getting to Application Insights as far as I could tell.

Additionally, it caused a huge memory leak. After I disabled the package, the memory profile was back to normal.

Here are the other packages we're using:
"engines": {
    "node": "^0.10.28",
    "npm": "^1.4.9"
  },
  "dependencies": {
    "amqp.channel": "^0.0.8",
    "applicationinsights": "^0.17.2",
    "args-list": "^0.3.3",
    "bluebird": "^2.1.2",
    "cookie-parser": "^1.3.4",
    "cool-ascii-faces": "^1.3.3",
    "cors": "^2.3.1",
    "ddi": "^1.1.0",
    "express": "^4.8.0",
    "express-winston": "^1.0.0",
    "formdefutils": "^0.0.2",
    "gift": "^0.6.0",
    "github-webhook-handler": "^0.5.0",
    "glob": "^4.3.2",
    "le_node": "^1.0.12",
    "lodash": "^3.10.1",
    "mailgun-js": "^0.7.7",
    "mapquest": "^0.2.0",
    "moment": "^2.10.3",
    "multiparty": "^4.1.2",
    "node-xlsx": "^0.5.1",
    "phone-formatter": "0.0.2",
    "request": "^2.34.0",
    "rimraf": "^2.2.8",
    "romis": "^2.0.0",
    "sprintf": "^0.1.4",
    "urlencoded-request-parser": "^1.0.1",
    "us": "^1.0.3",
    "uuid": "^2.0.1",
    "winston": "^1.0.0",
    "xml2js": "~0.4.4",
    "yamljs": "^0.2.1"
  }

SergeyKanzhelev:
@annanhan thanks for reporting. So Application Insights fail to communicate to the server. Does it repro on all computers or on your dev box? Do you think this is related to this issue: https://github.com/Microsoft/ApplicationInsights-node.js/issues/177?

Can you please try to open https://dc.services.visualstudio.com/api/ping from the same computer and see whether cert is trusted. 

[...]

PR #395, opened 2018-05-24
Fix incorrect identity being used for ACL assignment in App Services / Azure Functions and adjust connection error logging.

OsvaldoRosado:
This uses WindowsIdentity (through PowerShell) to get the current user's identity for use in ICACLS rather than relying on the username environment variable being accurate. (Discovered in #180 that the environment variable is incorrect for App Services / Functions)

This also tweaks the logic for logging connection errors to only log in the default (non-verbose) configuration if 5 consecutive connection failures occurred (or every time if retries are disabled). The connection error logs are updated to have helpful descriptions of what's going on before the technical network error from Node. (Will help clarify other issues in #180 by ensuring logs only appear when there's a likelihood of telemetry loss)

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -15,7 +15,9 @@ import Util = require("./Util");
 class Sender {
     private static TAG = "Sender";
     private static ICACLS_PATH = `${process.env.systemdrive}/windows/system32/icacls.exe`;
+    private static POWERSHELL_PATH = `${process.env.systemdrive}/windows/system32/windowspowershell/v1.0/powershell.exe`;
     private static ACLED_DIRECTORIES: {[id: string]: boolean} = {};
+    private static ACL_IDENTITY: string = null;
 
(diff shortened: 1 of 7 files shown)


Library/Sender.ts
```diff
@@ -213,9 +271,15 @@ class Sender {
         // For performance, only run ACL rules if we haven't already during this session
         if (Sender.ACLED_DIRECTORIES[directory] === undefined) {
             // Restrict this directory to only current user and administrator access
-            this._runICACLS(this._getACLArguments(directory), (err) => {
-                Sender.ACLED_DIRECTORIES[directory] = !err;
-                return callback(err);
+            this._getACLIdentity((err, identity) => {
+                if (err) {
+                    return callback(err);
```
SergeyKanzhelev: do we need to set a flag for directory here that will tell that operation didn't succeed and it should not try again?

OsvaldoRosado: Good catch!

OsvaldoRosado:
@SergeyKanzhelev added the test

Issue #406, opened 2018-06-26
Error with puppeteer: Cannot read property 'protocol' of undefined


spike008t:
I'm facing this issue when using puppeteer (v1.5.0 at the moment) with the latest version of ApplicationInsights-node.js

When calling:

```js
const browser = await puppeteer.launch();
```

I do have the following error:

```
Unhandled Promise rejection: Cannot read property 'protocol' of undefined ; Zone: <root> ; Task: null ; Value: TypeError: Cannot read property 'protocol' of undefined
    at Function.HttpDependencyParser._getUrlFromRequestOptions (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencyParser.js:135:62)
    at new HttpDependencyParser (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencyParser.js:28:46)
    at Function.AutoCollectHttpDependencies.trackRequest (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencies.js:105:29)
    at clientRequestPatch (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencies.js:46:45)
    at Object.AutoCollectHttpDependencies._initialize.http.request (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencies.js:59:13)
    at Object.AutoCollectHttpDependencies._initialize.http.get (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencies.js:80:52)
    at WebSocket.initAsClient (/src/node_modules/puppeteer/node_modules/ws/lib/websocket.js:522:33)
    at new WebSocket (/src/node_modules/puppeteer/node_modules/ws/lib/websocket.js:62:20)
    at EventEmitter.createForWebSocket.Promise (/src/node_modules/puppeteer/lib/Connection.js:32:18)
    at new ZoneAwarePromise (/src/node_modules/zone.js/dist/zone-node.js:551:29)
    at Function.createForWebSocket (/src/node_modules/puppeteer/lib/Connection.js:31:12)
    at Function.launch (/src/node_modules/puppeteer/lib/Launcher.js:163:39)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7) TypeError: Cannot read property 'protocol' of undefined
    at Function.HttpDependencyParser._getUrlFromRequestOptions (/src/node_modules/applicationinsights/out/AutoCollection/HttpDependencyParser.js:135:62)
[...]

PR #407, opened 2018-06-26
fix: Fix error with puppeteer, agent can be undefined

spike008t:
Fix the issue #406 

When puppeteer start, the protocol and agent can be undefined.

### DIFF
--- a/AutoCollection/HttpDependencyParser.ts
+++ b/AutoCollection/HttpDependencyParser.ts
@@ -139,7 +139,7 @@ class HttpDependencyParser extends RequestParser {
         }
 
         // Mix in default values used by http.request and others
-        options.protocol = options.protocol || (<any>request).agent.protocol;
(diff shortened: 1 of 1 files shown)


AutoCollection/HttpDependencyParser.ts
```diff
@@ -139,7 +139,7 @@ class HttpDependencyParser extends RequestParser {
         }
 
         // Mix in default values used by http.request and others
-        options.protocol = options.protocol || (<any>request).agent.protocol;
+        options.protocol = options.protocol || ((<any>request).agent && (<any>request).agent.protocol) || undefined;
```
OsvaldoRosado: Just to be clear, this doesn't do anything about `protocol` being undefined right? But rather if `agent` doesn't exist?

OsvaldoRosado:
Thanks for this PR!

AutoCollection/HttpDependencyParser.ts
spike008t: That’s right

PR #426, opened 2018-09-19
Remove dead code

jrieken:
The PR makes the app insights modules webpack'able by removing an unused dynamic require-call.


OsvaldoRosado:
Thanks!

Issue #371, opened 2018-02-28
Nested metadata from winston logger shows as [object Object]


ThomasKruegl:
We are using the winston logger in a node.js application. We are also using the library express-winston to log our requests. This library utilizes the metadata of winston for request and response details, which are complex objects. From my understanding in this case the following code of winston.sub.ts converts the metadata to properties for AI:
```
client.trackTrace(
            {
                message: event.data.message,
                severity: AIlevel,
                properties: event.data.meta
            });
```

edit: To be precise, it is not a problem of the winston subscription, it is also not possible to log nested metadata directly, using trackTrace(). The conversion actually happens in Util.validateStringMap.

But the type for properties has an index signature, values get typed to string. This causes nested objects to be converted to strings, resulting in [object Object]. Is it possible to change this? I can imagine several possible solutions, but i do not know if there are some limitations in the insights API, or the library. I can imagine:
- Change the type to 'any' (would have to check how this gets serialized and shows up in insights then)
- use util.inspect or JSON.stringify, may need some checks when to apply that.

We are also investigating how we can solve this on our side, finding a way around this in express-winston, or not using express-winston, but i wanted to open the issue, since it is a general problem with nested metadata objects.

regards,
Thomas

willmorgan:
Just a thought, if you're using Winston, have you considered using https://github.com/bragma/winston-azure-application-insights?

ThomasKruegl:
Thanks for the hint :) I think i had seen that library, but had forgotten about it.
If i understand it correctly, we could disable the AutoCollection for the winston console transport (or do not use a console transport, but we want that additionally), and instead use the AI transport provided there.
From a first glance, i guess we would have the same problem though, because that library is at some point just calling
```
this.client.trackTrace(msg, aiLevel, meta);
```
And this will run into exactly the same problem as described above, since it also does not check or convert the metadata before (only some changes for errors).

willmorgan:
I now maintain the library and I'm actually working on logging for my own app this afternoon.

Out of interest, can you paste a few traces from AI here so I can get a bit more of an idea on what you're seeing?

How do you have console logging switched on? Is it `true` or `true, true`?

ThomasKruegl:
We are using the defaults, which is only true, ignoring "console.log".
Here is the relevant code:
https://gist.github.com/ThomasKruegl/7d69cc57157d5e9a5023fd13a36c7524

[...]

PR #428, opened 2018-09-25
Convert null, undefined props to empty strings

willmorgan:
When I switched on debug logging for this library, I saw the warnings (logged at info) that the property types were invalid.

This change aims to normalise the shape of incoming `customDimensions` fields so that they aren't dropped. Previously, within normal operation of an application under instrumentation, if any properties were uninitialised then they would be silently dropped.


willmorgan:
@OsvaldoRosado Keen to get your thoughts on this.

Also interested in your thoughts regarding how to handle nested objects, as casting to `[object Object]` could get solved better too...

https://github.com/Microsoft/ApplicationInsights-node.js/issues/371#issuecomment-369684318

Library/Util.ts
```diff
@@ -158,6 +158,8 @@ class Util {
                 if (propertyType !== "string") {
                     if (property != null && typeof property.toString === "function") {
                         property = property.toString();
+                    } else if (property === null || propertyType === "undefined") {
+                        property = "";
```
OsvaldoRosado: Could you comment on the scenario where this is useful? If the value for the custom dimension is missing, why incur extra billing cost for bytes ingested with the dimension name?

OsvaldoRosado:
@willmorgan 
> Also interested in your thoughts regarding how to handle nested objects, as casting to [object Object] could get solved better too...

It would be great to handle nested objects properly! Especially because Analytics can handle complex JSON objects just fine when building your own queries. The only catch here is in how to implement. `JSON.stringify` is most straightforward but unreliable because it can throw an exception if it finds any field it can't serialize (rather than ignoring that field and moving on).

Perhaps looping the existing method into nested objects is the way to go? We'd have to be careful to properly handle objects that contain reference loops though..

Library/Util.ts
willmorgan: This is useful when you have a manager or operations person looking through logs, who is not necessarily able/willing to access the codebase, who's going to ask the question "why isn't this being logged?" when it actually is.

You have a good point regarding billing cost - I'm not the one footing the bill, here 😉

willmorgan:
@OsvaldoRosado It looks like we have to change the TS contract slightly to allow this, then?

[Within the Winston transport we're doing a util.inspect](https://github.com/bragma/winston-azure-application-insights/commit/7f5270a444ea238dc276100ccc09997bd0ce0d79#diff-3ec32060bf3df2946b53111a1b8f903fR54) to get _something_.

Regarding circular references, the only way I can think of detecting those would be to pass complex objects into a `Map` and test for pre-existence, and then use a [JSON path](https://www.npmjs.com/package/jsonpath) expression to refer to it. Does that sound good? I have some code leftover from another project...

willmorgan:
[Functionality above](https://github.com/Microsoft/ApplicationInsights-node.js/pull/428#issuecomment-424644816) was implemented in 6b80cee

dhensby:
Is that test failure just an anomaly? Can someone restart the build to double check?

OsvaldoRosado:
@dhensby Restarted!

OsvaldoRosado:
@willmorgan Which contract has to change? I might be mistaken but I believe customDimensions from an ingestion perspective must still be a pure dictionary of string to string - making that particular contract immutable from our perspective. So whatever we do for nested objects should eventually serialize into a JSON string.

willmorgan:
@OsvaldoRosado Thanks for checking over - I've taken your comments to mean I should:

1. Clarify the `properties` / `customDimensions` field's structure. I _think_ it can take:

```json
{
    "top_level_key": {
        "nested_key": "foo",
        "nested_key_2": "bar"
    },
    "is_boolean": true
}
```

But if you're saying that it actually needs to look more like the below, where anything nested gets serialized...

```json
{
    "top_level_key": '{ "nested_key": "foo", "nested_key_2": "bar" }',
    "is_boolean": true
}
```

...then we can look at doing that, but then I think we should replace JSON paths with just `[Circular]`, as traversing the structure will be more difficult due to needing to JSON.parse more than once for anything nested)

2. Update the type signature accordingly.
3. Looking at moving value trimming to somewhere better, if possible.

OsvaldoRosado:
@willmorgan Your understanding of how it should look (second example) is correct. Our contract with the backend is that custom dimensions is a `string, string` dictionary, but Analytics will be able to understand the JSON object we've placed into that string. Conceptually, `JSON.stringify` applied to any top level value that's not a primitive is the valid behavior.

As for how circular objects get handled. I'm totally fine with any approach, just as long as there's anything in place at all that prevents the SDK from infinite looping (or throwing) upon getting such an object - with preference being to solutions that retain at least some of the object (which is why `JSON.stringify` isn't the perfect option here - but could potentially still be acceptable if the alternatives are complex or meaningfully slower)

An extra note about moving the trimming too. I brought it up since I noticed the code where it's in has many different paths to `return`. My observation was that it might be more future proof to trim in one place rather than before each `return` so future editors of the code have less of a chance to get it wrong :smile:

Thanks for working on this!

willmorgan:
@OsvaldoRosado Any chance you can take a look?

Library/Util.ts
```diff
@@ -146,25 +146,62 @@ class Util {
...
+        const valType = typeof value;
+        if (valType === "string" || valType === "number" || valType === "boolean") {
+            return value.toString().trim(0, Util.MAX_PROPERTY_LENGTH);
+        } else if (value === null || value === undefined) {
+            return "";
+        } else if (Util.isArray(value)) {
+            return JSON.stringify(value.map((arrVal: any, arrKey: any) => {
+                return Util.toSimpleValue(arrVal, seenObjects, jpath.concat(arrKey));
+            }));
+        } else if (valType === "object") {
```
OsvaldoRosado: I think this will pick up Date objects as well. Is that okay?

Library/Util.ts
```diff
@@ -146,25 +146,62 @@ class Util {
...
+            let map = <{ [key: string]: string }>{};
+            for(let field in value) {
+                let fieldVal = value[field];
+                let simpleValue;
+                simpleValue = Util.toSimpleValue(fieldVal, seenObjects, jpath.concat(field));
+                if (fieldVal in seenObjects) {
+                    return JSON.stringify({ $ref: seenObjects[fieldVal] });
+                } else {
+                    seenObjects[fieldVal] = jpath.join(".");
```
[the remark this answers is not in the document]
willmorgan: This _should_ be storing things as `[any]: 'string'` as you can key objects by other objects in Node.

Library/Util.ts
willmorgan: I'll add a test case for this

Library/Util.ts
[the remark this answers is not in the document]
OsvaldoRosado: Are you sure? That's a node console I screenshotted. `Object.keys()` type for example always returns an array of strings. A quick way to test is to add a unit test for an object with two nested objects (eg. `{a: {b: 'c'}, d:{b: 'c'}}`) and see if it succeeds (treats them individual) or fails (thinks its circular).

willmorgan:
Hi @OsvaldoRosado - I've fixed the date stuff, but it seems that I was pretty wrong about keying objects by other objects. The "right" way of doing this, I think, is to use a `WeakMap`. I had initially taken the object-keying approach instead of using Map as this project's TS compilation target is `ES5`.

Given that [the majority of ES2015 support goes back to Node 4.x](https://node.green/#ES2015-built-ins-WeakMap), does it make sense therefore to bump the compatibility up and get this into a point release? That way we can use `WeakMap` and some other goodies.

If we can make a decision on this then I can go through and address some of the other comments (around `any` etc...)

Cheers

OsvaldoRosado:
@willmorgan I'm not sure nested customDependencies alone would be a good reason to bump major releases (I think this would be a major release due to the compatibility change - we currently work back to Node 0.10)

If we can't use weakmap, do you think it would be better to simply do JSON.stringify on each top level property with a try catch around it - thus dropping any objects that have nesting entirely? Or is there another good data structure to use here?

willmorgan:
@OsvaldoRosado What about using https://github.com/medikoo/es6-weak-map?

OsvaldoRosado:
@willmorgan I wouldn't like to add an extra dependency to this project without gaining a ton of functionality from it. Right now we have almost no dependencies, which helps out when it comes to compliance tasks like scans for vulnerabilities and threat models.

The lib block sounds great if es2015 is supported in the versions of Node we test against (0.10 +)

willmorgan:
After further reading on TS, it looks like declaring `lib` options just imports TS interfaces for Node core from a given feature set, so it doesn't actually bring in any supporting code. Therefore 0.10 isn't going to work there - which is a shame because I got that working locally.

[Node 0.10 is extremely old, has reached EOL, and is no longer receiving security updates](https://github.com/nodejs/Release#end-of-life-releases), so I think that bumping Node support to something a little more modern (6.x would be good, but 4.x would still be easy) is not going to be painful for the userbase.

The alternatives we have:

- performing a `util.inspect` on the object and letting it handle Circular references, but this won't return JSON
- replace `seenObjects` with an array which is searched every time `toSimpleValue` encounters an object, but this will be slow and... kinda nasty

The original aim of this PR was to improve the logging of nested objects so that as a user of AI, I'm not greeted with `[object Object]` or `[object Error]` in my customDimensions fields when logging things. I think it's reasonable to expect those data structures (maybe with the exception of circular refs) to serialize down into something useful with minimal input from userland code.

I've spent quite a lot of time on this trivial, but important feature. If there's no way forward then I can always port all of this to `winston-azure-application-insights` and leave it to MSFT/another OSS'er to patch the upstream dependency, but that would be a shame because I think it's quite an important feature to have.

willmorgan:
The problem with that example is that native Node errors can't be serialized, so something like this wouldn't pass (you actually get `{}`):

```js
assert.deepEqual(Util.validateStringMap({ error: new Error("Help") }), { error: { message: "Help" } });
```

I'd be interested to know how you handle logging errors in any projects you work on that use this library.

I'll try and adapt your simple example to cover the edge cases.

OsvaldoRosado:
@willmorgan Interesting edge case about error objects. I wouldn't have thought of that!

In general, the intended way to track error objects in this SDK is to use the `trackException` method. This allows Application Insights to properly understand the information contained in the exception, and present it in the UI appropriately - as well as relate it for things like RCA on failed requests. Even beyond the UI, we have a special table for exceptions with more error-related schema fields. Automatic dependency correlation should place the exception in the same transaction as other telemetry that it's related to.

image

That being said, I can see value in allowing error objects to be placed in custom dimensions if this information is passed in from logging frameworks that don't distinguish exceptions from other kinds of traces.

Library/Util.ts
```diff
@@ -146,33 +146,77 @@ class Util {
...
-                var propertyType = typeof property;
-                if (propertyType !== "string") {
-                    if (property != null && typeof property.toString === "function") {
-                        property = property.toString();
-                    } else {
-                        Logging.info("key: " + field + ", invalid property type: " + propertyType);
-                        continue;
-                    }
-                }
+    protected static extractError(err: Error): { message: string, stack_trimmed: string } {
```
OsvaldoRosado: It might be better to reuse the stack parsing that already exists in this SDK (see https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/EnvelopeFactory.ts#L242 )

If we don't reuse, we should ensure the `stack` object exists and is a string. Node code can use the `prepareStackTrace` method to change the construction of the `stack` object (see https://github.com/v8/v8/wiki/Stack-Trace-API#customizing-stack-traces).

That being said, I'm not sure more than message should be recorded to avoid reaching the quite short length limits on custom dimensions.

willmorgan: OK - I'll drop the stack.

willmorgan:
PR updated; thanks for explanation @OsvaldoRosado, that does seem the more correct way of doing it. Right now we're using traces and setting the severity level depending on the log level - much of what we're doing isn't absolute best practice which we're looking to change, but I agree that logging the whole stack isn't good for security/bloat reasons.

Are there any more items that need addressing, or do you think we're ready now?

Btw - CI seems to have failed on 0.12 for unrelated reasons, any chance you can restart?

Library/Util.ts
```diff
@@ -146,33 +146,67 @@ class Util {
...
+            let property: string = '';
+            const origProperty: any = obj[field];
+            const propType = typeof origProperty;
+
+            if (propType === "string" || propType === "number" || propType === "boolean") {
+                property = origProperty.toString();
+            } else if (origProperty === null || propType === "undefined") {
+                property = "";
+            } else if (propType === "object" && typeof origProperty.toJSON === "function") {
+                property = origProperty.toJSON();
```
OsvaldoRosado: Stringify should still be called on this, as `toJSON` should return an object for stringifying, not an already serialized representation. As such, we can't guarantee its return value is a string until it's stringified. (For reference, see this comment in Crockford's implementation of JSON.stringify https://github.com/douglascrockford/JSON-js/blob/master/json2.js#L25-L32)

Library/Util.ts
```diff
@@ -146,33 +146,67 @@ class Util {
...
+    }
 
-                map[field] = property.trim(0, Util.MAX_PROPERTY_LENGTH);
-            }
-        } else {
+    /**
+     * Capture all properties on an object rather than relying on JSON.stringify to do this,
+     * which may yield an unwanted result, depending on property visibility
+     */
+    protected static extractObject(origProperty: any): { [key: string]: any } {
```
OsvaldoRosado: I think the return type of this is still `any`, not `{ [key: string]: any }`

OsvaldoRosado:
@willmorgan Kicked off a new build for 0.12. Besides the missing forcing of toJSON to a string representation, I think these changes look good! Once it's fixed I think we can merge this in.

Thanks for this PR!

Library/Util.ts
willmorgan: Yes, it's impossible to guess what you get here, but we can rely on `JSON.stringify` and a primitive type check later.

willmorgan:
@OsvaldoRosado Feedback implemented. I've added some missing tests to actually cover the cases we discussed yesterday.

I've also removed function types from the returned property map and added a debugger message when we do this.

willmorgan:
@OsvaldoRosado thanks - my pleasure - it was educational!

PR #447, opened 2018-11-06
align custom property max length with backend one

### LINKED ISSUE #444: Custom dimensions are truncated to the wrong limit
See report at: https://stackoverflow.com/questions/52973839/nodejs-application-insigths-string-property-suddenly-limited-to-1024

https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/Library/Util.ts#L13 sets the limit for custom dimension properties values to length 1024. However, the limit specified in the schema is 8192 (see https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/EventData.bond#L19)

camous:
fix #444 
align value on backend specifications : https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/EventData.bond#L19

### DIFF
--- a/Library/Util.ts
+++ b/Library/Util.ts
@@ -10,7 +10,7 @@ import RequestResponseHeaders = require("./RequestResponseHeaders");
 
 
 class Util {
-    public static MAX_PROPERTY_LENGTH = 1024;
+    public static MAX_PROPERTY_LENGTH = 8192;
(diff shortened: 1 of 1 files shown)


camous:
seems there is a travis issue on one build step. will close & re-open the PR for another try.

OsvaldoRosado:
Thank you for the PR!

PR #453, opened 2018-11-14
Add autocollected flag to HttpRequests

### LINKED ISSUE #446: Apps with socket.io have 3x duplicate telemetry for incoming HTTP requests
Socket.io creates its own listeners on the HTTP server passed in, which later call the users’s own listeners. The SDK patches each listener, causing the unintentional duplication at each level.

We perform a simple check for avoiding duplication on outgoing HTTP dependencies (https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/AutoCollection/HttpDependencies.ts#L61) but no such check is done for incoming HTTP requests. Adding this simple check for ensuring the request object was not acted upon yet would fix this bug. In the meantime, it can also be implemented within a telemetry processor for anyone hitting this bug.

The bug can be reproduced with:
```
const appInsights = require('applicationinsights');
appInsights.setup('ikey').start();

const express = require('express');
const app = express();
const server = require('http').createServer(app);
const io = require('socket.io')(server);

const router = express.Router();

router.get('/test', (req, res, next) => {
        res.status(200).send('Hello!');
});

app.use('/', router);


io.on('connection', function(){ 

});
server.listen(3000);
```

markwolff:
Fixes #446 
Add field to incoming request to mark it as already collected

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -84,13 +84,13 @@ class AutoCollectHttpRequests {
                 throw new Error('onRequest handler must be a function');
             }
             return (request:http.ServerRequest, response:http.ServerResponse) => {
-                const shouldCollect: boolean = !(<any>request)[AutoCollectHttpRequests.alreadyAutoCollectedFlag];
-
-                // Set up correlation context
-                const requestParser = new HttpRequestParser(request);
-                const correlationContext = this._generateCorrelationContext(requestParser);
+                const shouldCollect: boolean = request && !(<any>request)[AutoCollectHttpRequests.alreadyAutoCollectedFlag];
 
                 if (request && shouldCollect) {
+                    // Set up correlation context
+                    const requestParser = new HttpRequestParser(request);
+                    const correlationContext = this._generateCorrelationContext(requestParser);
+
                     // Note: Check for if correlation is enabled happens within this method.
                     // If not enabled, function will directly call the callback.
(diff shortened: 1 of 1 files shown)


AutoCollection/HttpRequests.ts
```diff
@@ -89,7 +91,13 @@ class AutoCollectHttpRequests {
                 // Note: Check for if correlation is enabled happens within this method.
                 // If not enabled, function will directly call the callback.
                 CorrelationContextManager.runWithContext(correlationContext, () => {
-                    if (this._isEnabled) {
+                    let shouldCollect: boolean = !(<any>request)[AutoCollectHttpRequests.alreadyAutoCollectedFlag];
```
OsvaldoRosado: We should do this check before `runWithContext` to avoid introducing unused contexts and thus losing association with dependencies.

AutoCollection/HttpRequests.ts
```diff
@@ -82,28 +84,39 @@ class AutoCollectHttpRequests {
                 throw new Error('onRequest handler must be a function');
             }
             return (request:http.ServerRequest, response:http.ServerResponse) => {
+                const shouldCollect: boolean = !(<any>request)[AutoCollectHttpRequests.alreadyAutoCollectedFlag];
```
OsvaldoRosado: I wonder if it's possible for `request` to be undefined here. Looks like we guard for it later

Issue #461, opened 2018-11-29
Collect app performance metrics


markwolff:
Use newer node features `process.cpuUsage()` when available, instead of `os.cpus()` in order to more accurately measure single app's perf metrics

markwolff:
This was included in #462, and released as a part of the [`1.0.8`](https://github.com/Microsoft/ApplicationInsights-node.js/tree/1.0.8) release.

PR #462, opened 2018-11-30
Measure process cpu usage in Performance counters

markwolff:
#461

### DIFF
--- a/AutoCollection/Performance.ts
+++ b/AutoCollection/Performance.ts
@@ -17,8 +17,8 @@ class AutoCollectPerformance {
     private _handle: NodeJS.Timer;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
-    private _lastAppCpuUsage: { user: number, system: number } = {user: 0, system: 0};
-    private _lastHrtime: number[] = process.hrtime();
+    private _lastAppCpuUsage: { user: number, system: number };
+    private _lastHrtime: number[];
     private _lastCpus: { model: string; speed: number; times: { user: number; nice: number; sys: number; idle: number; irq: number; }; }[];
     private _lastRequests: { totalRequestCount: number; totalFailedRequestCount: number; time: number; };
 
@@ -46,6 +46,10 @@ class AutoCollectPerformance {
                     totalFailedRequestCount: AutoCollectPerformance._totalFailedRequestCount,
                     time: +new Date
                 };
+                if (typeof (process as any).cpuUsage === 'function'){
+                    this._lastAppCpuUsage = (process as any).cpuUsage();
+                }
+                this._lastHrtime = process.hrtime();
 
                 this._handle = setInterval(() => this.trackPerformance(), 60000);
                 this._handle.unref(); // Allow the app to terminate even while this loop is going on
@@ -144,9 +148,12 @@ class AutoCollectPerformance {
                 const hrtime = process.hrtime();
 
                 const totalApp = ((appCpuUsage.user - this._lastAppCpuUsage.user) + (appCpuUsage.system - this._lastAppCpuUsage.system)) || 0;
-                const elapsedTime = ((hrtime[0] - this._lastHrtime[0])*1e6 + (hrtime[1] - this._lastHrtime[1])/1e3) || 0; // convert to microseconds
 
-                appCpuPercent = 100 * totalApp / (elapsedTime * cpus.length);
+                if (typeof this._lastHrtime !== 'undefined' && this._lastHrtime.length === 2) {
+                    const elapsedTime = ((hrtime[0] - this._lastHrtime[0])*1e6 + (hrtime[1] - this._lastHrtime[1])/1e3) || 0; // convert to microseconds
+
+                    appCpuPercent = 100 * totalApp / (elapsedTime * cpus.length);
+                }
 
                 // Set previous
                 this._lastAppCpuUsage = appCpuUsage;
--- a/AutoCollection/Performance.ts
+++ b/AutoCollection/Performance.ts
@@ -17,6 +17,8 @@ class AutoCollectPerformance {
     private _handle: NodeJS.Timer;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
+    private _lastAppCpuUsage: { user: number, system: number } = {user: 0, system: 0};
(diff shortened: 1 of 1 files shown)


AutoCollection/Performance.ts
```diff
@@ -17,6 +17,8 @@ class AutoCollectPerformance {
     private _handle: NodeJS.Timer;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
+    private _lastAppCpuUsage: { user: number, system: number } = {user: 0, system: 0};
```
OsvaldoRosado: Nit: Because you did = here, you don't need to define the type explicitly. But also, it doesn't seem to be the pattern in this file to do assignment here.

OsvaldoRosado: I also wonder if these should be statics. I haven't thought it through very much, but CPU usage isn't really going to change from class instance to class instance. Not that it matters very much because only one instance of this class should exist anyway..

AutoCollection/Performance.ts
```diff
@@ -17,6 +17,8 @@ class AutoCollectPerformance {
...
+    private _lastHrtime: number[] = process.hrtime();
```
OsvaldoRosado: Probably don't want to record this until `enable` is called. Otherwise the first sample will be over a very long time range

AutoCollection/Performance.ts
```diff
@@ -135,10 +137,26 @@ class AutoCollectPerformance {
...
             }
 
+            // Calculate % of total cpu time (user + system) this App Process used (Only supported by node v6.1.0+)
+            let appCpuPercent: number = undefined;
+            if (typeof (process as any).cpuUsage === 'function') {
+                const appCpuUsage = (process as any).cpuUsage();
+                const hrtime = process.hrtime();
+
+                const totalApp = ((appCpuUsage.user - this._lastAppCpuUsage.user) + (appCpuUsage.system - this._lastAppCpuUsage.system)) || 0;
+                const elapsedTime = ((hrtime[0] - this._lastHrtime[0])*1e6 + (hrtime[1] - this._lastHrtime[1])/1e3) || 0; // convert to microseconds
```
OsvaldoRosado: Should we if guard this with the existence of `lastHrtime`?

AutoCollection/Performance.ts
```diff
@@ -135,10 +137,26 @@ class AutoCollectPerformance {
                 totalIrq += irq;
             }
 
+            // Calculate % of total cpu time (user + system) this App Process used (Only supported by node v6.1.0+)
```
OsvaldoRosado: Should we if guard the section above when the section below would be used? Just to avoid pointless arithmetic that we're going to throw away

AutoCollection/Performance.ts
```diff
@@ -46,6 +46,10 @@ class AutoCollectPerformance {
                     totalFailedRequestCount: AutoCollectPerformance._totalFailedRequestCount,
                     time: +new Date
                 };
+                if (typeof (process as any).cpuUsage === 'function'){
+                    this._lastAppCpuUsage = (process as any).cpuUsage();
```
OsvaldoRosado: Do we not have types for cpuUsage?

markwolff: nope. It uses 4.2.4 in package.json, this type isn't available until 6.1.0
```
"@types/node": "4.2.4",
```

AutoCollection/Performance.ts
markwolff: added a guard for both calculations

AutoCollection/Performance.ts
markwolff: Moved to where other _lastX vars are set to initial condition

AutoCollection/Performance.ts
OsvaldoRosado: If you'd like, we can probably upgrade this package

AutoCollection/Performance.ts
```diff
@@ -135,10 +141,29 @@ class AutoCollectPerformance {
...
+            // Calculate % of total cpu time (user + system) this App Process used (Only supported by node v6.1.0+)
+            let appCpuPercent: number = undefined;
+            if (typeof (process as any).cpuUsage === 'function') {
+                const appCpuUsage = (process as any).cpuUsage();
+                const hrtime = process.hrtime();
+
+                const totalApp = ((appCpuUsage.user - this._lastAppCpuUsage.user) + (appCpuUsage.system - this._lastAppCpuUsage.system)) || 0;
+
+                if (typeof this._lastHrtime !== 'undefined' && this._lastHrtime.length === 2) {
+                    const elapsedTime = ((hrtime[0] - this._lastHrtime[0])*1e6 + (hrtime[1] - this._lastHrtime[1])/1e3) || 0; // convert to microseconds
```
reyang: minor styling question, do we want space before/after * and /?

AutoCollection/Performance.ts
```diff
@@ -135,10 +141,29 @@ class AutoCollectPerformance {
                 totalIrq += irq;
             }
 
+            // Calculate % of total cpu time (user + system) this App Process used (Only supported by node v6.1.0+)
+            let appCpuPercent: number = undefined;
+            if (typeof (process as any).cpuUsage === 'function') {
+                const appCpuUsage = (process as any).cpuUsage();
+                const hrtime = process.hrtime();
+
+                const totalApp = ((appCpuUsage.user - this._lastAppCpuUsage.user) + (appCpuUsage.system - this._lastAppCpuUsage.system)) || 0;
```
reyang: curious, why do we have "|| 0" here, just in case something is undefined?

AutoCollection/Performance.ts
```diff
@@ -135,10 +141,29 @@ class AutoCollectPerformance {
...
+
+                if (typeof this._lastHrtime !== 'undefined' && this._lastHrtime.length === 2) {
```
reyang: do we still need the typeof x !== 'undefined" trick for legacy browsers, or we are good to use !== undefined?

OsvaldoRosado: Browsers aren't a concern for this SDK, so checking just against undefined should be fine. Though IMO just `!this._lastHrtime` is sufficient if we wanted to be really concise.

PR #464, opened 2018-12-05
Remove AppInsights from Error Stack

markwolff:
- Remove Zone artifacts from Error stack shown in console
- Modify error pointer shown in console to inform user that the error was rethrown and originated elsewhere

### DIFF
--- a/AutoCollection/CorrelationContextManager.ts
+++ b/AutoCollection/CorrelationContextManager.ts
@@ -30,7 +30,7 @@ export interface CorrelationContext {
         parentId: string; // Always used for dependencies, may be ignored in favor of incoming headers for requests
     };
 
-    /** Do not store sensitive information here. 
+    /** Do not store sensitive information here.
      *  Properties here are exposed via outgoing HTTP headers for correlating data cross-component.
      */
     customProperties: CustomProperties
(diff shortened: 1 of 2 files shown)


AutoCollection/CorrelationContextManager.ts
```diff
@@ -244,6 +244,9 @@ export class CorrelationContextManager {
                 if (stackFrames.length > 3) {
                     if (stackFrames[2].trim().indexOf("at Error.AppInsightsAsyncCorrelatedErrorWrapper") === 0) {
                         stackFrames.splice(2, 1);
+                    } else if (stackFrames[1].trim().indexOf("at AppInsightsAsyncCorrelatedErrorWrapper.ZoneAwareError") === 0
```
OsvaldoRosado: Was this logic previously covered by unit tests? If so, extra cases should be added for this.

markwolff:
@OsvaldoRosado Do we want to drop this PR if considering zone --> cls?

OsvaldoRosado:
@markwolff I think this will be useful in the meantime before that change happens, especially since it's a fairly trivial change with a lot of positive impact.

markwolff:
@GuyHarwood just published 1.0.8

Issue #452, opened 2018-11-09
Track Requests and Dependencies mannually but Request 


aparralorenzo:
I have a call from an express-nodejs  service to a next component and I would like to link nodejs dependency with next component request. I am already able to link browser dependency with express-nodejs request.

 1.	If I use tagoverride instead clientAI.context.tags[operationIdKey] = operationId we miss operation id. 
![image](https://user-images.githubusercontent.com/3873409/48277661-977c8380-e44b-11e8-9c6f-e90bd2081e6a.png)
If we uncomment clientAI.context.tags[operationIdKey] = operationId we get operation id but I have read that we won´t should use this to avoid problems with handling multiple requests at the same time.
![image](https://user-images.githubusercontent.com/3873409/48277727-d0b4f380-e44b-11e8-92a5-43d1926130ed.png)
How can I solve this?
3. When I console log clientAI.getCorrelationContext() it get error. Same issue here https://github.com/Microsoft/ApplicationInsights-node.js/issues/323#issue-257545223 but I´m not able to solve it even if I change the import of the appinsighs to the first. Any idea?
4. How can I get the dependency id from this express-node js service to pass throw the http call to the next component in the request-id header to be able to have the end to end correlation working?

This is part of my code:

./init/initApplicationInsights
```javascript
const config = require('../config');
const appInsights = require('applicationinsights');

module.exports = function initApplicationInsights() {
  if (config.INSTRUMENTATION_KEY) {
    appInsights.setup(config.INSTRUMENTATION_KEY)
      .setAutoDependencyCorrelation(true)
      .setAutoCollectRequests(false)
      .setAutoCollectPerformance(true)
      .setAutoCollectExceptions(true)
      .setAutoCollectDependencies(false)
      .setAutoCollectConsole(true)
      .setUseDiskRetryCaching(true);
    appInsights.defaultClient.context.tags[appInsights.defaultClient.context.keys.cloudRole] =
      'fes-air-availability';
    appInsights.start();
    return appInsights.defaultClient;
  }
  return null;
};
```
app.js
```javascript
const clientAI = require('./init/initApplicationInsights')();
    const requestId = req.headers["request-id"];
    const operationId = requestId.substr(1, requestId.indexOf('.')-1);  
    console.log("requestId " + requestId);  // |OjuCU.ClP5n
    console.log("operationId " + operationId);   // OjuCU
    const operationIdKey = clientAI.context.keys.operationId; 
[...]

PR #474, opened 2019-01-04
Deprecate zonejs dependency

markwolff:
- Deprecate zonejs and switch to [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked)
- add [`startOperation`](https://github.com/Microsoft/ApplicationInsights-node.js/issues/452#issuecomment-438383610)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,11 @@ services:
   - docker
 node_js:
   - "node"
+  - "11"
+  - "10"
   - "9"
+  - "8.2.0"
+  - "8.2.1"
   - "8"
   - "7"
   - "6"
@@ -21,7 +25,13 @@ before_script:
 script:
   - npm test
   # Only run funcs for LTS releases
+  - if [ "$TRAVIS_NODE_VERSION" = "11" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "10" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "9" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "8.2.1" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "8.2.0" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "8" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "7" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "6" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "4" ]; then npm run functionaltest; fi
   # Only run backcompat tests once
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,19 +16,23 @@ node_js:
   - "0.11"
   - "0.10"
   - "iojs"
-  - "iojs-v1.0.4" 
+  - "iojs-v1.0.4"
 before_install:
   - npm install -g npm@4.5.0
 before_script:
   - npm install
 script:
   - npm test
-  # Only run funcs for LTS releases
+  # Run funcs for current release
   - if [ "$TRAVIS_NODE_VERSION" = "11" ]; then npm run functionaltest; fi
+  # Run funcs for LTS releases (currently 6, 8, 10)
   - if [ "$TRAVIS_NODE_VERSION" = "10" ]; then npm run functionaltest; fi
-  - if [ "$TRAVIS_NODE_VERSION" = "8.0.0" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "8" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "6" ]; then npm run functionaltest; fi
   - if [ "$TRAVIS_NODE_VERSION" = "4" ]; then npm run functionaltest; fi
+  # 8.0.0 uses async-listener for dependency correlation instead of async_hooks
+  # as in latest node8, so run funcs for 8.0.0 as well
+  - if [ "$TRAVIS_NODE_VERSION" = "8.0.0" ]; then npm run functionaltest; fi
+
   # Only run backcompat tests once
   - if [ "$TRAVIS_NODE_VERSION" = "4" ]; then npm run backcompattest; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,12 @@ node_js:
   - "11"
   - "10"
   - "9"
-  - "8.2.0"
-  - "8.2.1"
   - "8"
+  - "8.2.1"
+  - "8.2.0"
+  - "8.1.3"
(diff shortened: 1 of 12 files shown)


AutoCollection/CorrelationContextManager.ts
```diff
@@ -39,6 +41,8 @@ export interface CorrelationContext {
 export class CorrelationContextManager {
     private static enabled: boolean = false;
     private static hasEverEnabled: boolean = false;
+    private static session: any;
+    private static cls: any;
```
OsvaldoRosado: You can use the import style seen in the top of `https://github.com/Microsoft/ApplicationInsights-node.js/blob/develop/AutoCollection/diagnostic-channel/initialization.ts` to get the proper types for this without requiring the cls-hooked package in the resulting JS.

AutoCollection/CorrelationContextManager.ts
```diff
@@ -49,7 +53,12 @@ export class CorrelationContextManager {
         if (!CorrelationContextManager.enabled) {
             return null;
         }
-        return Zone.current.get("context");
+        const context = CorrelationContextManager.session.get('context');
```
OsvaldoRosado: I wonder if we should choose a more AI-specific name than just `context` for this. I know it's also what we used in Zone, and it didn't seem to cause trouble there, but I'm thinking its possible to have conflicts if we're instrumenting an app that already uses cls.

AutoCollection/CorrelationContextManager.ts
```diff
@@ -116,30 +124,23 @@ export class CorrelationContextManager {
...
-                // averting require's cache somehow.
-                if (typeof Zone === "undefined") {
-                    require("zone.js");
+                if (typeof this.cls === "undefined") {
+                    this.cls = require('cls-hooked');
                 }
             } catch (e) {
-                // Zone was already loaded even though we couldn't find its global variable
-                Logging.warn("Failed to require zone.js");
+                // cls-hooked was already loaded even though we couldn't find its global variable
```
OsvaldoRosado: It looks like cls-hooked doesn't use a global ambient variable like Zone did. This try/catch might not be necessary. Even if we keep it, the comment is now slightly wrong :smile:

AutoCollection/CorrelationContextManager.ts
```diff
@@ -116,30 +124,23 @@ export class CorrelationContextManager {
...
+                Logging.warn("Failed to require cls-hooked");
             }
 
+            CorrelationContextManager.session = this.cls.createNamespace("AI-CLS-Session");
```
OsvaldoRosado: Out of curiosity, what does this do?

OsvaldoRosado: Also note, you're outside of the try/catch here. this.cls might be null.

OsvaldoRosado:
Should our HTTP patching be updated as well to call `bindEmitter`?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -161,137 +161,6 @@ export class CorrelationContextManager {
         var nodeVer = process.versions.node.split(".");
         return parseInt(nodeVer[0]) > 3 || (parseInt(nodeVer[0]) > 2 && parseInt(nodeVer[1]) > 2);
```
OsvaldoRosado: This version check needs to be bumped higher. [package.json of `cls-hooked`](https://github.com/Jeff-Lewis/cls-hooked/blob/master/package.json) says >= 4.7.0

AutoCollection/CorrelationContextManager.ts
```diff
@@ -97,6 +97,13 @@ export class CorrelationContextManager {
         }
     }
 
+    /**
+     * Wrapper for cls-hooked bindEmitter method
+     */
+    public static bindEmitter(reqres: any): void {
```
OsvaldoRosado: Should we name this `wrapEmitter` to be consistent with `wrapCallback`? Also, shouldn't the type here be `EventEmitter`?

OsvaldoRosado: Minor nit: Would prefer a generic argument name like `emitter` to `reqres`

OsvaldoRosado: This also needs if guards to only act when this is enabled

AutoCollection/HttpRequests.ts
```diff
@@ -90,6 +90,8 @@ class AutoCollectHttpRequests {
                     // Set up correlation context
                     const requestParser = new HttpRequestParser(request);
                     const correlationContext = this._generateCorrelationContext(requestParser);
+                    CorrelationContextManager.bindEmitter(request);
```
OsvaldoRosado: I wonder if we should do this in the auto collector for HTTP dependencies as well.

tsconfig.json
```diff
@@ -17,7 +17,7 @@
         "Tests/**/*.ts"
     ],
     "files": [
-        "./node_modules/zone.js/dist/zone.js.d.ts"
+        "./node_modules/@types/cls-hooked/index.d.ts"
```
OsvaldoRosado: Since cls' types don't appear to be ambient, you might not need this at all

.travis.yml
```diff
@@ -21,6 +24,9 @@ before_script:
 script:
   - npm test
   # Only run funcs for LTS releases
+  - if [ "$TRAVIS_NODE_VERSION" = "11" ]; then npm run functionaltest; fi
```
OsvaldoRosado: Nit: 11 isn't an LTS release. It's okay if we want to establish a new criteria for what belongs here, but the comment should be updated

.travis.yml
```diff
@@ -21,6 +24,9 @@ before_script:
...
+  - if [ "$TRAVIS_NODE_VERSION" = "10" ]; then npm run functionaltest; fi
+  - if [ "$TRAVIS_NODE_VERSION" = "8.0.0" ]; then npm run functionaltest; fi
```
OsvaldoRosado: I wonder if we should have a comment explaining why we add 8.0.0 in particular in addition to 8

.travis.yml
markwolff: updated to run func tests on current release as well

.travis.yml
markwolff: Added comment

Tests/FunctionalTests/Runner/TestValidation.js
```diff
@@ -83,7 +83,7 @@ module.exports.TestValidation = class TestValidation {
                 // Did we find all of the items in the data set?
                 if (dataSet.length > 1){
                     Utils.Logging.error("FAILED EXPECTATION - Unexpected child telemetry item(s)!");
-                    Utils.Logging.error(JSON.stringify(dataSet));
+                    Utils.Logging.error(JSON.stringify(dataSet, null, 2));
```
OsvaldoRosado: What does this change do?

OsvaldoRosado:
Not a blocker for this PR. But before a stable release we should expose a switch on `setAutoDependencyCorrelation` to force `continuation-local-storage` vs `cls-hooked`.

Tests/FunctionalTests/Runner/TestValidation.js
markwolff: pretty prints the errors for my sake

PR #476, opened 2019-01-11
bump version to 1.1.0

markwolff:


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "applicationinsights",
   "license": "MIT",
   "bugs": "https://github.com/Microsoft/ApplicationInsights-node.js/issues",
-  "version": "1.0.8",
+  "version": "1.1.0",
```
OsvaldoRosado: Should this be `1.1.0-beta.0` instead?

PR #478, opened 2019-01-15
fix bad requests causing unit tests to hang

### LINKED ISSUE #475: Perform UnitTest with real application insights too long
Hello,

I'm working on a Hapi server application.
I try to integrate the package `applicationinsights` to my application.

My application is cover by unit test (mocha), but when I use a real instrumentation key the unit test execution is quite long ( >60s) and when the instrumentation key is a fake `ikey` then the execution is more performant (<3s).

To help you to reproduce and understand my issue, I have push the code in this branch
https://github.com/guenogu/repro/tree/hapi-wreck-appinsights

Thanks for your help.

markwolff:
Fixes #475

- Fix hanging mocha tests
  - Stubbed some `http(s).request` methods in unit tests (fixes our unit tests)
  - Fix stubbed http request from returning undefined statusCode
  - Unref diskretry timer so that we don't block instrumented app from terminating 
- Upgrade to latest mocha version (edit: Staying with 3.5 because it breaks node 0.x, but added `--no-exit`)

[Mocha 4+ does not force quit tests on termination](https://mochajs.org/#--exit----no-exit), so tests involving application insights would hang for 30-60 seconds due to `fetchAppId` getting an undefined status code and retrying.

### DIFF
--- a/Library/CorrelationIdManager.ts
+++ b/Library/CorrelationIdManager.ts
@@ -65,7 +65,7 @@ class CorrelationIdManager {
                         }
                         delete CorrelationIdManager.pendingLookups[appIdUrlString];
                     });
-                } else if (res.statusCode >= 400 && res.statusCode < 500) {
+                } else if (res.statusCode >= 400 && res.statusCode < 500 || res.statusCode === undefined) {
                     // Not found, probably a bad key. Do not try again.
                     CorrelationIdManager.completedLookups[appIdUrlString] = undefined;
                     delete CorrelationIdManager.pendingLookups[appIdUrlString];
--- a/Library/CorrelationIdManager.ts
+++ b/Library/CorrelationIdManager.ts
@@ -65,7 +65,7 @@ class CorrelationIdManager {
                         }
                         delete CorrelationIdManager.pendingLookups[appIdUrlString];
                     });
-                } else if (res.statusCode >= 400 && res.statusCode < 500 || res.statusCode === undefined) {
+                } else if (res.statusCode >= 400 && res.statusCode < 500) {
                     // Not found, probably a bad key. Do not try again.
                     CorrelationIdManager.completedLookups[appIdUrlString] = undefined;
                     delete CorrelationIdManager.pendingLookups[appIdUrlString];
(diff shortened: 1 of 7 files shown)


Library/CorrelationIdManager.ts
```diff
@@ -65,7 +65,7 @@ class CorrelationIdManager {
                         }
                         delete CorrelationIdManager.pendingLookups[appIdUrlString];
                     });
-                } else if (res.statusCode >= 400 && res.statusCode < 500) {
+                } else if (res.statusCode >= 400 && res.statusCode < 500 || res.statusCode === undefined) {
```
OsvaldoRosado: Can undefined show up for reasons that _should_ retry? Like an abrupt disconnection

markwolff: I'm going based off of this
https://stackoverflow.com/questions/29752225/http-statuscode-sometimes-undefined
I haven't been able to repro this outside of unit tests, but I'm still hesitant of it because it seems like undefined behavior

Library/Sender.ts
```diff
@@ -135,7 +135,7 @@ class Sender {
                     if (this._enableDiskRetryMode) {
                         // try to send any cached events if the user is back online
                         if (res.statusCode === 200) {
-                            setTimeout(() => this._sendFirstFileOnDisk(), this._resendInterval);
+                            setTimeout(() => this._sendFirstFileOnDisk(), this._resendInterval).unref();
```
jpiyali: would this not cause the process to exist and not retry subsequently?

OsvaldoRosado: I think this is good behavior of the SDK. If the only reason for it keeping the app running is to check for stored telemetry on the disk every 60 seconds, it should let the app close. Quick closing Node scripts are especially important where you pay for execution time (Lambda / Functions / etc)

markwolff: This only prevents the process from not exiting if it tries to, as per [here](https://nodejs.org/api/timers.html#timers_timeout_unref). After this change, the only thing we block termination on is for pending telemetry in the buffer

jpiyali: Thanks for the explanation @OsvaldoRosado , @markwolff . So, this is intentional not to check for pending telemetry if thats the only thing keeping it active.

PR #479, opened 2019-01-23
add useAsyncHooks arg to CorrelationContextManager

markwolff:
https://github.com/Microsoft/ApplicationInsights-node.js/compare/markwolff/add-force-cls?expand=1#diff-1452dad132989865d44d978755072970L210

- `true`: use `cls-hooked`
- `false`: use `continuation-local-storage`
- `undefined`: select library based on `node` version

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -282,6 +282,8 @@ class AutoCollectHttpRequests {
          AutoCollectHttpRequests.INSTANCE = null;
          this.enable(false);
          this._isInitialized = false;
+         CorrelationContextManager.disable();
+         this._isAutoCorrelating = false;
     }
 }
 
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -43,9 +43,9 @@ class AutoCollectHttpRequests {
         }
     }
 
-    public useAutoCorrelation(isEnabled:boolean) {
+    public useAutoCorrelation(isEnabled:boolean, forceClsHooked?:boolean) {
         if (isEnabled && !this._isAutoCorrelating) {
-            CorrelationContextManager.enable();
+            CorrelationContextManager.enable(forceClsHooked);
         } else if (!isEnabled && this._isAutoCorrelating) {
             CorrelationContextManager.disable();
         }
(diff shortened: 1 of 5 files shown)


applicationinsights.ts
```diff
@@ -207,10 +208,11 @@ export class Configuration {
      * @param value if true dependencies will be correlated with requests
      * @returns {Configuration} this class
      */
-    public static setAutoDependencyCorrelation(value: boolean) {
+    public static setAutoDependencyCorrelation(value: boolean, forceClsHooked?: boolean) {
```
OsvaldoRosado: I wouldn't tie our public API surface to the specific 3rd party library providing correlation (which we could otherwise swap out at will). Consider using something like `useAsyncHooks` instead.

applicationinsights.ts
```diff
@@ -207,10 +208,11 @@ export class Configuration {
      * @param value if true dependencies will be correlated with requests
```
OsvaldoRosado: JSDoc here needs updating

AutoCollection/HttpRequests.ts
```diff
@@ -282,6 +282,8 @@ class AutoCollectHttpRequests {
          AutoCollectHttpRequests.INSTANCE = null;
          this.enable(false);
          this._isInitialized = false;
+         CorrelationContextManager.disable();
```
OsvaldoRosado: I'm curious. What does this change do?

markwolff: the call to `disable` isn't necessary, but I needed to add `this._isAutoCorrelation = false` in order to be able to load one of the cls libraries in tests. Since I was setting isAutoCorrelation to false, it just made sense to actually disable auto correlation here. 👍

applicationinsights.ts
markwolff: updated

applicationinsights.ts
markwolff: updated to `useAsyncHooks`

applicationinsights.ts
```diff
@@ -205,12 +206,14 @@ export class Configuration {
     /**
      * Sets the state of automatic dependency correlation (enabled by default)
      * @param value if true dependencies will be correlated with requests
+     * @param useAsyncHooks if true, forces use of experimental async_hooks module to provide correlation. If false, instead uses async-listener. If left blank, the best option is chosen for you based on your version on NodeJS.
```
OsvaldoRosado: Nit: "on your version on NodeJS" => "on your version **of** Node.js"
Also consider "instead uses async-listener" => "instead uses only patching-based techniques" or something along those lines in case we'd like to replace `async-listener` in particular.

markwolff: Updated

PR #492, opened 2019-02-21
update setup to overwrite existing defaultClient

markwolff:


applicationinsights.ts
```diff
@@ -53,16 +53,16 @@ export let defaultClient: TelemetryClient;
...
-        _performance = new AutoCollectPerformance(defaultClient);
-        _serverRequests = new AutoCollectHttpRequests(defaultClient);
-        _clientRequests = new AutoCollectHttpDependencies(defaultClient);
-    } else {
-        Logging.info("The default client is already setup");
+    if(defaultClient) {
+        Logging.info("setup was called previously. Overwriting existing defaultClient");
+        dispose();
     }
+    defaultClient = new TelemetryClient(instrumentationKey);
```
OsvaldoRosado: Even though you call dispose, I'd still be concerned about a reference to the old client sticking around somewhere - preventing gc. Consider resetting the config on the existing telemetryclient rather than constructing from scratch.

markwolff: Updated to only reconfigure. All of the other objects are okay to reconstruct since they shouldn't have references elsewhere

applicationinsights.ts
```diff
@@ -53,17 +53,28 @@ export let defaultClient: TelemetryClient;
...
-        _exceptions = new AutoCollectExceptions(defaultClient);
-        _performance = new AutoCollectPerformance(defaultClient);
-        _serverRequests = new AutoCollectHttpRequests(defaultClient);
-        _clientRequests = new AutoCollectHttpDependencies(defaultClient);
+    if(defaultClient) {
+        Logging.info("setup was called previously. Overwriting existing defaultClient");
+
+        // The 'if checks' here aren't necessary, but guard against
+        // any potential abnormal setup
+        if (_console) _console.dispose();
```
OsvaldoRosado: I'm curious. Do we need to dispose and recreate these? Shouldn't they work just fine if left alone (eg. calling this fn twice only accomplishes changing the current ikey more or less)

PR #493, opened 2019-03-01
add env variable to disable dep correlation

markwolff:
Currently `useAutoCorrelation()` does the check for the new environment variable. Am wondering if it should be moved to `CorrelationContextManager.enable()` instead, or the 2 places that call `useAutoCorrelation()` in `applicationinsights.ts` instead.

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -10,6 +10,7 @@ import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import HttpRequestParser = require("./HttpRequestParser");
 import { CorrelationContextManager, CorrelationContext, PrivateCustomProperties } from "./CorrelationContextManager";
 import AutoCollectPerformance = require("./Performance");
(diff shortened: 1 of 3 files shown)


AutoCollection/HttpRequests.ts
```diff
@@ -44,7 +45,7 @@ class AutoCollectHttpRequests {
     }
 
     public useAutoCorrelation(isEnabled:boolean, forceClsHooked?:boolean) {
-        if (isEnabled && !this._isAutoCorrelating) {
+        if (isEnabled && !this._isAutoCorrelating && !process.env[Config.ENV_disableDependencyCorrelation]) {
```
OsvaldoRosado: What's the intention of this env var? To provide a way to change the default setting to disabled? Or to override explicit enabling in code?

I was expecting the former, though this looks like the latter. Totally fine if intentional.

markwolff: Intentionally the latter. I'll look into whether or not or not this is the behavior of the other similar env vars

Issue #489, opened 2019-02-20
Uncaught exception in library when handling exception with no stack trace


js-kyle:
When promises are rejected without an error, the application crashes with the below Uncaught exception:

```
node_modules/applicationinsights/out/AutoCollection/Exceptions.js:34
                        throw THIS_IS_APPLICATION_INSIGHTS_RETHROWING_YOUR_EXCEPTION; // Error originated somewhere else in your app
                        ^

TypeError: Cannot read property 'stack' of undefined
    at Function.EnvelopeFactory.createExceptionData (/Users/kylemartin/Desktop/node_modules/applicationinsights/out/Library/EnvelopeFactory.js:132:40)
    at Function.EnvelopeFactory.createEnvelope (/Users/kylemartin/Desktop/node_modules/applicationinsights/out/Library/EnvelopeFactory.js:32:40)
    at NodeClient.TelemetryClient.track (/Users/kylemartin/Desktop/node_modules/applicationinsights/out/Library/TelemetryClient.js:102:44)
    at NodeClient.TelemetryClient.trackException (/Users/kylemartin/Desktop/node_modules/applicationinsights/out/Library/TelemetryClient.js:55:14)
    at AutoCollectExceptions.handle (/Users/kylemartin/Desktop/node_modules/applicationinsights/out/AutoCollection/Exceptions.js:30:35)
    at process.emit (events.js:182:13)
    at emitPromiseRejectionWarnings (internal/process/promises.js:119:20)
    at process._tickCallback (internal/process/next_tick.js:69:34)
    at Function.Module.runMain (internal/modules/cjs/loader.js:744:11)
    at startup (internal/bootstrap/node.js:285:19)
```

This can be reproduced on `1.2.0` as well as older versions, by:

test.js
```
let appInsights = require('applicationinsights');
appInsights.setup().start();

Promise.reject();
```

`APPINSIGHTS_INSTRUMENTATIONKEY="string" node test.js`

OsvaldoRosado:
Thanks for the report! It looks like we're also unintentionally suppressing important Node warnings from this scenario. For example, when I run this sample without the AI SDK, Node shows this warning (though does not crash).

```
(node:1808) UnhandledPromiseRejectionWarning: undefined
(node:1808) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
[...]

PR #494, opened 2019-03-15
add live metrics

markwolff:
Spins up a separate performance collector with `liveMetricsClient`, where `trackMetric` gets routed to Live Metrics instead. 

In addition, `quickPulseTelemetryProcessor` intercepts all telemetry and pipes it as a `document` (if supported by live metrics) to `liveMetricsClient`. This telemetry processor will increment the corresponding performance counters for `exception`, `dependency`. `request` does not increment here because it is incremented in `HttpRequest.ts`

> ***Note:*** Currently set to off by default


Declarations/Constants.ts
```diff
@@ -0,0 +1,55 @@
...
+    PROCESSOR_TIME= "\\Processor(_Total)\\% Processor Time",
+    PROCESS_TIME= "\\Process(??APP_WIN32_PROC??)\\% Processor Time",
+
+    // Requests
+    REQUEST_RATE= "\\ASP.NET Applications(??APP_W3SVC_PROC??)\\Requests/Sec",
+    REQUEST_DURATION= "\\ASP.NET Applications(??APP_W3SVC_PROC??)\\Request Execution Time"
+};
+
+export function mapPerformanceCounterToQuickPulseCounter(inPerformance: PerformanceCounter): QuickPulseCounter {
+    for (let key in QuickPulseCounter) {
```
OsvaldoRosado: These seems kind of wasteful, why not make a static map since this information can never change at runtime?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
 import Context = require("./Context");
 import { CorrelationContextManager } from "../AutoCollection/CorrelationContextManager";
 
+var StreamId = Util.w3cTraceId(); // Create a guid
+
 /**
  * Manages the logic of creating envelopes from Telemetry objects
  */
 class EnvelopeFactory {
+    public static createQuickPulseEnvelope(metrics: Contracts.MetricQuickPulse[], documents: Contracts.DocumentQuickPulse[], config: Config, context: Context): Contracts.EnvelopeQuickPulse {
```
OsvaldoRosado: Should these have been put into a separate class? Eg. QuickPulseEnvelopeFactory?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+        const machineName = os.hostname(); // Note: os.hostname() was added in node v0.3.3
+        const instance = (context.tags
+            && context.keys
+            && context.keys.cloudRoleInstance
+            && context.tags[context.keys.cloudRoleInstance]) || machineName;
+
+        var envelope: Contracts.EnvelopeQuickPulse = {
+            Documents: documents.length > 0 ? documents : null,
+            InstrumentationKey: config.instrumentationKey || "",
```
OsvaldoRosado: Is it possible to not have an ikey?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+        }
+        return null;
+    }
+
+    private static createQuickPulseTraceDocument(envelope: Contracts.Envelope): Contracts.MessageDocumentQuickPulse {
+        const document = EnvelopeFactory.createQuickPulseDocument(envelope);
+
+        var traceDocument: Contracts.MessageDocumentQuickPulse = {
+            ...document,
+            Properties: [{key: 'my key', value: 'my value'}],
```
OsvaldoRosado: Are these supposed to be here?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+            Message: ((envelope.data as any).baseData as Contracts.MessageData).message,
+            SeverityLevel: "Critical"
```
OsvaldoRosado: Intended to be hardcoded?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+        }
+
+        return traceDocument;
+    }
+
+    private static createQuickPulseExceptionDocument(envelope: Contracts.Envelope): Contracts.ExceptionDocumentQuickPulse {
+        const document = EnvelopeFactory.createQuickPulseDocument(envelope);
+
+        let exception = '';
+        ((envelope.data as any).baseData as Contracts.ExceptionData).exceptions[0].parsedStack.forEach(err => {
```
OsvaldoRosado: Is it possible for `exceptions[0]` to be undefined?

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+            DependencyTypeName: baseData.type,
+        }
+        return dependencyDocument;
+    }
+
+    private static createQuickPulseDocument(envelope: Contracts.Envelope): Contracts.DocumentQuickPulse {
+        let documentType, __type, id, operationId;
+
+        switch (envelope.data.baseType) {
+            case "EventData":
```
OsvaldoRosado: We might want to consider enums for all these constants

Library/EnvelopeFactory.ts
```diff
@@ -1,13 +1,163 @@
...
+            case "RequestData":
+                documentType = "Request";
+                break;
+            case "RemoteDependencyData":
+                documentType = "RemoteDependency";
+                __type = "DependencyTelemetryDocument";
+                break;
+        }
+        __type = __type || (documentType + "TelemetryDocument");
+        operationId = envelope.tags["ai.operation.id"];
```
OsvaldoRosado: `"ai.operation.id"` -> `ContextTagKeys.operationId`

Library/QuickPulseSender.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        // Important: When POSTing data, envelope must be an array
+        this._submitData([envelope], done, "post");
+    }
+
+    private _submitData(envelope: Contracts.EnvelopeQuickPulse | Contracts.EnvelopeQuickPulse[], done: (shouldPOST: boolean, res: http.IncomingMessage) => void, postOrPing: "post" | "ping"): void {
+        const payload = JSON.stringify(envelope);
+        var options = {
+            [AutoCollectHttpDependencies.disableCollectionRequestOption]: true,
+            host: "rt.services.visualstudio.com",
```
OsvaldoRosado: This should probably come from config

TelemetryProcessors/QuickPulseTelemetryProcessor.ts
```diff
@@ -0,0 +1,27 @@
...
+                AutoCollectPerformance.countException();
+                break;
+            case "RequestData":
+                // These are already autocounted by HttpRequest.
+                // Note: Not currently counting manual trackRequest calls
+                // here to avoid affecting existing autocollection metrics
+                break;
+            case "RemoteDependencyData":
+                const baseData: Contracts.DependencyTelemetry = (envelope.data as any).baseData;
+                AutoCollectPerformance.countDependency(baseData.duration, baseData.success);
```
OsvaldoRosado: Why not have the RDD autocollector just do this internally, like we do for request counting?

applicationinsights.ts
```diff
@@ -64,6 +68,15 @@ export function setup(instrumentationKey?: string) {
         Logging.info("The default client is already setup");
     }
 
+    if (!liveMetricsClient) {
```
OsvaldoRosado: Shouldn't this only get set up if enabled?

markwolff: The way it is now, it is only constructed. Enablement would happen at `AppInsights.start()`. I did it this way for consistency with `defaultClient`. This is probably not necessary and can be changed if you think it should only be constructed if enabled.

OsvaldoRosado: I'm just thinking a default telemetry client is always necessary to be available, but the live metrics stuff is just eating memory and compute without a purpose in the default state if this is made in `setup`

applicationinsights.ts
```diff
@@ -251,6 +267,18 @@ export class Configuration {
...
         return Configuration;
     }
+
+    /**
+     * Enables communication with Application Insights Live Metrics.
+     * @param enable if true, enables communication with the live metrics service
+     */
+    public static setSendLiveMetrics(enable = false) {
+        if (liveMetricsClient) {
+            _isSendingLiveMetrics = enable;
```
OsvaldoRosado: Is this right? Isn't the pattern
```
if (!liveMetricsClient) {
   _isSendingLiveMetrics = enable;
} else {
   liveMetricsClient.enable(enable);
}
```

markwolff: Good point. Will update

TelemetryProcessors/QuickPulseTelemetryProcessor.ts
markwolff: My thinking was that I also want to send 3rd party tracks as documents, and I don't think those pass through HttpDependencies at all? I could add this logic to `trackDependency`, but since I have to do it for exceptions as well, I kept it all here instead.

Library/EnvelopeFactory.ts
markwolff: Probably only for scenarios like #489, or manual `trackException` calls. Will clean this section up a bit

Library/EnvelopeFactory.ts
markwolff: Placeholder stuff. Need to cleanup

Library/EnvelopeFactory.ts
markwolff: I don't think so. I carried it over from the other envelopecreator stuff, but it probably doesn't need to be there either

TelemetryProcessors/QuickPulseTelemetryProcessor.ts
OsvaldoRosado: Will only quickpulse ever need these stats? Or are they also standard metrics we weren't reporting before? If its only quickpulse, keeping it here sounds good.

markwolff: Assuming you mean trackMetrics I added in `Performance.ts`. There may be extra ones that we aren't sending right now. I was planning on adding those in a separate PR since it'll change the default behavior of our perf telemetry.
https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/01f38eb50ffc05bd4e60c28754e8f5672356a307/Src/PerformanceCollector/Perf.Shared/PerformanceCollectorModule.cs#L154-L171

Declarations/Constants.ts
markwolff: Added map constant

Library/EnvelopeFactory.ts
markwolff: Moved to QuickPulseEnvelopeFactory

Library/EnvelopeFactory.ts
markwolff: Added properties/measurements aggregation

applicationinsights.ts
markwolff: Refactored how client is constructed. Currently constructed on first call of `setSendLiveMetrics`. Subsequent calls enable/disable it. Calling `start()` will enable the client if it exists and should be enabled

Library/QuickPulseSender.ts
OsvaldoRosado: Re-opening because this still isn't from the passed in Config object.

markwolff: Moved to configuration option

Library/EnvelopeFactory.ts
markwolff: updated to add constants and mappings for all of these. default EnvelopeFactory is using numbered enums for these fields, so I created new `TelemetryTypeString` enum for this next to existing one.

Declarations/Constants.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Requests
+    REQUEST_RATE= "\\ASP.NET Applications(??APP_W3SVC_PROC??)\\Requests/Sec",
+    REQUEST_DURATION= "\\ASP.NET Applications(??APP_W3SVC_PROC??)\\Request Execution Time"
+};
+
+/**
+ * Map a PerformanceCounter/QuickPulseCounter to a QuickPulseCounter. If no mapping exists, mapping is *undefined*
+ */
+export const PerformanceToQuickPulseCounter: {[key: string]: QuickPulseCounter} = {
```
OsvaldoRosado: You might be able to make the type of key `PerformanceCounter | QuickPulseCounter`

markwolff: Added `Contracts.TelemetryTypeKeys` and `Contracts.TelemetryTypeValues`

Declarations/Constants.ts
```diff
@@ -0,0 +1,100 @@
+import Contracts = require("../Declarations/Contracts")
+
+export const QuickPulseConfig = {
```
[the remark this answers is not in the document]
markwolff: moved to `QuickPulseSender.ts`

PR #504, opened 2019-04-04
livemetrics: add lastSendSucceeded to prevent permanent retries

markwolff:
- Fixes issue where if live metrics service or a customer's app loses connectivity for more than 20 seconds, sending live metrics wouldn't work unless app is restarted.

### DIFF
--- a/Library/QuickPulseStateManager.ts
+++ b/Library/QuickPulseStateManager.ts
@@ -8,13 +8,19 @@ import Context = require("./Context");
 import * as http from "http";
 import * as Contracts from "../Declarations/Contracts";
 
+
 /** State Container for sending to the QuickPulse Service */
 class QuickPulseStateManager {
     public config: Config;
     public context: Context;
 
-    private static _isCollectingData: boolean = false;
+    private static MAX_POST_WAIT_TIME = 20000;
+    private static MAX_PING_WAIT_TIME = 60000;
+    private static FALLBACK_INTERVAL = 60000;
+    private static PING_INTERVAL = 5000;
+    private static POST_INTERVAL = 1000;
 
+    private _isCollectingData: boolean = false;
     private _sender: QuickPulseSender;
     private _isEnabled: boolean;
     private _lastSuccessTime: number = Date.now();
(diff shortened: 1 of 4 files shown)


Library/QuickPulseStateManager.ts
```diff
@@ -124,14 +125,15 @@ class QuickPulseStateManager {
...
+        if (QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 20000 && !this._lastSendSucceeded) {
             // Haven't posted successfully in 20 seconds, so wait 60 seconds and ping
             QuickPulseStateManager._isCollectingData = false;
             currentTimeout = 60000;
-        } else if (!QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 60000) {
+        } else if (!QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 60000 && !this._lastSendSucceeded) {
             // Haven't pinged successfully in 60 seconds, so wait another 60 seconds
             currentTimeout = 60000;
         }
+        this._lastSendSucceeded = null;
```
ramthi: >null [](start = 34, length = 4)

Why not set this to false?

Library/QuickPulseStateManager.ts
```diff
@@ -124,14 +125,15 @@ class QuickPulseStateManager {
         }
 
         let currentTimeout = QuickPulseStateManager._isCollectingData ? 1000 : 5000;
-        if (QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 20000) {
+        if (QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 20000 && !this._lastSendSucceeded) {
```
OsvaldoRosado: What's the logic here for what's a static vs instance level property?

Library/QuickPulseStateManager.ts
```diff
@@ -124,14 +125,15 @@ class QuickPulseStateManager {
...
             // Haven't posted successfully in 20 seconds, so wait 60 seconds and ping
             QuickPulseStateManager._isCollectingData = false;
             currentTimeout = 60000;
-        } else if (!QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 60000) {
+        } else if (!QuickPulseStateManager._isCollectingData && Date.now() - this._lastSuccessTime >= 60000 && !this._lastSendSucceeded) {
```
OsvaldoRosado: Should the 60s be in a variable somewhere since it's repeated in a few places?

Library/QuickPulseStateManager.ts
markwolff: Meant to convey a "pending" state versus having explicitly failed ("false")

Library/QuickPulseStateManager.ts
```diff
@@ -144,7 +146,7 @@ class QuickPulseStateManager {
         this._sender.post(envelope, this._quickPulseDone.bind(this));
     }
 
-    private _quickPulseDone(shouldPOST: boolean, res: http.IncomingMessage): void {
+    private _quickPulseDone(shouldPOST: boolean, res: http.IncomingMessage | {statusCode: boolean}): void {
```
OsvaldoRosado: What does it mean for status code to be a boolean?

OsvaldoRosado: Also, I suspect this may be buggy, since a status code of `false` will pass the success if below, because `false < 300` is `true` (somehow)

markwolff: Meant to make it optional so that I can stop sending to QPS when the requests send an error. [`done(false)`](https://github.com/Microsoft/ApplicationInsights-node.js/pull/504/files#diff-50b8f385a3e9074029ab82dd36ecfa8cR56)

Library/QuickPulseStateManager.ts
markwolff: Added static consts

Library/QuickPulseStateManager.ts
markwolff: Moved to have everything as instance level

Issue #510, opened 2019-04-24
Unable to contact qps endpoint, dropping this Live Metrics packet


KrishnaPravin:
I am using ApplicationInsights-node.js version 1.3.1.
I am using it in express node server, with liveMetrics enabled. `appInsights.setup(_ikey_).setSendLiveMetrics(true).start()`

For couple of minutes live metrics is working fine, but after that I am getting ECONNRESET error.
I restarted the express server several times, the same problem comes again.

I guess [https://rt.services.visualstudio.com/QuickPulseService.svc/post?ikey=_ikey_](https://rt.services.visualstudio.com/QuickPulseService.svc/post?ikey=_ikey_) is closing the connection/dropping the request after a specific time.

This is the error I am getting:
```
ApplicationInsights:Unable to contact qps endpoint, dropping this Live Metrics packet [ { Error: read ECONNRESET
    at _errnoException (util.js:1024:11)
    at TLSWrap.onread (net.js:615:25) code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'read' } ]
```

markwolff:
Hey @KrishnaPravin I expect with this error that the SDK is able to leave the error state after some time. Are you seeing this as not the case after a few minutes? In the meantime I'll look into if the underlying can be resolved

OsvaldoRosado:
@markwolff If this error isn't critical, perhaps the message should be suppressed until it happens a couple of times in a row? (Unless verbose logging is enabled)

Like was done for the ingestion channel in https://github.com/Microsoft/ApplicationInsights-node.js/issues/180#issuecomment-394025939

markwolff:
@OsvaldoRosado agreed. I'm also considering not showing it unless it is actually POSTing data

KrishnaPravin:
> Hey @KrishnaPravin I expect with this error that the SDK is able to leave the error state after some time. Are you seeing this as not the case after a few minutes? In the meantime I'll look into if the underlying can be resolved

For first few minutes after connection reset error, I didn't see any data in the portal. But after few minutes, the error was still logging in my local console but in parallel some data was showing up in the live metrics portal. Another thing to note is, the number of online servers count in the live metrics portal was 0.

KrishnaPravin:
> @markwolff If this error isn't critical, perhaps the message should be suppressed until it happens a couple of times in a row? (Unless verbose logging is enabled)
> 
[...]

PR #511, opened 2019-04-24
only log extended duration QPS errors

markwolff:
- If verbose debugging is enabled, log QPS errors only when POSTing
- If verbose debugging is disabled, log QPS errors only if they persist for an extended amount of time
as per discussion in #510

### DIFF
--- a/Library/QuickPulseSender.ts
+++ b/Library/QuickPulseSender.ts
@@ -60,15 +60,10 @@ class QuickPulseSender {
             // Do nothing for now.
             this._consecutiveErrors++;
 
-            // Log every error, but instead warn when X number of consecutive errors occur
-            let shouldWarn = false;
-            let notice = `Live Metrics endpoint could not be reached ${this._consecutiveErrors} consecutive times. This packet will not appear in Live Metrics. Most recent error:`;
-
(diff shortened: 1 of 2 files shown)


Library/QuickPulseSender.ts
```diff
@@ -47,12 +49,20 @@ class QuickPulseSender {
...
             done(shouldPOSTData, res);
         });
         req.on("error", (error: Error) => {
             // Unable to contact qps endpoint.
             // Do nothing for now.
-            Logging.warn("Unable to contact qps endpoint, dropping this Live Metrics packet", error);
+            this._consecutiveErrors++;
+            if (postOrPing === "post") {
+                if (this._consecutiveErrors > 20) {
+                    Logging.warn("Unable to contact qps endpoint, dropping this Live Metrics packet", error);
```
OsvaldoRosado: You should follow the format used for ingestion and state how many consecutive errors have occurred. Also highly suggest using modulo instead of every request over > 20 failures for consistency.

OsvaldoRosado: Also, calls to `Logging` should get passed a tag to indicate where the failure came from

markwolff: Updated

Issue #414, opened 2018-08-11
Support W3C distributed tracing protocol


lmolkova:
[W3C distributed tracing protocol](https://github.com/w3c/distributed-tracing/tree/master/trace_context) is going to replace current `Request-Id` format.

The implementation should be split into several phases
1. Start emitting w3c-compatible operation ids, without supporting W3C headers.
**This phase should be implemented first and phase 2 could follow in several months.**

2. Opt-in support for W3C and Request-Id
Once we are sure p2 works well across different SDKs and don't break customers, we could switch to p3 

3. Supporting W3C enabled by default, in back-compat mode with Request-Id
4. Deprecating and retiring Request-Id (in several years)

# 1. W3C compatible ids
In order to enable customer transition story and backward compatibility, the first step is to start emitting  w3c-compatible ids:
32 [a-z][0-9] lowercase characters (i.e. lowercase guid without dashes or 16 random bytes encoded into hex string).
All zeros is considered invalid. 
Example of the valid id: `0af7651916cd43dd8448eb211c80319c`. 
The rest: header names and hierarchy in Request-Id could still be preserved.

If request-id was received from upstream, the SDK should continue to behave the same - propagate and do not attempt to change the format.

So, from node.js perspective, the only change here seems to be `Utils.newGuid` function

https://github.com/Microsoft/ApplicationInsights-node.js/blob/7588e9a106144adc228c4e44b188c3b898b726c6/Library/Util.ts#L78

# 2. Supporting W3C (opt-in) and Request-Id. 
there are several existing implementations that could be used as reference: .NET AI SDK, Python OpenCensus.
 
There should be a setting that enables W3C support.

When W3C is enabled:
   1. If incoming request contains `traceparent` only: parse it, set operation, parent and telemetry id accordingly. `traceparent` should be injected into outgoing requests. request-id should be injected in back-compat format `|traceId.spanId.` so that older SDKs could understand it.
   2. If incoming request contains only `Request-Id`, new traceId and spanId should be started, Request-Id value should be used as a parent. Root part of Request-Id should be stored in custom dimension on the request telemetry if root part is different from traceId. On the outgoing request side, Request-Id should be emitted in the `|traceId.spanId.` format
   3. if request contains both: request-id and traceparent, request-id should be ignored, so it becomes the same as p1
   4. if request does not contain any correlation headers, see case p2

PR #519, opened 2019-05-21
add w3c distributed tracing

markwolff:
Implements (off-by-default, P2 of #414) [W3C Distributed Tracing Protocol](https://github.com/w3c/trace-context). Back compat headers will be parsed and included if W3C Tracing mode is enabled.

Currently, this feature will be off by default, and requires `.setDistributedTracingMode(appInsights.DistributedTracingModes.AI_AND_W3C)` to be called during setup to enable. The reasoning behind this is to maintain alignment with other Application Insights SDKs (.NET, Java) which already support this feature. The intention is to enable this by default in a few months (P3 of #414).

Enabling W3C tracing will not break existing AI correlation. Instead, it will send back-compatibility AI headers in addition to the new traceparent/tracestate headers. The general Request/Dependency telemetry flow of handling old/new headers is as follows

| Incoming Header Types    | Request.operation.parentId | Request.id                          | Request.operation.id | Request.legacyRootId | Dependency.context.operation.id | Dependency.context.operation.parentId | Dependency.id                       | 
|-----------------------------|----------------------------|-------------------------------------|----------------------|----------------------|---------------------------------|---------------------------------------|-------------------------------------| 
| traceparent && request-id   | traceparent.backCompat     | traceparent.updateSpanId.backCompat | traceparent.traceId  |                      | Request.operation.id            | Request.id                            | traceparent.updateSpanId.backCompat | 
| traceparent && !request-id  | traceparent.backCompat     | traceparent.updateSpanId.backCompat | traceparent.traceId  |                      | Request.operation.id            | Request.id                            | traceparent.updateSpanId.backCompat | 
| !traceparent && request-id  | request-id                 | new traceparent().backCompat        | new w3c id           | request-id.rootId    | Request.operation.id            | Request.id                            | traceparent.updateSpanId.backCompat | 
| !traceparent && !request-id |                            | new traceparent().backCompat        | new w3c id           |                      | Request.operation.id            | Request.id                            | traceparent.updateSpanId.backCompat |


markwolff:
@SergeyKanzhelev @reyang @lmolkova

reyang:
Just a quick reference https://github.com/census-instrumentation/opencensus-node/blob/fa97a9b6f19b97e1038ffa9e1be4b407f3844df2/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts.

lmolkova:
LGTM overall, thanks! Let me know if I can play with it and test back-compat scenarios

Tests/EndToEnd.tests.ts
```diff
@@ -273,6 +275,95 @@ describe("EndToEnd", () => {
...
+            sandbox.stub(CorrelationContextManager, "getCurrentContext", () => ({
+                operation: {
+                    traceparent: new TraceParent("00-5e84aff3af474588a42dcbf3bd1db95f-1fc066fb77fa43a3-00"),
+                    tracestate: "sometracestate"
+                },
+                customProperties: {
+                    serializeToHeader: (): null => null
+                }
+            }));
+            https.request(<any>'https://httpbin.org/status/200', (c) => {
```
OsvaldoRosado: Are these calls actually made during unit tests? Or is request mocked out here?

AutoCollection/HttpRequestParser.ts
```diff
@@ -204,23 +247,27 @@ class HttpRequestParser extends RequestParser {
...
-                // For the scenarios that don't guarantee an AI-created context,
-                // override the requestId with the provided one.
-                this.requestId = requestId;
+                if (request.headers[RequestResponseHeaders.requestIdHeader]) {
+                    this.parentId = request.headers[RequestResponseHeaders.requestIdHeader];
+                    this.requestId = CorrelationIdManager.generateRequestId(this.parentId);
+                    this.correlationContextHeader = request.headers[RequestResponseHeaders.correlationContextHeader];
+                } else {
+                    // Legacy fallback
+                    const rootId = request.headers[RequestResponseHeaders.rootIdHeader];
```
OsvaldoRosado: Are these doubly-old headers still read in w3cEnabled mode?

AutoCollection/HttpRequestParser.ts
```diff
@@ -204,23 +247,27 @@ class HttpRequestParser extends RequestParser {
...
-                this.parentId = request.headers[RequestResponseHeaders.parentIdHeader];
-                this.requestId = CorrelationIdManager.generateRequestId(rootId || this.parentId);
-                this.correlationContextHeader = null;
-            }
-            if (requestId) {
-                // For the scenarios that don't guarantee an AI-created context,
-                // override the requestId with the provided one.
-                this.requestId = requestId;
+                if (request.headers[RequestResponseHeaders.requestIdHeader]) {
+                    this.parentId = request.headers[RequestResponseHeaders.requestIdHeader];
```
OsvaldoRosado: It's a bit strange to have this logic duplicated elsewhere in w3c mode - seems like a good way to have hard to find bugs. Any way we can unify some of this header parsing so headers of one kind only have one chunk of code that handles them?

AutoCollection/CorrelationContextManager.ts
```diff
@@ -7,6 +7,8 @@ import * as DiagChannel from "./diagnostic-channel/initialization";
 
 // Don't reference modules from these directly. Use only for types.
 import * as cls from "cls-hooked";
+import TraceParent = require("../Library/TraceParent");
+import TraceState = require("../Library/TraceState");
```
reyang: Most of the existing implementations use "Tracestate" versus "TraceState". Refer to https://github.com/census-instrumentation/opencensus-python/pull/238#discussion_r204805320.

AutoCollection/HttpRequestParser.ts
```diff
@@ -196,6 +213,32 @@ class HttpRequestParser extends RequestParser {
...
+
+
+        this.traceparent = new TraceParent(traceparentHeader, requestIdHeader);
+        this.tracestate = new TraceState(tracestateHeader);
+
+        // If !traceparent && tracestate, expectation is to discard tracestate and generate new traceparent
+        if (!traceparentHeader && tracestateHeader) {
+            this.tracestate = null;
+        }
+        this.operationId = this.traceparent.traceId;
```
reyang: I suggest that we separate/decouple W3C TraceContext and legacy AI operation compat, instead of putting compat logic inside W3C implementation.

Library/TraceParent.ts
```diff
@@ -0,0 +1,121 @@
...
+                // Version validation
+                if (!this.version.match(/^[0-9a-f]{2}$/g)) {
+                    this.version = TraceParent.DEFAULT_VERSION;
+                    this.traceId = Util.w3cTraceId();
+                }
+                if (this.version === "00" && len !== 4) { // 0x00 (and perhaps future versions) require exactly 4 fields. This strict check  will need to be updated on each spec release
+                    this.traceId = Util.w3cTraceId();
+                    this.spanId = Util.w3cTraceId().substr(0, 16);
+                }
+                if(this.version === "ff") { // 0xff is forbidden, generate new traceparent
```
reyang: minor: space after if, btw, do we have lint in CI?

Library/TraceParent.ts
```diff
@@ -0,0 +1,121 @@
...
+        } else if (parentId) { // backcompat constructor
+            // If incoming request contains only request-id, new traceid and spanid should be started, request-id value should be used as a parent. Root part of request-id should be stored in custom dimension on the request telemetry if root part is different from traceid. On the outgoing request side, request-id should be emitted in the |traceId.spanId. format.
+            this.parentId = parentId.slice(); // copy
+            let operationId = CorrelationIdManager.getRootId(parentId);
+            if (!TraceParent.isValidTraceId(operationId)) {
+                this.legacyRootId = operationId;
+                operationId = Util.w3cTraceId();
+            }
+            if (parentId.indexOf("|") !== -1) {
+                parentId = parentId.substring(1+parentId.substring(0, parentId.length-1).lastIndexOf("."), parentId.length-1);
```
reyang: minor, lint (spaces around plus sign)

reyang:
> Implements (off-by-default) W3C Distributed Tracing Protocol. Back compat headers will be parsed and included if W3C Tracing mode is enabled.

Would you briefly explain the design decision here?

lmolkova:
> Back compat headers will be parsed and included if W3C Tracing mode is enabled.
> Would you briefly explain the design decision here?

We need to keep correlation for existing customers and enable for them path to upgrade. Note that none of current AI SDKs does W3C by default (only functions do).
To correlate with legacy world, we need to have back-compat mode where we parse incoming headers and propagate legacy headers. Happy to chat more about it.

reyang:
> We need to keep correlation for existing customers and enable for them path to upgrade. Note that none of current AI SDKs does W3C by default (only functions do).

Yes, I understand this part.

> To correlate with legacy world, we need to have back-compat mode where we parse incoming headers and propagate legacy headers. Happy to chat more about it.

This is the part I want to have more clarification. There are many ways of providing compat:
1. Service A sends W3C to service B, while service B communicates with service C which only understands AI headers.
2. Service A sends AI header to service B, while service B communicates with service C which only understands W3C.
3. Service A communicates directly to both service B and C, while B only understands W3C and C only understands AI headers.

Tests/EndToEnd.tests.ts
markwolff: I believe these calls are mocked. Will switch to url that the rest of the tests are using to avoid confusion

AutoCollection/HttpRequestParser.ts
markwolff: They should still be read, but seem to not be read as it is now. I will re-arrange some stuff so that w3c mode is run in addition existing logic, instead of completely replacing it

markwolff:
@reyang I've updated this PR to add a clarification table on how this SDK will handle back-compatibility headers.

Library/Traceparent.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (traceparent) { // traceparent constructor
+            // If incoming request contains traceparent: parse it, set operation, parent and telemetry id accordingly. traceparent should be injected into outgoing requests. request-id should be injected in back-compat format |traceId.spanId. so that older SDKs could understand it.
+            if (traceparent.split(",").length > 1) { // If more than 1 traceparent is present, discard both
+                this.traceId = Util.w3cTraceId();
+                this.spanId = Util.w3cTraceId().substr(0, 16);
+            } else {
+                const traceparentArr = traceparent.split("-");
+                const len = traceparentArr.length;
+                if (len >= 4) { // traceparent must contain at least 4 fields
+                    this.version = traceparentArr[0].trim();
```
reyang: Curious, why do we trim here? to remove heading/trailing spaces?
If it is for getting rid of spaces, might make more sense to move them to header parsing logic versus here.

Library/Traceparent.ts
```diff
@@ -0,0 +1,121 @@
...
+    public spanId: string;
+    public traceFlag: string = Traceparent.DEFAULT_TRACE_FLAG;
+    public traceId: string;
+    public version: string = Traceparent.DEFAULT_VERSION;
+
+    constructor(traceparent?: string, parentId?: string) {
+        if (traceparent) { // traceparent constructor
+            // If incoming request contains traceparent: parse it, set operation, parent and telemetry id accordingly. traceparent should be injected into outgoing requests. request-id should be injected in back-compat format |traceId.spanId. so that older SDKs could understand it.
+            if (traceparent.split(",").length > 1) { // If more than 1 traceparent is present, discard both
+                this.traceId = Util.w3cTraceId();
```
reyang: I know this is added before this PR, the name is weird, especially when you look at `Util.w3cTraceId().substr(0, 16)`.

Library/Traceparent.ts
```diff
@@ -0,0 +1,121 @@
...
+                    this.spanId = Util.w3cTraceId().substr(0, 16);
+                }
+
+                // Version validation
+                if (!this.version.match(/^[0-9a-f]{2}$/g)) {
+                    this.version = Traceparent.DEFAULT_VERSION;
+                    this.traceId = Util.w3cTraceId();
+                }
+                if (this.version === "00" && len !== 4) { // 0x00 (and perhaps future versions) require exactly 4 fields. This strict check will need to be updated on each spec release
+                    this.traceId = Util.w3cTraceId();
```
reyang: It seems we don't need this assignment in so many places? The logic can be simplified.

AutoCollection/HttpRequestParser.ts
markwolff: Resolving this as I refactored it bit so that code looks cleaner. Each parsing scenario is now split into its own segment. The bit where the generated Traceparent saves back-compat info is shared between all w3c parsing segments.

AutoCollection/HttpRequestParser.ts
markwolff: Refactored to separate each parsing scenario

Library/Traceparent.ts
markwolff: You are right. I've removed these two trims and moved them to before the `split("-")`. I am unsure if I want to move this to the header parsing logic as it is a part of the tracecontext spec and I'd like to have as much of the logic for that in the same place as possible.

markwolff:
Considered enablement via enum string name as well. Seemed like too many ways to enable so I ripped it out. Forgot to remove it here as well

PR #538, opened 2019-07-19
Add ConnectionStringParser

markwolff:
Adds `ConnectionStringParser`, which parses both in-code and environment variable connection strings and combines them based on code > environment variable. Instrumentation key priority is `.setup(CS)` > `.setup(ikey)` > CS env var > ikey env var.
@Mikhail-msft

TODOS
- [x] Add CS parsing in `.setup(...)`


Declarations/Contracts/Constants.ts
```diff
@@ -18,4 +18,18 @@ export function domainSupportsProperties(domain: Domain): domain is ISupportProp
...
+ * allow for case-insensitivity across field names
+ * @interface IConnectionStringFields
+ */
+export interface IConnectionStringFields {
+    authorization?: string;
+    instrumentationkey?: string;
+    ingestionendpoint?: string;
+    liveendpoint?: string;
+
+    [key: string]: string; // catch all for non-recognized fields
```
markwolff: I don't like including this here. Ideally there is a better way to do case-insensitive parsing of the string without having to check which interface member each connection string member potentially belongs to

Declarations/Contracts/Constants.ts
```diff
@@ -18,4 +18,18 @@ export function domainSupportsProperties(domain: Domain): domain is ISupportProp
...
-}
\ No newline at end of file
+}
+
+/**
+ * Subset of Connection String fields which this SDK can parse. Lower-typecased to
+ * allow for case-insensitivity across field names
+ * @interface IConnectionStringFields
+ */
+export interface IConnectionStringFields {
```
OsvaldoRosado: Nit: This is more like a static type than an interface that someone is expected to conform to and extend in an OO way. Usually interfaces like these are defined in this SDK without the `I` prefix.

Library/ConnectionStringParser.ts
```diff
@@ -0,0 +1,23 @@
...
+
+class ConnectionStringParser {
+    private static _FIELDS_SEPARATOR = ";";
+    private static _FIELD_KEY_VALUE_SEPARATOR = "=";
+
+    public static parse(connectionString: string): IConnectionStringFields {
+        const kvPairs = connectionString.split(ConnectionStringParser._FIELDS_SEPARATOR);
+
+        return kvPairs.reduce((fields: IConnectionStringFields, kv: string) => {
+            const equalsIndex = kv.indexOf(ConnectionStringParser._FIELD_KEY_VALUE_SEPARATOR);
```
OsvaldoRosado: Why not use split?

Library/Config.ts
```diff
@@ -53,9 +58,21 @@ class Config {
     /** Host name for quickpulse service */
     private _quickPulseHost: string;
 
-    constructor(instrumentationKey?: string) {
-        this.instrumentationKey = instrumentationKey || Config._getInstrumentationKey();
-        this.endpointUrl = `${this.endpointBase}/v2/track`;
+
+    constructor(instrumentationKey?: string, connectionStringCode?: string) {
```
OsvaldoRosado: How does `connectionStringCode` get populated? I don't see a change to the `setup` method to pass it.

How will `setup` know which field to fill?

Library/ConnectionStringParser.ts
markwolff: I don't think the values are prohibited from using `=`, so it is more clear to slice before&after parts at the first `=` instead of doing a join on `parts[1...n]`

Library/Config.ts
markwolff: This constructor is incorrect right now. I am going to combine functionalities into the 1st arg. `setup` will behave the same

Library/ConnectionStringParser.ts
MS-TimothyMothra: Comments in the design spec state that '=' and ';' should never appear in the values :)

Library/Config.ts
```diff
@@ -13,6 +15,9 @@ class Config {
     public static ENV_profileQueryEndpoint = "APPINSIGHTS_PROFILE_QUERY_ENDPOINT";
     public static ENV_quickPulseHost = "APPINSIGHTS_QUICKPULSE_HOST";
 
+    // Azure Connection String
+    public static ENV_connectionString = "APPINSIGHTS_CONNECTIONSTRING";
```
MS-TimothyMothra: This is the wrong environment variable. Please use "APPLICATIONINSIGHTS_CONNECTION_STRING"

applicationinsights.ts
```diff
@@ -69,15 +69,15 @@ let _performanceLiveMetrics: AutoCollectPerformance;
  * Initializes the default client. Should be called after setting
  * configuration options.
  *
- * @param instrumentationKey the instrumentation key to use. Optional, if
+ * @param setupString the Connection String or Instrumentation Key to use. Optional, if
  * this is not specified, the value will be read from the environment
- * variable APPINSIGHTS_INSTRUMENTATIONKEY.
+ * variable APPINSIGHTS_CONNECTIONSTRING or APPINSIGHTS_INSTRUMENTATIONKEY.
```
MS-TimothyMothra: This is the wrong environment variable. Please use "APPLICATIONINSIGHTS_CONNECTION_STRING"

Library/ConnectionStringParser.ts
markwolff: Updated to use split

Library/Config.ts
markwolff: Resolved in 91297e5

applicationinsights.ts
markwolff: Resolved in 91297e5

Declarations/Contracts/Constants.ts
markwolff: Renamed to `ConnectionStringFields`

Library/ConnectionStringParser.ts
```diff
@@ -0,0 +1,45 @@
...
+            }
+            return fields;
+        }, {});
+
+        if (Object.keys(result).length > 0) {
+            // this is a valid connection string, so parse the results
+
+            if (result.endpointsuffix) {
+                // use endpoint suffix where overrides are not provided
+                const locationPrefix = result.location ? result.location + "." : "";
```
lzchen: Can't the location prefix be used even if an endpoints suffix isn't provided?

Library/ConnectionStringParser.ts
```diff
@@ -0,0 +1,45 @@
...
+                result.ingestionendpoint = result.ingestionendpoint || ("https://" + locationPrefix + "dc." + result.endpointsuffix);
+                result.liveendpoint = result.liveendpoint || ("https://" + locationPrefix + "live." + result.endpointsuffix);
+            }
+
+            // apply the default endpoints
+            result.ingestionendpoint = result.ingestionendpoint || Constants.DEFAULT_BREEZE_ENDPOINT;
+            result.liveendpoint = result.liveendpoint || Constants.DEFAULT_LIVEMETRICS_ENDPOINT;
+        }
+
```
lzchen: Should we check for valid "Authorization" key?

Library/ConnectionStringParser.ts
markwolff: From what I could tell from the spec and published unit tests, it only works with endpoint suffix. @TimothyMothra to confirm

Library/ConnectionStringParser.ts
markwolff: Since there is no information on how an SDK should parse a non-ikey `Authorization` field, I decided not to support it. Also, `InstrumentationKey=a-b-c-d-e` is listed as a valid connection string since it defaults to ikey anyways.

Library/ConnectionStringParser.ts
TimothyMothra: Location will only apply to the EndpointSuffix. 

1. check for explicit endpoint (location is ignored)
2. check for endpoint suffix (location is optional)
3. use classic endpoint (location is ignored)

lzchen: Location can be pre-pended even if endpoints suffix is not provided.

markwolff: Can you clarify what you mean?

TimothyMothra: Please do not delete my comments.

We have an email thread where we've defined how we intend to handle Locations. Locations are to only apply to the Endpoint Suffix. 
If you have an issue with this please contact me offline.

Library/Config.ts
```diff
@@ -13,6 +15,9 @@ class Config {
     public static ENV_profileQueryEndpoint = "APPINSIGHTS_PROFILE_QUERY_ENDPOINT";
     public static ENV_quickPulseHost = "APPINSIGHTS_QUICKPULSE_HOST";
 
+    // Azure Connection String
+    public static ENV_connectionString = "APPINSIGHTS_CONNECTION_STRING";
```
TimothyMothra: This is the wrong environment variable. Please use: "APPLICATIONINSIGHTS_CONNECTION_STRING"

markwolff: updated to `APPLICATIONINSIGHTS_CONNECTION_STRING`

Library/TelemetryClient.ts
```diff
@@ -29,10 +29,10 @@ class TelemetryClient {
 
     /**
      * Constructs a new client of the client
-     * @param iKey the instrumentation key to use (read from environment variable if not specified)
+     * @param setupString the Connection String or Instrumentation Key to use (read from environment variable if not specified)
      */
-    constructor(iKey?: string) {
-        var config = new Config(iKey);
+    constructor(setupString?: string) {
```
TimothyMothra: Do you have unit tests to make sure this still initializes correctly with either an Ikey or ConnString? Maybe I missed it but i didn't see any

Tests/Library/ConnectionStringParser.tests.ts
```diff
@@ -0,0 +1,104 @@
...
+            const authorization = "ikey"
+            const instrumentationKey = "ikey";
+            const ingestionEndpoint = "ingest";
+            const liveEndpoint = "live";
+            const connectionString = `Autho.rization=${authorization};Instrume.ntationKey=${instrumentationKey};Ingestion.Endpoint=${ingestionEndpoint};LiveEnd.point=${liveEndpoint}`;
+
+            const result = ConnectionStringParser.parse(connectionString);
+
+            assert.deepEqual(result.authorization, undefined);
+            assert.deepEqual(result.instrumentationkey, undefined);
```
TimothyMothra: If an Ikey is not found, I would expect this to throw some kind of exception.

TimothyMothra:
Hey @markwolff,
It looks like you covered all the correct use cases in your tests. :)
I don't see many tests for invalid cases; Incorrect delimiters, duplicate keys, spaces in key.
You may want to consider adding some tests to verify that you can detect invalid connection strings without crashing and can provide meaningful error messages so customers can self diagnose.

Otherwise, this LGTM.

Library/TelemetryClient.ts
markwolff: Yep these are in `Config.ts` tests.

Tests/Library/ConnectionStringParser.tests.ts
markwolff: Currently I'm relying on [this](https://github.com/microsoft/ApplicationInsights-node.js/blob/88b38caffaad1310ceb8242c5cae9bc40bf1e573/Library/Config.ts#L125) to show errors to users after all parsing is finished. I'm unsure on if I should show parsing errors here or where the results are actually used (or both?). For the most part errors are shown where the stuff is used.

markwolff:
> Hey @markwolff,
> It looks like you covered all the correct use cases in your tests. :)
> I don't see many tests for invalid cases; Incorrect delimiters, duplicate keys, spaces in key.
> You may want to consider adding some tests to verify that you can detect invalid connection strings without crashing and can provide meaningful error messages so customers can self diagnose.
> 
> Otherwise, this LGTM.

Synced with Mothra offline. Will merge PR this in now and bring validation testing&logging to .NET parity in a separate PR.

Declarations/Contracts/Constants.ts
```diff
@@ -18,4 +18,23 @@ export function domainSupportsProperties(domain: Domain): domain is ISupportProp
...
-}
\ No newline at end of file
+}
+
+/**
+ * Subset of Connection String fields which this SDK can parse. Lower-typecased to
+ * allow for case-insensitivity across field names
+ * @type ConnectionStringKey
+ */
+export interface ConnectionString {
```
hectorhdzg: IConnectionString, we are already using that format for interface names here

Declarations/Contracts/Constants.ts
```diff
@@ -18,4 +18,23 @@ export function domainSupportsProperties(domain: Domain): domain is ISupportProp
...
+    authorization?: string;
+    instrumentationkey?: string;
```
hectorhdzg: Why not camel case?

hectorhdzg: I guess is because of this line

 const key = kvParts[0].toLowerCase() as ConnectionStringKey;

PR #551, opened 2019-09-20
change(winston/console): send exception if error provided to logger

markwolff:
For scenarios where `winston.error` is used to log an `Error`, we will now call `trackException` instead of sending log data with `trackTrace`. A similar change is made to all other supported loggers (`console` and `bunyan`) An example:
```js
try {
  ...
} catch (e) {
  winston.error(e); // this will now send 'e' as exception telemetry instead of log telemetry
}
```
This PR also fixes the unit test reliability of `Performance.tests.ts`

### DIFF
--- a/AutoCollection/diagnostic-channel/bunyan.sub.ts
+++ b/AutoCollection/diagnostic-channel/bunyan.sub.ts
@@ -23,7 +23,7 @@ const subscriber = (event: IStandardEvent<bunyan.IBunyanData>) => {
     const message = event.data.result as Error | string;
     clients.forEach((client) => {
         const AIlevel = bunyanToAILevelMap[event.data.level];
-        if ((message) instanceof Error) {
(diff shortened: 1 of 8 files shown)


hectorhdzg:
Why this change is being introduced?, logger and actual errors are usually handled in different places in Applications, logger.error doesn't necessary mean there was a code Exception

markwolff:
> Why this change is being introduced?, logger and actual errors are usually handled in different places in Applications, logger.error doesn't necessary mean there was a code Exception

Consistency with other language SDKs. For example, with .NET SDK, calling ILogger `logger.Log(msg, exception)` would send exception telemetry. The logs currently sent with SeverityLevel.error are not included in the App Insights curated "Failures" experiences, since no callstack is included. Also, for those who rely 100% on a single logger (e.g. winston), there is no trivial way to send exception telemetry with this SDK.

PR #552, opened 2019-09-24
feat(diag): add mongo enhanced query collection

markwolff:
Adds experimental (opt-in) configuration option on `Config` to opt into sending more detailed query information for autocollected dependencies. Currently only mongodb queries will work. Support for other libraries will be added in separate PRs (and future releases).

This feature is intended to be experimental in that the format of the data or the actual data could change in future releases. Other libraries will also be supported, which would cause more telemetry volume to be sent if this feature is enabled.

### DIFF
--- a/AutoCollection/diagnostic-channel/mongodb.sub.ts
+++ b/AutoCollection/diagnostic-channel/mongodb.sub.ts
@@ -8,12 +8,17 @@ import { mongodb } from "diagnostic-channel-publishers";
 let clients: TelemetryClient[] = [];
 
 export const subscriber = (event: IStandardEvent<mongodb.IMongoData>) => {
+    const dbName = (event.data.startedData && event.data.startedData.databaseName) || "Unknown database";
     clients.forEach((client) => {
-        const dbName = (event.data.startedData && event.data.startedData.databaseName) || "Unknown database";
+        const data = client.config.enhancedDependencyCollection
+            && event.data.startedData.command
+            && event.data.startedData.command[event.data.event.commandName]
+            ? JSON.stringify(event.data.startedData.command[event.data.event.commandName])
+            : event.data.event.commandName;
         client.trackDependency(
             {
                 target: dbName,
-                data: event.data.event.commandName,
+                data: data,
                 name: event.data.event.commandName,
                 duration: event.data.event.duration,
                 success: event.data.succeeded,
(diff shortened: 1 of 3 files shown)


Library/Config.ts
```diff
@@ -46,6 +46,11 @@ class Config {
     public httpAgent: http.Agent;
     /** An https.Agent to use for SDK HTTPS traffic (Optional, Default undefined) */
     public httpsAgent: https.Agent;
+    /** (Off by default) If true, when depdency telemetry is autocollected, more information about it
```
hectorhdzg: Specify this only works with MongoDB, we can update later if MySQL and PostgreSQL support is added

AutoCollection/diagnostic-channel/mongodb.sub.ts
```diff
@@ -36,4 +41,9 @@ export function enable(enabled: boolean, client: TelemetryClient) {
             channel.unsubscribe("mongodb", subscriber);
         }
     }
-}
\ No newline at end of file
+}
+
+export function dispose() {
```
hectorhdzg: Where this will be called?

AutoCollection/diagnostic-channel/mongodb.sub.ts
```diff
@@ -8,12 +8,17 @@ import { mongodb } from "diagnostic-channel-publishers";
 let clients: TelemetryClient[] = [];
 
 export const subscriber = (event: IStandardEvent<mongodb.IMongoData>) => {
+    const dbName = (event.data.startedData && event.data.startedData.databaseName) || "Unknown database";
     clients.forEach((client) => {
-        const dbName = (event.data.startedData && event.data.startedData.databaseName) || "Unknown database";
+        const data = client.config.enhancedDependencyCollection
+            && event.data.startedData.command
+            && event.data.startedData.command[event.data.event.commandName]
+            ? JSON.stringify(event.data.startedData.command[event.data.event.commandName])
```
hectorhdzg: JSON.stringify could throw an exception adding try/catch here would be good

AutoCollection/diagnostic-channel/mongodb.sub.ts
markwolff: It's currently a convenience function for tests only since `clients` is not exported. Perhaps instead I should just export `clients` and implement `dispose()` in tests?

abondoa:
Now that this is rejected, is there an alternative way to get more detailed logs from Mongo DB in Application Insights? In particular I am looking for getting the name of the collection being queried/operated on.

PR #554, opened 2019-10-09
feat: Allow to specify severityLevel when tracking Exception

Socolin:


Library/EnvelopeFactory.ts
```diff
@@ -152,6 +152,11 @@ class EnvelopeFactory {
         exception.severityLevel = Contracts.SeverityLevel.Error;
         exception.measurements = telemetry.measurements;
         exception.exceptions = [];
+        if (!isNaN(telemetry.severity)) {
+            exception.severityLevel = telemetry.severity;
+        } else {
+            exception.severityLevel = Contracts.SeverityLevel.Error;
+        }
```
markwolff: Can this be moved to where `exception.severityLevel` is initially set?

Socolin: I missed that line :p I'm fixing it

Socolin:
I've updated my commit

PR #562, opened 2019-10-30
feat: add Azure SDK Patching Subscriber

markwolff:
- Adds subscriber for new azure SDK tracer patching
- Changes functional tests to use W3C header propagation
- Add AzureSDK functional test
- Support nested propagation testing in functional test suite


AutoCollection/AsyncHooksScopeManager.ts
```diff
@@ -0,0 +1,62 @@
+import { CorrelationContextManager, CorrelationContext } from "./CorrelationContextManager"
+import { ISpanContext } from "diagnostic-channel";
+import { EventEmitter } from "events";
+
+export interface Span {
```
markwolff: A clone of Span interface so that we don't need to take a Typescript 3 dependency

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -0,0 +1,72 @@
...
+                    properties: properties,
+                } as Contracts.RequestTelemetry & Contracts.Identified);
+            } else {
+                // Client
+                client.trackDependency({
+                    id: id,
+                    name: span.name,
+                    duration: duration,
+                    data: span.attributes.component || span.name,
+                    dependencyTypeName: span.attributes.component || span.name,
```
markwolff: The span attributes aren't guaranteed to provide meaningful information in a standardized way right now, so for now just populate them with current "standards" and iterate back on this in the future

Tests/FunctionalTests/Runner/Main.js
```diff
@@ -12,6 +12,12 @@ let startTime = null;
 
 let perfMode = process.argv.indexOf("-perfmode") !== -1;
 
+const skipAsyncHooksTests = () => {
```
markwolff: Skip "azure sdk" tests on versions of node where azure SDK won't work (must be node 8+, but not 8.0.0)

Tests/FunctionalTests/Runner/TestValidation.js
```diff
@@ -36,11 +37,16 @@ module.exports.TestValidation = class TestValidation {
                         return false;
                     }
 
-                    correlationId = requestTelemetry.tags["ai.operation.id"];
+                    correlationId = requestTelemetry.data.baseData.id;
```
markwolff: Functional tests are using W3C propagation now, so ids & parent ids will always have the format `|<traceId>.<spanId>.`. Previously, this stuff depended on the child's parent ids being `|<traceId>.` which can no longer happen. A spanId is always included!

Tests/FunctionalTests/TestApp/Config.js
```diff
@@ -16,6 +18,7 @@ var Config = {
     UseAutoConsole: true,
     UseAutoConsoleLog: true,
     UseDiskCaching: false,
+    DistributedTracingMode: appInsights.DistributedTracingModes.AI_AND_W3C,
```
markwolff: Functional tests will use W3C propagation

Tests/FunctionalTests/TestApp/Tasks/AzureSDK.js
```diff
@@ -0,0 +1,31 @@
+try {
+    var storage = require("@azure/storage-blob");
```
markwolff: Wrap in try/catch so that functionals can "try" to import this file for node4 and node6

Tests/FunctionalTests/TestApp/Tasks/Winston.js
```diff
@@ -3,14 +3,14 @@ winston.level = 'silly';
...
-        winston.log(type, "test %s", type);
+        winston.log(type, `test ${type}`);
         callback();
     }
 }
 
 function logFn2(type) {
     return (callback) => {
-        winston[type]("test %s", type);
+        winston[type](`test ${type}`);
```
markwolff: winston3 does not support string formatters like winston2 does, but winston3 does not support <node8. Sticking with winston2 for now, so this change probably isn't necessary.

AutoCollection/CorrelationContextManager.ts
```diff
@@ -1,6 +1,4 @@
-import http = require("http");
```
hectorhdzg: Not used?

AutoCollection/HttpDependencies.ts
```diff
@@ -63,6 +64,11 @@ class AutoCollectHttpDependencies {
             var shouldCollect = !(<any>options)[AutoCollectHttpDependencies.disableCollectionRequestOption] &&
                 !(<any>request)[AutoCollectHttpDependencies.alreadyAutoCollectedFlag];
 
+            // If someone else patched traceparent headers onto this request
+            if (options.headers && options.headers['user-agent'] && options.headers['user-agent'].toString().indexOf('azsdk-js') !== -1) {
```
hectorhdzg: Can you explain what is going on here?, what is azsdk-js is and why ignoring it?

AutoCollection/CorrelationContextManager.ts
markwolff: Yes it seems like it wasn't being used so I cleaned it up.

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const properties = {
+        ...span.attributes,
+    };
+    if (span.status.message) {
+        properties["status.message"] = span.status.message;
+    }
+    AsyncScopeManager.with(span, () => {
+        clients.forEach((client) => {
+            if (span.kind === 1) {
```
hectorhdzg: Enum for SpanKind would be good here

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -0,0 +1,72 @@
...
+                    id: id,
+                    name: span.name,
+                    duration: duration,
+                    resultCode: span.status.code,
+                    success: span.status.code === 0,
+                    url: span.attributes.component,
+                    properties: properties,
+                } as Contracts.RequestTelemetry & Contracts.Identified);
+            } else {
+                // Client
```
hectorhdzg: Not client only
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#spankind

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -0,0 +1,72 @@
...
+                    resultCode: span.status.code,
+                    success: span.status.code === 0,
+                    properties: properties
+                } as Contracts.DependencyTelemetry & Contracts.Identified);
+            }
+        });
+    });
+};
+
+export function enable(enabled: boolean, client: TelemetryClient) {
```
hectorhdzg: client is not used, is this a typo with clients?

AutoCollection/HttpDependencies.ts
markwolff: I should update the comment here because its not 100% clear.

Azure SDK patches outgoing http calls made via their SDK (e.g. GET a storage account blob). We patch ALL outgoing http calls, which intersects with azure SDK patching, so we need some way to prevent this double telemetry/patching. Another possibility, which I think Java SDK does, is to check for presence of `traceparent` header in the outgoing http call. The user will never set this header themselves, it is only set by APM SDKs. However, this would break patching if another SDK (e.g. opentelemetry) is used. Once this SDK supports opentelemetry patching, then this could be switched to check for traceparent header.  Until then, azure sdk sets this user-agent field on all of their outgoing http headers, so I think its okay to just check for this for now.

Summary:
1. Azure SDK patches their outgoing http calls to add `traceparent` and `user-agent`
2. This SDK can see the headers of an outgoing http call before it gets sent out, so we can not track the outgoing http call if Azure SDK tracer will already track it.
3. If I just look for `traceparent` and user instruments with multiple W3C compatible SDKs, then our SDK won't track, even though it should. Should only step-back for Azure SDK

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
markwolff: I should add a comment here. I saw some "Client" spans created by Azure SDK being labelled as "INTERNAL", e.g. `SpanKind.INTERNAL === 0`. Although I wonder if I should check for consumer vs producer here. Would consumer = server and producer = client?

hectorhdzg: Yes so in this case you send dependencies for Producer kind and that would be a problem right

markwolff: Agreed. Fixed in c6fb329

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const properties = {
+        ...span.attributes,
+    };
+    if (span.status.message) {
+        properties["status.message"] = span.status.message;
+    }
+    AsyncScopeManager.with(span, () => {
+        clients.forEach((client) => {
+            if (span.kind === SpanKind.SERVER || span.kind === SpanKind.CONSUMER) {
```
hectorhdzg: PRODUCER and SERVER

Issue #568, opened 2019-11-19
Typo in readme.md, develop branch


keithfree:
https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/README.md

There is a small typo in a method name in a code snippet within the configuration section. Method setDistributedTracingMode is written incorrectly as setDistributedTracing**Tracing**Mode.

salmanmkc:
Fixed, thanks.

markwolff:
Fixed by #567

PR #569, opened 2019-11-19
setDistributedTracingMode should be the method name

salmanmkc:
Fixing issue https://github.com/microsoft/ApplicationInsights-node.js/issues/568


markwolff:
Duplicate of #567 so I think this can be closed

PR #570, opened 2019-11-22
Include port name in "target" for dependency telemetry

xiao-lix:


AutoCollection/HttpDependencyParser.ts
```diff
@@ -64,6 +64,10 @@ class HttpDependencyParser extends RequestParser {
             remoteDependencyType = Contracts.RemoteDependencyDataConstants.TYPE_HTTP;
         }
 
+        if (urlObject.port) {
+            remoteDependencyTarget = remoteDependencyTarget + ":" + urlObject.port;
```
markwolff: nit: use `+=`

markwolff: Also what is the type of `urlObject.port`? If it is a number, would `0` be something we want to append or ignore?

PR #575, opened 2019-12-10
feat: add In-Proc-Agent

markwolff:
- Adds in-proc-agent for node.js SDK which can be directly imported by compute platforms.
  - `Default.ts`: Base IPA which other compute speciffic platforms will start from
  - `Oryx.ts`: Sets `alr_` prefix and sets up app insights via `Default.ts`

### DIFF
--- a/Bootstrap/Default.ts
+++ b/Bootstrap/Default.ts
@@ -55,7 +55,6 @@ export function setUsagePrefix(prefix: string) {
  * @param setupString connection string or instrumentation key
  */
 export function setupAndStart(setupString = _setupString): typeof types | null {
-    console.log("Trying to setup Application Insights", setupString, _extensionEnabled);
     if (!_extensionEnabled && sdkAlreadyExists() === false) {
         return null;
     }
--- a/Bootstrap/Default.ts
+++ b/Bootstrap/Default.ts
@@ -14,6 +14,26 @@ export interface AgentLogger {
     error(message?: any, ...optional: any[]): void;
 }
 
+function sdkAlreadyExists(): boolean {
+    try {
+        // appInstance should either resolve to user SDK or crash. If it resolves to attach SDK, user probably modified their NODE_PATH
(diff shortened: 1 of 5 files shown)


Bootstrap/Default.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Try to setup and start this app insights instance, if attach is enabled.
+ * @param setupString connection string or instrumentation key
+ */
+export function setupAndStart(setupString = _setupString): typeof types | null {
+    if (!_extensionEnabled) {
+        return null;
+    }
+    if (!setupString) {
+        _logger.error(
+            "Application Insights wanted to be started, but no Connection String was provided",
```
xiao-lix: I'm wondering if this error message is not accurate, should it mention instrumentation key is also not provided? WDYT?

Tests/Bootstrap/Default.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    it("should not setup and start the SDK if no setupString is provided", () => {
+        // Setup
+        const logger = new LoggerSpy();
+        const origEnv = process.env.APPLICATIONINSIGHTS_EXTENSION_VERSION;
+        process.env.APPLICATIONINSIGHTS_EXTENSION_VERSION = "~2";
+
+        // Test
+        const Default = require("../../Bootstrap/Default") as typeof types;
+        Default.setLogger(logger);
+        const instance = Default.setupAndStart();
```
xiao-lix: Just curious, why is _setupstring not used as default param here? Or is the origEnv disabled?

markwolff:
Also /cc @arroyc

Bootstrap/Default.ts
markwolff: Good point. Adjusted this error log

Tests/Bootstrap/Default.spec.ts
markwolff: I want to make sure it will fail gracefully when no param or env var is provided. In this case, origEnv is deleted beforehand.

PR #582, opened 2020-01-21
fix: cut a name for trackDependency if a sql is too big

sumbad:
I have found that if a SQL to Postgres DB is big we won't be able to see this dependency in Application Insights. I guess this happens because a name property that sends to the trackDependency function must be limited to 1024 symbols.

In my tests, I had limited a name property then info about Postgres dependency was saved to Application Insights.

### DIFF
--- a/AutoCollection/diagnostic-channel/postgres.sub.ts
+++ b/AutoCollection/diagnostic-channel/postgres.sub.ts
@@ -13,10 +13,11 @@ export const subscriber = (event: IStandardEvent<pg.IPostgresData>) => {
         const sql = (q.preparable && q.preparable.text) || q.plan || q.text || "unknown query";
         const success = !event.data.error;
         const conn = `${event.data.database.host}:${event.data.database.port}`;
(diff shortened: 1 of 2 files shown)


AutoCollection/diagnostic-channel/postgres.sub.ts
```diff
@@ -13,10 +13,11 @@ export const subscriber = (event: IStandardEvent<pg.IPostgresData>) => {
         const sql = (q.preparable && q.preparable.text) || q.plan || q.text || "unknown query";
         const success = !event.data.error;
         const conn = `${event.data.database.host}:${event.data.database.port}`;
+        const name = sql.length > 1024 ? sql.slice(0, 1021) + '...' : sql;
```
markwolff: This can be moved to a more generic location here:
https://github.com/microsoft/ApplicationInsights-node.js/blob/34a23fc9c869ecdd3b800d7748cf5ee81db6411f/Library/EnvelopeFactory.ts#L116-L118

sumbad: Thanks, I have moved it.

PR #588, opened 2020-01-28
fix: skip rethrow when other handlers exist

### LINKED ISSUE #584: [BUG] AutoCollectExceptions.enable re-throws an uncaught exception into the ether which can crash the process with status 7
**Description/Screenshot**
The `redis` package `retry_strategy` handler appears to have a problem with this feature of the `applicationinsights` package, such that a retry never happens due to this extra throw.

Screen Shot 2020-01-20 at 1 56 42 PM

The screenshot above has an event registered for `uncaughtException` but it doesn't execute due to the process exiting with status 7; the trigger of the exception is node.js > net.js > client which is utilized for a `redis` client.

The screenshot below is proper behavior as coded in my app:

Screen Shot 2020-01-20 at 2 45 57 PM

The `uncaughtException` event handler in my code would do some ops, then `process.exit(1)` to trigger a restart with a revised configuration; because this never runs I had processes restart with the same config in production:

Screen Shot 2020-01-20 at 1 55 41 PM

**Steps to Reproduce**
1. (with app insights active) using latest redis package connect to a server with a retry strategy defined for the client
2. stop redis, or disconnect the client uncleanly

 - OS/Browser:
 - SDK Version [e.g. 22]:
 - How you initialized the SDK:

**Expected behavior**
Application error is logged, application continues to run (no exit 7 status)

**Additional context**
Application is utilizing `tenso` for API server functionality, but that does not handle `uncaughtException` events.

markwolff:
Fixes #584 
Handle-time check to suppress rethrows when other handles exist. Alternative would be to expose 2nd `boolean | Function` arg to `setAutoCollectionExceptions(true, rethrowHandler)`. 

Node<13.7 Behavior
- If multiple uncaught handlers exist: no crash
- If this SDK is the only handler: crash (status=1)

Node13.7+ behavior
- For node >=13.7, use `uncaughtExceptionMonitor` instead. Process will only crash (default status code) if no `uncaughtException` handlers exist.

/cc @avoidwork - Please let me know your reasoning if you prefer explicitly providing an exception handler!

### DIFF
--- a/AutoCollection/Exceptions.ts
+++ b/AutoCollection/Exceptions.ts
@@ -38,16 +38,17 @@ class AutoCollectExceptions {
             if (!this._exceptionListenerHandle) {
                 // For scenarios like Promise.reject(), an error won't be passed to the handle. Create a placeholder
                 // error for these scenarios.
-                var handle = (reThrow: boolean, error: Error = new Error(AutoCollectExceptions._FALLBACK_ERROR_MESSAGE)) => {
+                var handle = (reThrow: boolean, name: string, error: Error = new Error(AutoCollectExceptions._FALLBACK_ERROR_MESSAGE)) => {
                     this._client.trackException({ exception: error });
                     this._client.flush({ isAppCrashing: true });
-                    if (reThrow) {
+                    // only rethrow when we are the only listener
+                    if (reThrow && process.listeners(name).length === 1) {
                         var THIS_IS_APPLICATION_INSIGHTS_RETHROWING_YOUR_EXCEPTION = error;
                         throw THIS_IS_APPLICATION_INSIGHTS_RETHROWING_YOUR_EXCEPTION; // Error originated somewhere else in your app
                     }
                 };
-                this._exceptionListenerHandle = handle.bind(this, true);
-                this._rejectionListenerHandle = handle.bind(this, false);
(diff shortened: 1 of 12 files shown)


AutoCollection/Exceptions.ts
```diff
@@ -38,16 +38,17 @@ class AutoCollectExceptions {
...
                 // error for these scenarios.
-                var handle = (reThrow: boolean, error: Error = new Error(AutoCollectExceptions._FALLBACK_ERROR_MESSAGE)) => {
+                var handle = (reThrow: boolean, name: string, error: Error = new Error(AutoCollectExceptions._FALLBACK_ERROR_MESSAGE)) => {
                     this._client.trackException({ exception: error });
                     this._client.flush({ isAppCrashing: true });
-                    if (reThrow) {
+                    // only rethrow when we are the only listener
+                    if (reThrow && process.listeners(name).length === 1) {
                         var THIS_IS_APPLICATION_INSIGHTS_RETHROWING_YOUR_EXCEPTION = error;
                         throw THIS_IS_APPLICATION_INSIGHTS_RETHROWING_YOUR_EXCEPTION; // Error originated somewhere else in your app
```
OsvaldoRosado: Rethrowing here is broken behavior. See the related thread for why

markwolff: Updated to use `process.exit(1)`

markwolff:
https://travis-ci.org/microsoft/ApplicationInsights-node.js/builds/643100644

PR #596, opened 2020-02-21
feat: add attach diagnostic logger

markwolff:
- Adds `FileWriter`, matches interface of `console` and ETW logger (aka `AgentLogger` interface)
  -  drop-in replacement for `console.log`/`console.error`, but writes to file instead
- Refactor: Each logger type (`StatusLogger` vs `DiagnosticLogger`) takes an `AgentLogger` to act as its output stream
  - `console` --> output to console
  - `ETW` --> output to ETW
  - `FileWriter` --> output to file


Bootstrap/Default.ts
```diff
@@ -82,12 +80,18 @@ export function setupAndStart(setupString = _setupString): typeof types | null {
 
     try {
         _appInsights = require("../applicationinsights");
+        if (_appInsights.defaultClient) {
```
hectorhdzg: Was this scenario causing issues?

markwolff: Not currently but if this function is called twice for whatever reason, the sdkVersion prefix telemetry processor would be run twice

PR #601, opened 2020-03-09
Initialize config before set profileQueryEndpoint

qetza:
The set on `profileQueryEndpoint` is using this current `Config` instance which is not fully configured. This fixes it by moving the set after all properties are initialized.

### DIFF
--- a/Library/Config.ts
+++ b/Library/Config.ts
@@ -83,11 +83,11 @@ class Config {
 
         this.setCorrelationId = (correlationId) => this.correlationId = correlationId;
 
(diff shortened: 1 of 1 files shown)


markwolff:
Was this causing issues for you? Which field is not configured here?

qetza:
When setting `profileQueryEndpoint` it will internally call `CorrelationIdManager.queryCorrelationId` passing the current instance of `Config` (where the proxy properties are not set yet) which will queue a call to `fetchAppId` which uses those properties to make an HTTP call.
Due to the nature of javascript the `fetchAppId` should not be dequeued before the `Config` instance is fully initialized but just in case (and for people like me note familiar at start with javascript "threading" model) i think it's better when reading the code have a fully set object passed to external objects.

PR #613, opened 2020-04-09
only create qps client when live metrics is enabled

### LINKED ISSUE #612: Possible memory leak
I've noticed a suspicious memory leak issue as of late. I've been using app insights to investigate but I'm starting to think that app insights is the cause of the memory leak, which I find somewhat ironic.

image

To the best of my ability, `_documents` seem to grow indefinitely. Can anyone help explain why this would be the case. All I can see is that my heap is growing for each snapshot and that the `_documents` array grows with it.

I did some further digging, and it looks like this leak happens because live metrics is not enabled? I'm I right in assuming that the `QuickPulseStateManager` manage live metrics? If live metrics is not enabled why does it retain so much data?

markwolff:
fixes #612

### DIFF
--- a/Tests/applicationInsights.tests.ts
+++ b/Tests/applicationInsights.tests.ts
@@ -85,6 +85,11 @@ describe("ApplicationInsights", () => {
             AppInsights.setup("key").start();
             assert.equal(AppInsights.liveMetricsClient, undefined, "live metrics client is not defined");
         });
+
+        it("should not start live metrics", () => {
+            AppInsights.setup("key").setSendLiveMetrics(false).start();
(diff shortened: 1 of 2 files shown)


leidegre:
I would like to add that I think it it important that you include that this change actually fixes an actual memory leak.

This leak took down our production environment and caused an outage for about 2 hours. I spent the last week trying to root cause the issue and all that I did to deserve this was that I used the default config as per the examples given for the package.

Now that it has been fixed, it is less likely that newcomers will run into this problem but every system out there, that is running the older code, is leaking memory in such a way that the system will degrade to a point where it will run out of memory completely and fail. I think it deserves some kind of special mention.

Issue #628, opened 2020-05-06
Appinsights breaks node's http.createServer function


Pixelatex:
So after a lot of digging and an [issue](https://github.com/nodejs/node/issues/33245) at the node repo, I was pointed here.

The culprit are these lines:

https://github.com/microsoft/ApplicationInsights-node.js/blob/44a0e7f6454a859a15f31b43c448513e92d01855/AutoCollection/HttpRequests.ts#L150-L155

Node should have an option for the HTTP create server as well. disabling app insights from my code proves that it works.

Pixelatex:
I opened a PR to add the missing options parameter

DevRCRun:
We're currently looking at trialling appinsights and I think this would cause an issue for us too. I note the PR is still outstanding. Anyone at MS able to provide an update / eta?

hectorhdzg:
Fixed #629

PR #629, opened 2020-05-06
Http.createServer is missing the options variable

Pixelatex:
As mentioned in [this](https://github.com/microsoft/ApplicationInsights-node.js/issues/628) issue. The http.createServer function was missing the options that the original node function has.

I simply added the missing options.


AutoCollection/HttpRequests.ts
```diff
@@ -147,9 +147,9 @@ class AutoCollectHttpRequests {
         }
 
         const originalHttpServer = http.createServer;
-        http.createServer = (onRequest) => {
+        http.createServer = (options: https.ServerOptions, onRequest?: Function) => {
```
markwolff, line 150: I believe `options` is also [optional](https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener).

Pixelatex: Same with the https function though.

Also, this is how it's documented and works by node:

```
http.createServer([options][, requestListener])
```

markwolff: The current definition specifies provides 2 overloads. 1 optional callback or `options` and 1 optional callback. This patch should handle `options` being that optional callback or the required `options`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f1008ba6f314c49abe8747e1da52ef684bce9d29/types/node/v12/http.d.ts#L355-L356

Pixelatex: Correct! My bad! I added the overload

markwolff: Awesome, thanks! Last thing would be that javascript does not actually support function overloading. Although Typescript "supports" it for types, this is all lost when compiled to usable javascript. A function can only be defined once, and it must "support" all of the possible overload shapes

https://www.typescriptlang.org/docs/handbook/functions.html#overloads

Pixelatex: So there seems to be an issue. 
`originalHttpServer(options, wrapOnRequestHandler(onRequest));` won't accept a second parameter.

This seems to be defined as such in the `@types/node`  as:
`    export function createServer(requestListener?: (request: IncomingMessage, response: ServerResponse) =>void ): Server;`

With no additional overloads :/

Pixelatex: App insights seems to be using an older version of node types, they're at v13 while the project seems to be at v4 right now?

Pixelatex: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33011 is the PR that added the fix to the node types, so the lowest we'd need to update to is `10.12.25`

However, this causes a cascading  effect of issues by jumping 6 major versions up :s

Pixelatex: @markwolff Do you have some ideas on how to tackle this?

markwolff: You can update to the min node typings you need. We have a separate test `npm run backcompat` which tests compiling on older node typings. I'll add v4 to this list before cutting a release for this.

Pixelatex: Alright, I'll try to fix the issues but some are in `AutoCollection/Exceptions` for example and those might need refactoring to work with the new typings.

Pixelatex: @markwolff there seem to be quite a few issues:

`AutoCollection/Exceptions.ts`
Line: 66~

```
this._exceptionListenerHandle = handle.bind(this, true, AutoCollectExceptions.UNCAUGHT_EXCEPTION_HANDLER_NAME);

  process.on(AutoCollectExceptions.UNCAUGHT_EXCEPTION_HANDLER_NAME, this._exceptionListenerHandle);
```

This listener is defined as `(reThrow:boolean, error: Error) => void`, but the type definitions are expecting the type to be:

`   type UncaughtExceptionListener = (error: Error) => void;`

And no matter what I did to fix it. It kept nagging about the static variables.

adding 'as any' silenced the warning but not sure if this is a good solution

markwolff: I'm not sure why `as any` there silences the problem, as that field is a string. We are binding the 1st arg of the handler to the rethrow, so from `process.on` perspective, we should be passing it a `string` and a `(error: Error) => void` type.

Pixelatex: The typescript overloads don't seem to recognize public static variables as a variable, they try to match it to the signals pipe.

hectorhdzg:
@Pixelatex I updated the code slightly and checks are successful now, let me know if you have any comments. We will add integration tests with node.js > 9 using options parameter in another PR and update node types to latest as well.

xiao-lix (approved):

PR #635, opened 2020-05-22
do not create resend timer if exists

### LINKED ISSUE #634: Potential memory leak?
Hi App insights team, 

I'm trying to identify a memory leak in one of our apps. 

Please see the below screenshot

Screenshot 2020-05-22 at 16 21 04

Comparing the two memory dumps, seems that there are many closures being added to the memory from this line of code.

https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/Library/Sender.ts#L138

```
setTimeout(() => this._sendFirstFileOnDisk(), this._resendInterval).unref();
```

Quickly commenting out that line of code seems to fix the problem.

What are your thoughts about this? 

Thanks!

markwolff:
Fixes #634

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -33,6 +33,7 @@ class Sender {
     private _onError: (error: Error) => void;
     private _enableDiskRetryMode: boolean;
     private _numConsecutiveFailures: number;
+    private _resendTimer: NodeJS.Timer | null;
     protected _resendInterval: number;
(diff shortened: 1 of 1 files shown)


asos-VasilisKortsimelidis:
@markwolff Hi Mark I tried your code and indeed now I see only one reference of the closure on memory
Screenshot 2020-05-26 at 11 26 51

PR #660, opened 2020-07-31
Heartbeat metric of web app and function app

xiao-lix:


AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    public isInitialized() {
+        return this._isInitialized;
+    }
+
+    public static isEnabled() {
+        return HeartBeat.INSTANCE && HeartBeat.INSTANCE._isEnabled;
+    }
+
+    public trackHeartBeat(config: Config, callback: () => void) {
```
xiao-lix, line 60: hi @markwolff , is this the callback you mentioned? @MSNev helped me work this out here. Do you see any issues? The promisify from util is "undefined" here.

Tests/AutoCollection/Heartbeat.tests.ts
```diff
@@ -0,0 +1,249 @@
...
+                assert.equal(properties3["azInst_subscriptionId"], 2, "azInst_subscriptionId is read from response");
+                assert.equal(properties3["azInst_osType"], "Linux", "azInst_osType is read from response");
+
+                stub3.restore();
+                heartbeat3.dispose();
+                done();
+            });
+        });
+
+        it("should read correct VM information from response", (done) => {
```
xiao-lix, line 161: @markwolff @MSNev I'm still not able to make this VM test work.. I have two versions. heartbeat3 is using nock... heartbeat4 is to stub out makeRequest, but I'm not sure how to give the callback of "makeRequest" a parameter (fake res)? please see comment below.

Tests/AutoCollection/Heartbeat.tests.ts
```diff
@@ -0,0 +1,249 @@
...
+            const stub4 = sinon.stub(heartbeat4["_client"], "trackMetric");
+
+            heartbeat4["_vmData"] = {
+                vmId: "1",
+                subscriptionId: "2",
+                osType: "Linux"
+            };
+            heartbeat4["_isVM"] = true;
+            // sinon.stub(Util, "makeRequest").returns(new fakeResponse(false));
+            sinon.stub(Util, 'makeRequest', (config: any, requestUrl: string, requestOptions: any, requestCallback: Function) => {
```
xiao-lix, line 174: Without this stub, the trackHeartBeat will call _getAzureComputeMetadata which calls Util.makeRequest that will assign _isVM = false (reset what I set on line 172) so the assertions won't pass. This stub here is trying to force makeRequest return a fakeresponse with statusCode=200. But the way I code here will report error as "statusCode undefined". @markwolff @MSNev Can you please take a look. I think heartbeat4 test is more straightforward than heartbeat3 test... Thanks a lot! I'm stuck on this test for too long... 😩

lzchen:
@xiao-lix 
Can you include a description for exactly what features are included here and the purpose?
Also, a more descriptive title would be useful.

AutoCollection/HeartBeat.ts
markwolff: Yes this looks good

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+import Constants = require("../Declarations/Constants");
+import Util = require("../Library/Util");
+import Config = require("../Library/Config");
+import Context = require("../Library/Context");
+
+class HeartBeat {
+
+    public static INSTANCE: HeartBeat;
+
+    private static AIMS_URI = "http://169.254.169.254/metadata/instance/compute";
```
MSNev, line 12: Do these need to be available on the object or are they internal?
If only used internally then I would suggest making them const values outside of the class -- for better minification

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+
+        if (isEnabled) {
+            if (!this._handle) {
+                this._handle = setInterval(() => this.trackHeartBeat(config, () => {}), this._collectionInterval);
+                this._handle.unref(); // Allow the app to terminate even while this loop is going on
+            }
+        } else {
+            if (this._handle) {
+                clearInterval(this._handle);
+                this._handle = undefined;
```
markwolff, line 47: Change the type of `_handle` to `NodeJS.Timer | null` and set to `null` here

Tests/AutoCollection/Heartbeat.tests.ts
MSNev: As your using sinon, you could use the sinon.useFakeServer() which provides a mock XHR request that you can respond() with the status and values inline -- rather than wait for a real response etc

Tests/AutoCollection/Heartbeat.tests.ts
MSNev: Using the useFakeRequest() to avoid needing to mock like this -- plenty of examples in 1P/3P Js code bases.

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+    constructor(client: TelemetryClient) {
+        if (!HeartBeat.INSTANCE) {
+            HeartBeat.INSTANCE = this;
+        }
+
+        this._isInitialized = false;
+        this._client = client;
+    }
+
+    public enable(isEnabled: boolean, config?: Config) {
```
lzchen, line 33: How is this being used? I do not believe heartbeat can be configurable.

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+            properties["azfunction_appId"] = process.env.WEBSITE_HOSTNAME;
+        } else if (config) { // VM
+            waiting = true;
+            this._getAzureComputeMetadata(config, () => {
+                if (this._isVM && Object.keys(this._vmData).length > 0) {
+                    properties["azInst_vmId"] = this._vmData["vmId"] || "";
+                    properties["azInst_subscriptionId"] = this._vmData["subscriptionId"] || "";
+                    properties["azInst_osType"] = this._vmData["osType"] || "";
+                }
+                this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
```
lzchen, line 80: What is the purpose of `waiting`?

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+        const sdkVersion = Context.sdkVersion; // "node" or "node-nativeperf"
+        properties["sdk"] = sdkVersion;
+        properties["osType"] = os.type();
+        if (process.env.WEBSITE_SITE_NAME) { // Web apps
+            properties["appSrv_SiteName"] = process.env.WEBSITE_SITE_NAME || "";
+            properties["appSrv_wsStamp"] = process.env.WEBSITE_HOME_STAMPNAME || "";
+            properties["appSrv_wsHost"] = process.env.WEBSITE_HOSTNAME || "";
+        } else if (process.env.FUNCTIONS_WORKER_RUNTIME) { // Function apps
+            properties["azfunction_appId"] = process.env.WEBSITE_HOSTNAME;
+        } else if (config) { // VM
```
lzchen, line 72: Does this mean that every case that doesn't hit the above to will be assumed to be VM?

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+            waiting = true;
+            this._getAzureComputeMetadata(config, () => {
```
lzchen, line 74: Won't this continuously call the azure metadata service even if the app is already found to be not part of a VM?

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+                if (this._isVM && Object.keys(this._vmData).length > 0) {
+                    properties["azInst_vmId"] = this._vmData["vmId"] || "";
+                    properties["azInst_subscriptionId"] = this._vmData["subscriptionId"] || "";
+                    properties["azInst_osType"] = this._vmData["osType"] || "";
+                }
+                this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
+                callback();
```
lzchen, line 81: Nit: Why not just have `callback()` called at the end instead instead of in these if blocks?

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+                res.on('data', (data: any) => {
+                    data += data;
+                });
+                res.on('end', () => {
+                    this._vmData = JSON.parse(data);
+                    callback();
+                });
+            } else if (res.statusCode >= 400 && res.statusCode < 500) {
+                // Not found; Not in VM; Do not try again.
+                this._isVM = false;
```
lzchen, line 119: I think even if you set this, you still call `_getAzureComputeMetadata` right?

AutoCollection/HeartBeat.ts
markwolff: This is the pattern followed by all autocollectors in this SDK. If a particular autocollector should not be configurable, `setAutoCollectHeartbeat` should be omitted from `applicationinsights.ts` instead.

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+    private _getAzureComputeMetadata(config: Config, callback: () => void) {
+        const metadataRequestUrl = `${HeartBeat.AIMS_URI}?${HeartBeat.AIMS_API_VERSION}&${HeartBeat.AIMS_FORMAT}`;
+        const requestOptions = {
+            method: 'GET',
+            headers: {
+                "Metadata": "true"
+            }
+        };
+
+        const req = Util.makeRequest(config, metadataRequestUrl, requestOptions, (res) => {
```
lzchen, line 105: There was a bug in Opencensus in which the application hangs if there is no timeout set for the request. Might be the same case here.

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,138 @@
...
+                    this._vmData = JSON.parse(data);
+                    callback();
+                });
+            } else if (res.statusCode >= 400 && res.statusCode < 500) {
+                // Not found; Not in VM; Do not try again.
+                this._isVM = false;
+                callback();
+            } else {
+                // else Retry on next heartbeat metrics call
+                callback();
```
lzchen, line 123: I'm not sure how `callback` works. Does this give control back to the outer function? Wouldn't that mean that the below logic will never be run?

AutoCollection/HeartBeat.ts
xiao-lix: To check if it's waiting on getting VM information. If true, then wait until the VM information retrieved then to call trackMetric

AutoCollection/HeartBeat.ts
xiao-lix: I moved `// VM` to line 75. This means if not web app or function app, it goes into `else if (config)` to check if it's VM. And only if _getAzureComputeMetadata set `_isVM` to true and assign correct `_vmData` then it confirmed to be VM. Otherwise trackMetric is sent with name, value, and sdk and osType properties only.

AutoCollection/HeartBeat.ts
xiao-lix: What do you mean? If the first or second if` true`, it breaks and won't run third if.

AutoCollection/HeartBeat.ts
xiao-lix: Because checking for environment variables no need to call callback.

AutoCollection/HeartBeat.ts
xiao-lix: Sorry I don't get what you mean 🤔

AutoCollection/HeartBeat.ts
```diff
@@ -0,0 +1,143 @@
...
+                this._isVM = true;
+                let virtualMachineData = "";
+                res.on('data', (data: any) => {
+                    virtualMachineData += data;
+                });
+                res.on('end', () => {
+                    this._vmData = JSON.parse(virtualMachineData);
+                    callback();
+                });
+                if (this._isVM && Object.keys(this._vmData).length > 0) { // add this to make sure test call callback after return res with statusCode 200
```
xiao-lix, line 119: This if added to make sure VM test call callback() - fake response has no res.on('data') or res.on('end') event listener, without this check the callback will never gets called. @markwolff does this make sense? I was thinking to add a fakeResponse class like what E2E test does, but nock response doesn't allow to add self defined class as response body.

AutoCollection/HeartBeat.ts
```diff
@@ -94,26 +95,30 @@ class HeartBeat {
...
 
     private _getAzureComputeMetadata(config: Config, callback: () => void) {
-        const metadataRequestUrl = `${HeartBeat.AIMS_URI}?${HeartBeat.AIMS_API_VERSION}&${HeartBeat.AIMS_FORMAT}`;
+        const metadataRequestUrl = `${AIMS_URI}?${AIMS_API_VERSION}&${AIMS_FORMAT}`;
         const requestOptions = {
             method: 'GET',
             headers: {
-                "Metadata": "true"
+                "Metadata": "true",
+                [AutoCollectHttpDependencies.disableCollectionRequestOption]: true,
```
markwolff, line 103: move this from `requestOptions.headers` to `requestOptions`

AutoCollection/HeartBeat.ts
markwolff: Yes we can look into this in a later PR. I think nock is conflicting with sinon fakeResponse stuff, so we should remove nock altogether and use fakeResponse like the e2e tests already use.

AutoCollection/HeartBeat.ts
markwolff: I think to solve this, you should only call `_getAzureComputeMetadata` when `this._isVM` is `undefined`.

AutoCollection/HeartBeat.ts
xiao-lix: I'll remove VM part in this PR but why move headers out to requestOptions? I see the interface RequestOptions has `headers?: { [key: string]: any };` property?
image

markwolff: only `[AutoCollectHttpDependencies.disableCollectionRequestOption]: true,` should be moved options. It should not be a header

PR #661, opened 2020-08-03
add support for URL dependency parsing

markwolff:
`got(https://example.com)` sends requests through to the `http` module with an instance of a `URL` class. Previously, the parsing here would not pick up the class prototype methods, so it looks like an empty object and defaults to a `localhost` dependency.

https://nodejs.org/api/https.html#https_https_get_url_options_callback

### DIFF
--- a/AutoCollection/HttpDependencyParser.ts
+++ b/AutoCollection/HttpDependencyParser.ts
@@ -16,7 +16,7 @@ import CorrelationIdManager = require("../Library/CorrelationIdManager");
 class HttpDependencyParser extends RequestParser {
     private correlationId: string;
 
-    constructor(requestOptions: URL | string | http.RequestOptions | https.RequestOptions, request: http.ClientRequest) {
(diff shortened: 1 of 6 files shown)


AutoCollection/HttpDependencyParser.ts
```diff
@@ -114,6 +114,8 @@ class HttpDependencyParser extends RequestParser {
     private static _getUrlFromRequestOptions(options: any, request: http.ClientRequest) {
         if (typeof options === 'string') {
             options = url.parse(options);
+        } else if (options && typeof (url as any).URL === 'function' && options instanceof (url as any).URL) {
```
xiao-lix, line 117: I don't understand (url as any).URL, does `URL` exist on module `url`?

markwolff: @xiao-lix
It exists only in node 7+
https://nodejs.org/api/url.html#url_class_url

Since this SDK uses older node types, I `any` cast it for now.

PR #666, opened 2020-08-07
Add heartbeat metric for VM

xiao-lix:
Continuation of [#660 ].

Detects whether current app is hosted in an Azure VM. We utilize [AzureMetaDataService ](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service)to retrieve vm information.

### DIFF
--- a/AutoCollection/HeartBeat.ts
+++ b/AutoCollection/HeartBeat.ts
@@ -21,6 +21,9 @@ class HeartBeat {
     private _isInitialized: boolean;
     private _isVM: boolean;
     private _vmData = <{[key: string]: string}>{};
+    private _azInst_vmId: string = "";
+    private _azInst_subscriptionId: string = "";
+    private _azInst_osType: string = "";
 
     constructor(client: TelemetryClient) {
         if (!HeartBeat.INSTANCE) {
@@ -70,17 +73,26 @@ class HeartBeat {
             properties["appSrv_wsHost"] = process.env.WEBSITE_HOSTNAME || "";
         } else if (process.env.FUNCTIONS_WORKER_RUNTIME) { // Function apps
             properties["azfunction_appId"] = process.env.WEBSITE_HOSTNAME;
-        } else if (config && this._isVM === undefined) {
-            waiting = true;
-            this._getAzureComputeMetadata(config, () => {
-                if (this._isVM && Object.keys(this._vmData).length > 0) { // VM
-                    properties["azInst_vmId"] = this._vmData["vmId"] || "";
-                    properties["azInst_subscriptionId"] = this._vmData["subscriptionId"] || "";
-                    properties["azInst_osType"] = this._vmData["osType"] || "";
-                }
-                this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
-                callback();
-            });
+        } else if (config) {
+            if (this._isVM === undefined) {
+                waiting = true;
+                this._getAzureComputeMetadata(config, () => {
+                    if (this._isVM && Object.keys(this._vmData).length > 0) { // VM
+                        properties["azInst_vmId"] = this._vmData["vmId"] || "";
+                        properties["azInst_subscriptionId"] = this._vmData["subscriptionId"] || "";
+                        properties["azInst_osType"] = this._vmData["osType"] || "";
+                        this._azInst_vmId = this._vmData["vmId"] || "";
+                        this._azInst_subscriptionId = this._vmData["subscriptionId"] || "";
+                        this._azInst_osType = this._vmData["osType"] || "";
+                    }
+                    this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
+                    callback();
+                });
+            } else if (this._isVM) {
+                properties["azInst_vmId"] = this._azInst_vmId;
+                properties["azInst_subscriptionId"] = this._azInst_subscriptionId;
+                properties["azInst_osType"] = this._azInst_osType;
+            }
         }
         if (!waiting) {
             this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
@@ -100,7 +112,7 @@ class HeartBeat {
             method: 'GET',
             [AutoCollectHttpDependencies.disableCollectionRequestOption]: true,
             headers: {
-                "Metadata": "true",
+                "Metadata": "True",
             }
         };
 
--- a/AutoCollection/HeartBeat.ts
+++ b/AutoCollection/HeartBeat.ts
@@ -1,8 +1,14 @@
 import os = require("os");
 import TelemetryClient = require("../Library/TelemetryClient");
 import Constants = require("../Declarations/Constants");
+import Util = require("../Library/Util");
 import Config = require("../Library/Config");
 import Context = require("../Library/Context");
+import AutoCollectHttpDependencies = require("../AutoCollection/HttpDependencies");
+
+const AIMS_URI = "http://169.254.169.254/metadata/instance/compute";
+const AIMS_API_VERSION = "api-version=2017-12-01";
+const AIMS_FORMAT = "format=json";
 
 class HeartBeat {
 
@@ -13,6 +19,8 @@ class HeartBeat {
     private _handle: NodeJS.Timer | null;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
+    private _isVM: boolean;
+    private _vmData = <{[key: string]: string}>{};
 
     constructor(client: TelemetryClient) {
         if (!HeartBeat.INSTANCE) {
@@ -51,6 +59,7 @@ class HeartBeat {
     }
 
     public trackHeartBeat(config: Config, callback: () => void) {
+        let waiting: boolean = false;
         let properties: {[key: string]: string} = {};
         const sdkVersion = Context.sdkVersion; // "node" or "node-nativeperf"
         properties["sdk"] = sdkVersion;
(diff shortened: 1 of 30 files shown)


Tests/EndToEnd.tests.ts
```diff
@@ -53,7 +58,7 @@ class fakeResponse {
     }
 
     public pass(test = false): void {
-        this.callbacks["data"] ? this.callbacks["data"]("data") : null;
+        this.callbacks["data"] ? this.callbacks["data"](this._responseData) : null;
```
xiao-lix, line 61: Modified this in order to have response return a JSON object as response data, other than string "data".

Tests/EndToEnd.tests.ts
```diff
@@ -70,8 +75,11 @@ class fakeRequest {
     public write(): void { }
     public headers: { [id: string]: string } = {};
     public agent = { protocol: 'http' };
+    private _responseData: any;
```
xiao-lix, line 78: Add this variable so when creating a fakeRequest, I can specify what response data I'll receive. So when it creates a faceResponse instance on line 108, it passes in this responseData.

Tests/EndToEnd.tests.ts
```diff
@@ -968,4 +976,94 @@ describe("EndToEnd", () => {
...
+
+        it("should collect correct VM information from JSON response", (done) => {
+            // set up stub
+            const vmDataJSON = `{
+                "vmId": "1",
+                "subscriptionId": "2",
+                "osType": "Windows_NT"
+            }`;
+            var stub: sinon.SinonStub = sandbox.stub(http, "request", (options: any, callback: any) => {
+                var req = new fakeRequest(false, "http://169.254.169.254", vmDataJSON);
```
xiao-lix, line 999: make sure I return a JSON object as response - then metrics properties populate correct vm info from the respond data.

Tests/EndToEnd.tests.ts
```diff
@@ -968,4 +976,94 @@ describe("EndToEnd", () => {
...
+                heartbeat.dispose();
+                stub.restore();
+                done();
+            });
+        });
+
+        it("should only send name and value properties for heartbeat metric when get VM request fails", (done) => {
+            // set up stub
+            var stub: sinon.SinonStub = sandbox.stub(http, "request", (options: any, callback: any) => {
+                var req = new fakeRequest(true, "http://169.254.169.254");
```
xiao-lix, line 1039: Note the different of this req with line 1000 is req.on('end') not exist - so the response is null. and the fakeRequest fails immediately - so the get vm API throws error.

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +70,79 @@ class HeartBeat {
...
+            waiting = true;
+            this._getAzureComputeMetadata(config, () => {
+                if (this._isVM && Object.keys(this._vmData).length > 0) { // VM
+                    properties["azInst_vmId"] = this._vmData["vmId"] || "";
+                    properties["azInst_subscriptionId"] = this._vmData["subscriptionId"] || "";
+                    properties["azInst_osType"] = this._vmData["osType"] || "";
+                }
+                this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
+                callback();
+            });
```
markwolff, line 83: `trackMetric` here will only get called the first time. You should call `_getAzureComputeMetadata` only once ever, but trackMetric should still be called at the interval

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +70,79 @@ class HeartBeat {
...
+        const req = Util.makeRequest(config, metadataRequestUrl, requestOptions, (res) => {
+            if (res.statusCode === 200) {
+                // Success; VM
+                this._isVM = true;
+                let virtualMachineData = "";
+                res.on('data', (data: any) => {
+                    virtualMachineData += data;
+                });
+                res.on('end', () => {
+                    this._vmData = this._isJSON(virtualMachineData) ? JSON.parse(virtualMachineData) : {};
```
markwolff, line 116: For the else case here, is it still guaranteed it is a VM? Maybe we can fill in some fallback defaults here for trying to figure out the OS, that we know it is a VM, etc

AutoCollection/HeartBeat.ts
xiao-lix: @markwolff I see, that makes sense. Do you think it makes sense to write `_isVM`, and other properties `azInst_vmId`, `azInst_subscriptionId`, and `azInst_osType` to environment variable?

markwolff: I think you can just write them to `this`

AutoCollection/HeartBeat.ts
xiao-lix: I thought about how to reply your comment on this question, I added this check was in order to pass test... 😅 because the E2E test used a string "data" as default response data content, and failed JSON.parse, so added this to make sure the test won't break. But I get what you mean, and read the doc but don't find any case that it won't return a JSON. Check this [sample](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#sample-3-getting-more-information-about-the-vm-during-support-case). The `.../metadata/instance/compute?...` API is to get more information of the VM instance, and it returns a JSON object. So in here, if the status code is 200, it should return a JSON string... so I don't think we need a fallback? WDYT?

AutoCollection/HeartBeat.ts
xiao-lix: oh! I misunderstood. I thought heartmetric instance is created each time... I see.. will update thanks.

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +73,88 @@ class HeartBeat {
...
         this._isInitialized = false;
     }
+
+    private _getAzureComputeMetadata(config: Config, callback: () => void) {
+        const metadataRequestUrl = `${AIMS_URI}?${AIMS_API_VERSION}&${AIMS_FORMAT}`;
+        const requestOptions = {
+            method: 'GET',
+            [AutoCollectHttpDependencies.disableCollectionRequestOption]: true,
+            headers: {
+                "Metadata": "True",
```
xiao-lix, line 115: Update to `True` based on this [sample](https://github.com/microsoft/azureimds/blob/bcdd07c4a1f328ab6d0c48263dd8c61b600111f1/IMDSSample.js#L14)

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +73,88 @@ class HeartBeat {
...
+            }
+        };
+
+        const req = Util.makeRequest(config, metadataRequestUrl, requestOptions, (res) => {
```
lzchen, line 119: I think I made this comment on the other PR. There was a bug in OpenCensus Python implementation of heartbeat where an application hangs because the request to azuremetadataservice did not specify a timeout. Might want to do that here too to be safe?

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +73,88 @@ class HeartBeat {
             properties["appSrv_wsHost"] = process.env.WEBSITE_HOSTNAME || "";
         } else if (process.env.FUNCTIONS_WORKER_RUNTIME) { // Function apps
             properties["azfunction_appId"] = process.env.WEBSITE_HOSTNAME;
-        } 
-        this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
-        callback();
+        } else if (config) {
```
lzchen, line 76: I'm not too familiar on how config is defined. Will this ever be undefined?

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +73,88 @@ class HeartBeat {
...
+                        properties["azInst_osType"] = this._vmData["osType"] || "";
+                        this._azInst_vmId = this._vmData["vmId"] || "";
+                        this._azInst_subscriptionId = this._vmData["subscriptionId"] || "";
+                        this._azInst_osType = this._vmData["osType"] || "";
+                    }
+                    this._client.trackMetric({name: Constants.HeartBeatMetricName, value: 0, properties: properties});
+                    callback();
+                });
+            } else if (this._isVM) {
+                properties["azInst_vmId"] = this._azInst_vmId;
```
lzchen, line 92: You need to continuously make requests to the azuremetadataservice because the vm fields can change if you find that the application in running in a VM. This is the whole point of tracking `_isVM`.

AutoCollection/HeartBeat.ts
```diff
@@ -61,16 +73,88 @@ class HeartBeat {
...
+            } else {
+                // else Retry on next heartbeat metrics call
+                callback();
+            }
+        });
+        if (req) {
+            req.on('error', (error: Error) => {
+                // Unable to contact endpoint.
+                // Do nothing for now.
+                this._isVM = false;
```
lzchen, line 144: How come we don't attempt to retry if we get a different exception besides not being in a vm?

AutoCollection/HeartBeat.ts
lzchen: In the Python [implementation](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py#L105) there are certain exceptions that are thrown when the app is not a vm (the endpoint simply doesn't exist). This should be the same case for Node correct? I'm not sure how the http client you are using handles exceptions, but I'm curious if a status code is even returned?

xiao-lix: The status code should get returned, check this [sample ](https://github.com/microsoft/ApplicationInsights-node.js/blob/53b440bdcb2eaa1cd6a04f125533fbc5c2f0ab10/Library/CorrelationIdManager.ts#L54) of imds api for node js. So if the status code is not 200, the request fails that it's not a vm. So here in my implementation, if the request fails it never gets to line 116 to check if the response data is JSON or not. And I learned the if elseif else from code [here](https://github.com/microsoft/ApplicationInsights-node.js/blob/53b440bdcb2eaa1cd6a04f125533fbc5c2f0ab10/Library/CorrelationIdManager.ts#L54) from node sdk repo.

AutoCollection/HeartBeat.ts
xiao-lix: The config param is only used when checking if it's vm and making http request, because the first param for Util.makeRequest is config. So I pass config as an optional param. For web app and function app it's not necessary.

AutoCollection/HeartBeat.ts
xiao-lix: oh ok.. i'm confused. @markwolff so should I call _getAzureComputeMetadata multiple times then?

AutoCollection/HeartBeat.ts
MSNev: I believe this is an async request and therefore the code will not block -- additionally generally XHR requests will fail with a timeout @markwolff Is this your understanding of the Util.makeRequest() as well?

AutoCollection/HeartBeat.ts
xiao-lix: So is it only confirm it's not vm when connection error happens? if I understand your [implementation ](https://github.com/census-instrumentation/opencensus-python/blob/67502e512ab073382b4fd12580c3b4534968129b/contrib/opencensus-ext-azure/opencensus/ext/azure/metrics_exporter/heartbeat_metrics/heartbeat.py#L105)correctly. But @markwolff I don't think there's a way to check it like python does? My assumption is 'error' === can't connect === not vm

lzchen: I don't think you should make an assumption. I suggest you should try out azuremetadataservice yourself, within a VM, outside a VM, within a VM but exception happens, outside a VM and exception happens and see how it behaves.

AutoCollection/HeartBeat.ts
lzchen: I only see `config` being used `Util.makeRequest`. What is actually being used in `config` that `makeRequest` uses? I do not believe we have any user configurable things so why have a config at all?

AutoCollection/HeartBeat.ts
xiao-lix: yes i did. outside a vm is Error: connect ENETUNREACH, it's the connection error, in a vm it returns JSON.

AutoCollection/HeartBeat.ts
lzchen: Right in Python the request is made via a separate thread. I'm not sure if this is even a problem in NodeJS (this only came up when a user tried to initialize a Django app). So the real question is does `Util.makeRequest()` EVER hang if a timeout is not set? The `requests` library MUST set a timeout or the request itself will wait forever, regardless of whether it is done asynchronously or not.

AutoCollection/HeartBeat.ts
xiao-lix: And I think the statuscode scenarios is similar to what you check for exception types. 

`1xx informational response – the request was received, continuing process` - retry
`2xx successful – the request was successfully received, understood, and accepted` - vm true
`3xx redirection – further action needs to be taken in order to complete the request` - retry
`4xx client error – the request contains bad syntax or cannot be fulfilled` - vm false
`5xx server error – the server failed to fulfil an apparently valid request` - retry

AutoCollection/HeartBeat.ts
lzchen: I'm not referring to how other apps handle error cases with status code. I'm asking about the functionality of the `http` library that is used in NodeJS (which `Util.makeRequests calls under the hood). Looking at the [docs](https://nodejs.org/api/http.html#http_http_request_options_callback) it looks like all errors are piped to `req.on('error...` so it should be handled already.

AutoCollection/HeartBeat.ts
lzchen: Right, but when you are not in a VM, and it throws Error: connect ENETUNREACH, does that mean it still returns a RESPONSE with a status code? If it does not, you can't rely on `else if (res.statusCode >= 400 && res.statusCode < 500) {` to be doing the check correct?

AutoCollection/HeartBeat.ts
markwolff: If a user has specific proxy/agent security requirements for their services, we must honor those for all http(s) requests made by this SDK.

AutoCollection/HeartBeat.ts
xiao-lix: update in next commit

AutoCollection/HeartBeat.ts
xiao-lix: discussed offline, update in next commit

PR #667, opened 2020-08-07
Instrumentation Key validation.

kryalama:
Specs taken from breeze repo(https://dev.azure.com/msazure/One/_git/Mgmt-AppInsights-Breeze)
The definition of a VALID instrumentation key is as follows:

Not none
Not empty
Every character is a hex character [0-9a-f]
32 characters are separated into 5 sections via 4 dashes
First section has 8 characters
Second section has 4 characters
Third section has 4 characters
Fourth section has 4 characters
Fifth section has 12 characters

As of now we are just logging a warning, so that it dont break existing applications.

### DIFF
--- a/Library/Config.ts
+++ b/Library/Config.ts
@@ -141,7 +141,7 @@ class Config {
 
     /**
     * Validate UUID Format
-    * Specs taken from breeze repo(https://dev.azure.com/msazure/One/_git/Mgmt-AppInsights-Breeze)
+    * Specs taken from breeze repo
     * The definition of a VALID instrumentation key is as follows:
     * Not none
     * Not empty
--- a/Library/Config.ts
+++ b/Library/Config.ts
@@ -1,5 +1,6 @@
 import CorrelationIdManager = require('./CorrelationIdManager');
 import ConnectionStringParser = require('./ConnectionStringParser');
+import Logging = require('./Logging');
 import Constants = require('../Declarations/Constants');
 import http = require('http');
 import https = require('https');
@@ -71,6 +72,10 @@ class Config {
             : setupString; // CS was invalid, so it must be an ikey
 
         this.instrumentationKey = csCode.instrumentationkey || iKeyCode /* === instrumentationKey */ || csEnv.instrumentationkey || Config._getInstrumentationKey();
+        // validate ikey. If fails throw a warning
+        if(!Config._validateInstrumentationKey(this.instrumentationKey)) {
+            Logging.warn("instrumentation key validation failed");
+        }
 
         this.endpointUrl = `${csCode.ingestionendpoint || csEnv.ingestionendpoint || this.endpointBase}/v2/track`;
         this.maxBatchSize = 250;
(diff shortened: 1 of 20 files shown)


Tests/Library/Config.tests.ts
```diff
@@ -126,15 +126,37 @@ describe("Library/Config", () => {
...
+                assert.ok(warnStub.notCalled, "warning was not raised");
+                warnStub.restore();
+            });
+
+            it("instrumentation key validation-invalid key passed", () => {
+                var warnStub = sinon.stub(console, "warn");
+                var config = new Config("abc");
+                assert.ok(warnStub.calledOn, "warning was not raised");
+                warnStub.restore();
+            });
```
markwolff, line 158: should these be failing?

Library/Config.ts
```diff
@@ -71,6 +72,10 @@ class Config {
             : setupString; // CS was invalid, so it must be an ikey
 
         this.instrumentationKey = csCode.instrumentationkey || iKeyCode /* === instrumentationKey */ || csEnv.instrumentationkey || Config._getInstrumentationKey();
+        // validate ikey. If fails throw a warning
+        if(!Config._validateInstrumentationKey(this.instrumentationKey)) {
+            Logging.warn("instrumentation key validation failed");
```
markwolff, line 77: Include `this.instrumentionKey` as parameter to `Logging.warn`

Tests/Library/Config.tests.ts
kryalama: actually we accept either with 4(dashes) or no dashes.
 And I will fix the second message. It should have been "warning was raised."

Library/Config.ts
```diff
@@ -133,6 +138,18 @@ class Config {
...
         return iKey;
     }
+
+    /**
+    * Validate UUID Format
+    * Specs taken from https://tools.ietf.org/html/rfc4122
+    */    
+    private static _validateInstrumentationKey(iKey:string): boolean {
+        const UUID_Regex = '^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$';
+        const UUID_Regex_noDash = '^[0-9a-f]{8}[0-9a-f]{4}[1-5][0-9a-f]{3}[89ab][0-9a-f]{3}[0-9a-f]{12}$';
```
lzchen, line 148: Does breeze really accept no dash ikeys?

Tests/Library/Config.tests.ts
kryalama: fixed

Library/Config.ts
kryalama: fixed

Library/Config.ts
```diff
@@ -74,7 +74,7 @@ class Config {
         this.instrumentationKey = csCode.instrumentationkey || iKeyCode /* === instrumentationKey */ || csEnv.instrumentationkey || Config._getInstrumentationKey();
         // validate ikey. If fails throw a warning
         if(!Config._validateInstrumentationKey(this.instrumentationKey)) {
-            Logging.warn("instrumentation key validation failed");
+            Logging.warn("invalid instrumentation key - "+this.instrumentationKey);
```
markwolff, line 77:
```suggestion
            Logging.warn("An invalid instrumentation key was provided. There may be resulting telemetry loss", this.instrumentationKey);
```

Library/Config.ts
```diff
@@ -133,6 +138,16 @@ class Config {
 
         return iKey;
     }
+
+    /**
+    * Validate UUID Format
+    * Specs taken from https://tools.ietf.org/html/rfc4122
```
lzchen, line 144: Since we are following what Breeze is doing now, I suggest linking to that code snippet instead of the UUID Format doc.

Library/Config.ts
kryalama: fixed

Library/Config.ts
```diff
@@ -133,6 +138,26 @@ class Config {
 
         return iKey;
     }
+
+    /**
+    * Validate UUID Format
+    * Specs taken from breeze repo(https://dev.azure.com/msazure/One/_git/Mgmt-AppInsights-Breeze)
```
markwolff, line 144: I would just remove this link altogether. It's not accessible externally

PR #678, opened 2020-09-08
[Codeless] change format of diag logger output

markwolff:
Changes required to get attach-time telemetry to work end-to-end
- Adds required `properties` values for PaaS environments
- Changes JSON output to be stringified to a single line instead of using standard 2-space indentation format
- Writes `date` using correct ISO string format

### DIFF
--- a/Tests/Bootstrap/DiagnosticLogger.spec.ts
+++ b/Tests/Bootstrap/DiagnosticLogger.spec.ts
@@ -1,11 +1,55 @@
 import assert = require("assert");
+import sinon = require("sinon");
+import { AgentLogger } from "../../Bootstrap/DataModel";
 import { DiagnosticLogger } from "../../Bootstrap/DiagnosticLogger";
+import { NoopLogger } from "../../Bootstrap/NoopLogger";
+import * as DataModel from "../../Bootstrap/DataModel";
+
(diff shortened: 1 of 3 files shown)


Tests/Bootstrap/DiagnosticLogger.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+import { DiagnosticLogger } from "../../Bootstrap/DiagnosticLogger";
+
+describe("DiagnosticLogger", () => {
+    describe("#DiagnosticLogger.DefaultEnvelope", () => {
+        it("should have the correct version string", () => {
+            const version = require("../../../package.json").version;
+            assert.equal(DiagnosticLogger.DefaultEnvelope.properties.sdkVersion, version);
+        });
+    });
+});
```
xiao-lix, line 11: Does it necessary to add tests for other fields in properties bag? Looks like all of the fields are required.

xiao-lix (approved):

Tests/Bootstrap/DiagnosticLogger.spec.ts
markwolff: :+1: added test for rest of the properties bag

PR #683, opened 2020-09-24
Adding web snippet injection

hectorhdzg:


applicationinsights.ts
```diff
@@ -111,12 +115,13 @@ export function setup(setupString?: string) {
...
  */
 export function start() {
-    if(!!defaultClient) {
+    if (!!defaultClient) {
         _isStarted = true;
         _console.enable(_isConsole, _isConsoleLog);
         _exceptions.enable(_isExceptions);
         _performance.enable(_isPerformance);
         _heartbeat.enable(_isHeartBeat, defaultClient.config);
+        _webSnippet.enable(process.env.APPINSIGHTS_JAVASCRIPT_ENABLED == "true"); // Use environment variable by default
```
markwolff, line 124: This is the existing flag for legacy injection. Should we use a different one to prevent enabling both in windows?

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,128 @@
...
+    private _isInitialized: boolean;
+
+    constructor(client: TelemetryClient) {
+        if (!!WebSnippet.INSTANCE) {
+            throw new Error("Web snippet injection should be configured from the applicationInsights object");
+        }
+
+        WebSnippet.INSTANCE = this;
+        WebSnippet._aiUrl = "https://az416426.vo.msecnd.net/scripts/b/ai.2.min.js";
+        WebSnippet._snippet = '<script type="text/javascript">' +
```
MSNev, line 22: As this is a nodejs (server side) do we want to publish the snippet to the CDN so that this can fetch it and update dynamically? So that when we update the snippet we don't need to update this as well?

MSNev: Maybe we "publish" an autoinject_snippet.json or something, so that we could also change the default URL as well?

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,128 @@
...
...Child(e) : setTimeout(function () { l.getElementsByTagName(u)[0].parentNode.appendChild(e) }, y.ld || 0) } try { m.cookie = l.cookie } catch (p) { } function t(e) { for (; e.length;)!function (t) { m[t] = function () { var e = arguments; g || m.queue.push(function () { m[t].apply(m, e) }) } }(e.pop()) } var n = "track", r = "TrackPage", o = "TrackEvent"; t([n + "Event", n + "PageView", n + "Exception", n + "Trace", n + "DependencyData", n + "Metric", n + "PageViewPerformance", "start" + r, "stop" + r, "start" + o, "stop" + o, "addTelemetryInitializer", "setAuthenticatedUserContext", "clearAuthenticatedUserContext", "flush"]), m.SeverityLevel = { Verbose: 0, Information: 1, Warning: 2, Error: 3, Critical: 4 }; var s = (d.extensionConfig || {}).ApplicationInsightsAnalytics || {}; if (!0 !== d[C] && !0 !== s[C]) { method = "onerror", t(["_" + method]); var c = T[method]; T[method] = function (e, t, n, a, i) { var r = c && c(e, t, n, a, i); return !0 !== r && m["_" + method]({ message: e, url: t, lineNumber: n, columnNumber: a, error: i }), r }, d.autoExceptionInstrumented = !0 } return m }(y.cfg); (T[t] = n).queue && 0 === n.queue.length && n.trackPageView({}) }(window, document, {' +
```
markwolff, line 23: Move to another file. Either in this folder or in `Constants/`

applicationinsights.ts
markwolff, line 124: Move environment variable & parsing it to `Config`. Use `_isSnippetInjection` (same pattern as other autocollectors)

markwolff (commented):
Overall looks great. Few style comments

AutoCollection/WebSnippet.ts
markwolff: Agree with doing something to make updating it easier. My take on this would be to add a `npm run snippet:update` script which goes and fetches  the snippet from `ApplicationInsights-JS` and writes it to some checked in file in this repo, `Snippet.ts`

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,128 @@
...
+    private _isEnabled: boolean;
+    private _isInitialized: boolean;
+
+    constructor(client: TelemetryClient) {
+        if (!!WebSnippet.INSTANCE) {
+            throw new Error("Web snippet injection should be configured from the applicationInsights object");
+        }
+
+        WebSnippet.INSTANCE = this;
+        WebSnippet._aiUrl = "https://az416426.vo.msecnd.net/scripts/b/ai.2.min.js";
```
markwolff, line 21: I think it would be cool to have a config that would cause the SDK to fetch the unminified sdk and also uses the unminified version of the snippet.

applicationinsights.ts
hectorhdzg: Having another env variable would make it harder for customers to enable this, we need to update existing IIS URL rewrite before enabling this functionality to avoid multiple injections

AutoCollection/WebSnippet.ts
hectorhdzg: Unminified snippet is not even provided in Web SDK repo documentation, maybe loading min and unminified version of the SDK would be nice to help debugging not sure about the snippet itself, let me know what you think

markwolff: I agree it would be mainly for debugging purposes. Are the ones [here](https://github.com/microsoft/ApplicationInsights-JS/tree/master/AISKU/snippet) valid?

hectorhdzg: Added config option

applicationinsights.ts
hectorhdzg: Added new env variable to avoid double injection until IIS URL Rewrite feature is replaced

AutoCollection/WebSnippet.ts
hectorhdzg: Added snippet retrieval logic

AutoCollection/WebSnippet/fetchSnippet.js
```diff
@@ -0,0 +1,26 @@
+const path = require('path');
+const fs = require('fs');
+const https = require('https');
+
+let snippetUrl = 'https://raw.githubusercontent.com/microsoft/ApplicationInsights-JS/master/AISKU/snippet/snippet.js';
+let snippetMinUrl = 'https://raw.githubusercontent.com/microsoft/ApplicationInsights-JS/master/AISKU/snippet/snippet.min.js';
```
MSNev, line 6: nice! - Although this does mean that anything checked in -- even if not released will break this...

Perhaps we should consider publishing the snippet to the CDN as well and then change this to fetch from there.

AutoCollection/WebSnippet/snippet.js
```diff
@@ -0,0 +1,316 @@
...
+        // Only set if supplied or another name is defined to avoid polluting the global namespace
+        win[sdkInstanceName] = aiName;
+    }
+    var aisdk = win[aiName] || (function (aiConfig) {
+        var loadFailed = false;
+        var handled = false;
+        var appInsights = {
+            initialize: true,   // initialize sdk on download
+            queue: [],
+            sv: "4",            // Track the actual snippet version for reporting.
```
MSNev, line 29: This is v:"4" which has a bug when an exception occurs before the SDK is loaded -- grab the current one.

hectorhdzg:
Closing in favor of #891

PR #692, opened 2020-11-02
RDD: Correctly parse urls without protocols

### LINKED ISSUE #689: Wrong target reported when using AutoDependencyTracking and HTTP_PROXY environment variable
On our environment we're using request library together with HTTP_PROXY setting. All dependency requests that go through proxy have target of 443 (this are https requests).

As I've debugged through the code, it looks like [HttpDependencyParser.getDependencyTelemetry](https://github.com/microsoft/ApplicationInsights-node.js/blob/6a772adb58dcd1a1e5577dfb071efa7414326db2/AutoCollection/HttpDependencyParser.ts#L49) gets the url without protocol in such case. For example `host.example.com:443`. With such value, the parser treats host.example.com as protocol name, and 443 as host name. This gives useless results as `target` value and also - if we have multiple hosts - all of them are aggregated into one 443 value.

I'd like to have real host value in the target field, the same as it is in case, when no HTTP_PROXY is used.

markwolff:
Fix issue where dependency URLs without protocol prefixes are incorrectly parsed

Fixes #689

### DIFF
--- a/Tests/AutoCollection/HttpDependencyParser.tests.ts
+++ b/Tests/AutoCollection/HttpDependencyParser.tests.ts
@@ -44,6 +44,36 @@ describe("AutoCollection/HttpDependencyParser", () => {
             assert.equal(dependencyTelemetry.data, "http://bing.com:123/search");
             assert.equal(dependencyTelemetry.target, "bing.com:123 | abcdefg");
         });
+        
+        it("should return correct data for a URL without a protocol (https)", () => {
+            (<any>request)["method"] = "GET";
+            let parser = new HttpDependencyParser("a.bing.com:443/search", request);
+
+            response.statusCode = 200;
+            parser.onResponse(response);
+
+            let dependencyTelemetry = parser.getDependencyTelemetry();
+            assert.equal(dependencyTelemetry.dependencyTypeName, Contracts.RemoteDependencyDataConstants.TYPE_HTTP);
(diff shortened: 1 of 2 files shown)


Tests/AutoCollection/HttpDependencyParser.tests.ts
```diff
@@ -44,6 +44,36 @@ describe("AutoCollection/HttpDependencyParser", () => {
...
+        
+        it("should return correct data for a URL without a protocol (https)", () => {
+            (<any>request)["method"] = "GET";
+            let parser = new HttpDependencyParser("a.bing.com:443/search", request);
+
+            response.statusCode = 200;
+            parser.onResponse(response);
+
+            let dependencyTelemetry = parser.getDependencyTelemetry();
+            assert.equal(dependencyTelemetry.dependencyTypeName, Contracts.RemoteDependencyDataConstants.TYPE_HTTP);
```
xiao-lix, line 56: Should this type be "TYPE_HTTPS"?

xiao-lix (approved):

Tests/AutoCollection/HttpDependencyParser.tests.ts
markwolff: There is currently no distinction in this field between http/s. The other option for the field, `TYPE_AI` means that the RDD has an appId correlated with it.

PR #696, opened 2020-11-03
fix stackframe regex on scoped packages

### LINKED ISSUE #695: PII Leak: Stack parsing is broken in scoped npm packages
My package is @react-native-windows/cli . When an exception is thrown from inside of one of our tools, the EnvelopeFactory tries parsing the stack into a parsedStack : StackFrame[]
However the regex used doesn't work for packages (or file paths) that have an @ sign in the name:
https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/Library/EnvelopeFactory.ts#L306-L313

The regex is `/^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\@\n]+):([0-9]+):([0-9]+)(\)?)$/`
I'm throwing an exception from a lambda so there is no method name. The stack string looks like:
```
Error: solutionFile invalid. Error: Too many app solution files found, please specify in react-native.config.
    at E:\rnw\packages\@react-native-windows\cli\lib-commonjs\runWindows\utils\autolink.js:131:23
    at Array.forEach (<anonymous>)
    at Object.updateAutoLink [as func] (E:\rnw\packages\@react-native-windows\cli\lib-commonjs\runWindows\utils\autolink.js:125:24)
    at runWindowsInternal (E:\rnw\packages\@react-native-windows\cli\lib-commonjs\runWindows\runWindows.js:204:42)
    at Object.runWindows [as func] (E:\rnw\packages\@react-native-windows\cli\lib-commonjs\runWindows\runWindows.js:139:15)
    at Command.handleAction (E:\test63\node_modules\@react-native-community\cli\build\index.js:186:23)
    at Command.listener (E:\test63\node_modules\commander\index.js:315:8)
    at Command.emit (events.js:209:13)
    at Command.parseArgs (E:\test63\node_modules\commander\index.js:651:12)
    at Command.parse (E:\test63\node_modules\commander\index.js:474:21)
```

When parseStack tries the second line, it finds a match:
image

This ends up putting the path as part of the method name. 
This causes a potential PII leak since file path information is being exposed through a method name which is not expected to have any PII.

markwolff:
Updates regex to support stackframes from scoped packages, e.g. `@microsoft/applicationinsights-web`

Fixes #695

### DIFF
--- a/Library/EnvelopeFactory.ts
+++ b/Library/EnvelopeFactory.ts
@@ -356,7 +356,7 @@ class _StackFrame {
 
     // regex to match stack frames from ie/chrome/ff
     // methodName=$2, fileName=$4, lineNo=$5, column=$6
-    public static regex = /^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\n]+):([0-9]+):([0-9]+)(\)?)$/;
+    public static regex = /^(\s+at)?(.*?)(\@|\s\(|\s)([^\(\n]+):(\d+):(\d+)(\)?)$/;
     public static baseSize = 58; //'{"method":"","level":,"assembly":"","fileName":"","line":}'.length
     public sizeInBytes = 0;
     public level: number;
--- a/Library/EnvelopeFactory.ts
+++ b/Library/EnvelopeFactory.ts
@@ -356,7 +356,7 @@ class _StackFrame {
 
     // regex to match stack frames from ie/chrome/ff
     // methodName=$2, fileName=$4, lineNo=$5, column=$6
-    public static regex = /^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\@\n]+):([0-9]+):([0-9]+)(\)?)$/;
+    public static regex = /^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\n]+):([0-9]+):([0-9]+)(\)?)$/;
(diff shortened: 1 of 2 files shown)


asklar:
thanks for fixing this. One other case that the regex doesn't match:
```
    at Array.forEach (<anonymous>)
```

Library/EnvelopeFactory.ts
```diff
@@ -356,7 +356,7 @@ class _StackFrame {
 
     // regex to match stack frames from ie/chrome/ff
     // methodName=$2, fileName=$4, lineNo=$5, column=$6
-    public static regex = /^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\@\n]+):([0-9]+):([0-9]+)(\)?)$/;
+    public static regex = /^([\s]+at)?(.*?)(\@|\s\(|\s)([^\(\n]+):([0-9]+):([0-9]+)(\)?)$/;
```
asklar, line 359: This can be simplified:
```suggestion
    public static regex = /^(\s+at)?(.*?)(\@|\s\(|\s)([^\(\n]+):(\d+):(\d+)(\)?)$/;
```

asklar (approved):

asklar:
I'll file a separate bug to track the fact that `    at Array.forEach (<anonymous>)` isn't captured but I don't want to block the PR on that, we should get this one change out the door asap IMO.

PR #733, opened 2021-02-11
App Insights should use @azure/logger and send valid envelope contents for ingestion

### LINKED ISSUE #706: [Help] Structured logging
I try to keep my logs structured in a json file but then something fails in application Insights i get a console output in multiple lines.

__Something like this__

````
ApplicationInsights:Sender [
 'Ingestion endpoint could not be reached 5 consecutive times. There may be resulting telemetry loss. Most recent error:',
     at TCPConnectWrap.callbackTrampoline (internal/async_hooks.js:120:14) {
      syscall: 'connect',
      port: 443
      errno: 'ETIMEDOUT',
      Error: connect ETIMEDOUT 52.169.64.244:443
         at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1141:16)
    code: 'ETIMEDOUT',
    address: '52.169.64.244',
  }
]
````

But what i do with other logging is format it as json and output it to a single line

Is there a way to customise it in Application Insight framework?

### LINKED ISSUE #731: Use @azure/logger for logging
AppInsights should use @azure/logging like other azure modules for Node.js.

Related to #706

### LINKED ISSUE #732: Validate envelope after telemetryProcessors run
AppInsights should validate envelope contents (`tag`/`basedata.properties`) after telemetry processors run. Prior to the processors running, all values are cast to strings, this behavior doesn't carry over.

### PR DESCRIPTION
**Pull request by jmealo**
Resolves #731
Resolves #732
Resolves #706

### DIFF
--- a/AutoCollection/Exceptions.ts
+++ b/AutoCollection/Exceptions.ts
@@ -51,7 +51,7 @@ class AutoCollectExceptions {
                     this._client.flush({ isAppCrashing: true });
                     // only rethrow when we are the only listener
                     if (reThrow && name && process.listeners(name).length === 1) {
-                        console.error(error);
+                        Logging.error(error);
                         process.exit(1);
                     }
                 };
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -204,7 +204,7 @@ class AutoCollectHttpRequests {
      */
     public static trackRequest(client: TelemetryClient, telemetry: Contracts.NodeHttpRequestTelemetry, _requestParser?: HttpRequestParser) {
         if (!telemetry.request || !telemetry.response || !client) {
-            Logging.info("AutoCollectHttpRequests.trackRequest was called with invalid parameters: ", !telemetry.request, !telemetry.response, !client);
+            Logging.warn("AutoCollectHttpRequests.trackRequest was called with invalid parameters: ", !telemetry.request, !telemetry.response, !client);
             return;
         }
 
--- a/Bootstrap/Helpers/FileHelpers.ts
+++ b/Bootstrap/Helpers/FileHelpers.ts
@@ -3,6 +3,7 @@
 import * as path from "path";
 import * as fs from "fs";
 import * as os from "os";
+import Logging = require("../../Library/Logging")
 
 export const homedir = os.homedir ? os.homedir() : (process.env[(process.platform == "win32") ? "USERPROFILE" : "HOME"]);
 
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by jmealo** [comment]
@hectorhdzg: These tests pass locally; could I have a hand with the CI failures?

**Comment by hectorhdzg** [comment]
@jmealo thanks for submitting a PR to improve logging, main issue I can see here is regarding node.js support, I can see @azure/logger only runs in node>8 [link](https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/core/logger/package.json#L13) this is a potential breaking issue for customers using older versions of node and our SDK, I can see Azure logger code is not really big, is there a specific feature that you are looking for that is part of Azure logger and not to our internal one?

**Comment by jmealo** [comment]
@hectorhdzg: There is no internal logging. It just sends logs to console. We can add a fallback for that. If customers use any other Azure packages that's what they have to use to hook up their own structured logger.

**Comment by jmealo** [comment]
@hectorhdzg: @azure/logger should accept a standard logging instance compatible with structured loggers but it does not. This is the workaround that we are using right now:

```javascript
// From the @azure/logger docs:
// Note that AZURE_LOG_LEVEL, if set, takes precedence over DEBUG.
// Only use DEBUG without specifying AZURE_LOG_LEVEL or calling setLogLevel.
if (!DEBUG_WILDCARD && !process.env.AZURE_LOG_LEVEL) {
  setLogLevel(DEV ? 'verbose' : 'warning')
}

const azureLevelToPinoLevel = {
  verbose: 'debug',
  error: 'error',
  warning: 'warn',
  info: 'info'
}

// override logging to output to console.log (default location is stderr)
AzureLogger.log = (msg, ...args) => {
  let azureNamespace, logLevel

  // attempt to extract the azure namespace and log-level from the line prefix and map it to pino's
  if (msg.startsWith('azure:')) {
    const nsOffset = msg.indexOf(' ')
    const tokens = msg.slice(0, nsOffset).split(':')
    logLevel = azureLevelToPinoLevel[tokens.pop()]
    azureNamespace = tokens.join(':')
    msg = msg.slice(nsOffset + 1)
  }

  logger[logLevel || DEV ? 'info' : 'warn']({
    ns: azureNamespace || 'azure',
    azure: true,
    ...args
  }, msg)
}
```

I would make the argument that this change happens there so that it benefits all Azure packages.

I have made a separate PR #734 for the more pressing issue, which is #732.

jmealo:
@hectorhdzg: any update on when we can get these fixes merged?

hectorhdzg:
@jmealo only blocker here is node.js version support like I mentioned before, we are having internal talks regarding this topic, I will keep you informed here

jmealo:
> @jmealo only blocker here is node.js version support like I mentioned before, we are having internal talks regarding this topic, I will keep you informed here

Thanks, I agree that the @azure/logger module could/should do more.

Ideally it would expose a log4j compatible logger and each library that uses it would allow you to pass in your own logger as well.

Structured logging is becoming the norm on new projects so some standardizing could really help folks get the most out of the sdk.

hectorhdzg:
@jmealo we are dropping support for node.js before version 8 in next release, if you can resolve conflicts we can merge this one

jmealo:
@hectorhdzg: That's great news (for you and I!). I'll rebase this sometime this week.

PR #734, opened 2021-02-12
#732: Fix: Validate envelope after telemetry processors

### LINKED ISSUE #732: Validate envelope after telemetryProcessors run

jmealo:
Fixes #732 

When setting non-string values in telemetry processors the logging is hidden unless you use `.setInternalLogging()` and lower your `maxBatchSize` and `maxBatchIntervalMs` appropriately.

This code only runs when telemetry processors are present. (https://github.com/microsoft/ApplicationInsights-node.js/blob/c56f1933755c22999272f2299d37624426ddc125/Library/TelemetryClient.ts#L198)

I have tested this and it works to correct issues that occur when non-string values are added in telemetry processors. This makes the behavior consistent with adding custom properties at other points in the event lifetime.

### DIFF
--- a/Library/TelemetryClient.ts
+++ b/Library/TelemetryClient.ts
@@ -13,6 +13,7 @@ import Logging = require("./Logging");
 import FlushOptions = require("./FlushOptions");
 import EnvelopeFactory = require("./EnvelopeFactory");
 import QuickPulseStateManager = require("./QuickPulseStateManager");
+import {Tags} from "../Declarations/Contracts";
 
 /**
  * Application Insights telemetry client provides interface to track telemetry items, register telemetry initializers and
@@ -214,6 +215,15 @@ class TelemetryClient {
             } catch (error) {
                 accepted = true;
                 Logging.warn("One of telemetry processors failed, telemetry item will be sent.", error, envelope);
+            }
+        }
+
+        // Sanitize tags and properties after running telemetry processors
+        if (accepted) {
+            envelope.tags = Util.validateStringMap(envelope.data.baseData.properties) as Tags & Tags[];
+
+            if (envelope.data.baseData.properties) {
+                envelope.data.baseData.properties = Util.validateStringMap(envelope.data.baseData.properties);
             }
         }
 
--- a/Library/TelemetryClient.ts
+++ b/Library/TelemetryClient.ts
@@ -220,9 +220,10 @@ class TelemetryClient {
 
         // Sanitize tags and properties after running telemetry processors
         if (accepted) {
-            envelope.tags = Util.validateStringMap(envelope.data.baseData.properties) as Tags & Tags[];
-
-            if (envelope.data.baseData.properties) {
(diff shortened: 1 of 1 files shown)


jmealo:
@hectorhdzg: any update on when we can get these fixes merged?

Library/TelemetryClient.ts
```diff
@@ -217,6 +218,15 @@ class TelemetryClient {
             }
         }
 
+        // Sanitize tags and properties after running telemetry processors
+        if (accepted) {
+            envelope.tags = Util.validateStringMap(envelope.data.baseData.properties) as Tags & Tags[];
```
hectorhdzg, line 223: This sanitization will be triggered for all events even if no telemetry processors and tags were added, in the case of properties will be executed twice always, I wonder if this could be fixed having correct types in typescript preventing this to happen in the first place
https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/Declarations/Contracts/TelemetryTypes/EnvelopeTelemetry.ts#L64

jmealo: Hm... I intended for this to only run when Telemetry processors are attached. I wouldn't want others to take the performance hit.

Unfortunately, for those of us who don't use type script (or have dynamic inputs at runtime) is that this a non-string value fails silently and users are unlikely to have `setInternalLogging` on. Customer support isn't trained on this scenario and it took months to resolve and we figured it out in-house.

Rather than iterating over the map it might be cheaper to have the tags/properties objects use getters and setters that casts to strings passed to the telemetry processors.

Another place where this type of silent conversion happens is `process.env`; so it's not unheard of.

jmealo: @hectorhdzg: it looks like `runTelemetry Processors` exits early if none are attached?

hectorhdzg: You are right, I got confused github file changes view didn't help this time

Library/TelemetryClient.ts
```diff
@@ -217,6 +218,15 @@ class TelemetryClient {
...
+
+            if (envelope.data.baseData.properties) {
```
hectorhdzg, line 225: Because this code is executed after telemetry processors are executed we need to be extra careful with undefined references adding checks for envelope.data, and envelope.data.baseData

Library/TelemetryClient.ts
hectorhdzg, line 223: You need to sanitize envelope.tags not baseData properties

jmealo: Good catch! I'll fix these.

jmealo (commented):
@hectorhdzg: Good catch, I made the requested changes.

hectorhdzg (approved):

PR #737, opened 2021-03-03
Roles support, Endpoint redirects and ping interval hints for Live Metrics

rambhatt-msft:


Library/QuickPulseStateManager.ts
```diff
@@ -29,6 +29,8 @@ class QuickPulseStateManager {
     private _metrics: {[name: string]: Contracts.MetricQuickPulse} = {};
     private _documents: Contracts.DocumentQuickPulse[] = [];
     private _collectors: {enable: (enable: boolean) => void}[] = [];
+    private _rediectedHost: string = null;
```
OsvaldoRosado, line 32: Just curious - why this vs updating the host in config? Is the idea that we don't want to expose the redirection when it happens?

rambhatt-msft: We do not want to make this change persistent, if the agent is restarted, then it should start polling the endpoint it was originally configured to. It will be redirected again

Library/QuickPulseStateManager.ts
```diff
@@ -29,6 +29,8 @@ class QuickPulseStateManager {
...
+    private _poliingIntervalHint: number = -1;
```
ramthi, line 33: NIT: typos.  _redirectedHost and _pollingInterval

Library/QuickPulseStateManager.ts
```diff
@@ -145,24 +148,33 @@ class QuickPulseStateManager {
...
+    private _quickPulseDone(shouldPOST?: boolean, res?: http.IncomingMessage, 
+            redirectedHost?: string, pollingIntervalHint?: number): void {
         if (shouldPOST != undefined) {
             if (this._isCollectingData !== shouldPOST) {
                 Logging.info("Live Metrics sending data", shouldPOST);
                 this.enableCollectors(shouldPOST);
             }
             this._isCollectingData = shouldPOST;
 
+            if (redirectedHost && redirectedHost.length > 0) {
```
ramthi, line 170: Please do a console.info here to aid debugging.

Library/QuickPulseStateManager.ts
rambhatt-msft: Ah, thanks for pointing out.

Library/QuickPulseStateManager.ts
rambhatt-msft: Sure. Done!

Library/QuickPulseStateManager.ts
```diff
@@ -130,7 +132,8 @@ class QuickPulseStateManager {
             this._ping(envelope);
         }
 
-        let currentTimeout = this._isCollectingData ? QuickPulseStateManager.POST_INTERVAL : QuickPulseStateManager.PING_INTERVAL;
+        let pingInterval = this._pollingIntervalHint > 0 ? this._pollingIntervalHint : QuickPulseStateManager.PING_INTERVAL;
```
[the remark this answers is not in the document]
rambhatt-msft: No, the endpoint will not return negative values, and the SDK must not go back to the original ping interval unless it is restarted.

PR #745, opened 2021-03-25
Make native metrics handle creation singleton

### LINKED ISSUE #743: Native-metrics collection handle creation is not singleton - results in increased logging frequency
### Background:

I am using **applicationinsights** and **applicationinsights-native-metrics** packages for tracking metrics & logs for our azure functions (developed in Node.js). By default, the native metrics logs has to be emitted once a min. But I noticed that the native-metrics are logged more than once a min. In fact, the metrics were logged at a much higher frequency that resulted in the log/metric data volume in azure application-insights exceeded 800GB last month. We ended up paying more as the current plan allowed only 800GB of logs per month.

### Issue analysis

While analyzing the issue, I found out that our code is calling **require('applicationinsights').setup().start()** method at every function/API invocation. So if API/function was invoked 10 times, the frequency increased to 10 times/min. Similarly, if it was invoked 1000 times,  the frequency of native-metrics logging increased to 1000 times/min. Over the course of time, this build up resulted in huge data consumption that caused the 800+GB log data consumption per month.
 
### Suggestion

Even though I fixed the issue in our code, I wanted to discuss and see if the handle creation can be made singleton to avoid these kind of issues. When looking at the native-metrics interval creation, I found out that the Application Insights Native-metrics collection handle creation is not singleton. 

Please refer the following code block in
**AutoCollection/NativePerformance.ts** lines 80-93
```
       // Enable the emitter if we were able to construct one
        if (this._isEnabled && AutoCollectNativePerformance._emitter) {
            // enable self
            AutoCollectNativePerformance._emitter.enable(true, collectionInterval);
            this._handle = setInterval(() => this._trackNativeMetrics(), collectionInterval);     // This can be made singleton
            this._handle.unref();
        } else if (AutoCollectNativePerformance._emitter) {
            // disable self
            AutoCollectNativePerformance._emitter.enable(false);
            if (this._handle) {
                clearInterval(this._handle);
                this._handle = undefined;
            }
        }
```

When enabling the emitter, if the handle is already created, we should avoid calling the setInterval method again to create a new timer. This results in creation of multiple timers and the logging frequency increases. 

### Suggested Solution

We can add a check before and handle creation to see if there's a handle already been created. This will avoid creating multiple timers and the logging frequency will never exceed once a min.

```
      // Enable the emitter if we were able to construct one
        if (this._isEnabled && AutoCollectNativePerformance._emitter) {
            // enable self
            AutoCollectNativePerformance._emitter.enable(true, collectionInterval);
            if (!this._handle) {
                this._handle = setInterval(() => this._trackNativeMetrics(), collectionInterval);
                this._handle.unref();
            }
        } else if (AutoCollectNativePerformance._emitter) {
            // disable self
            AutoCollectNativePerformance._emitter.enable(false);
            if (this._handle) {
                clearInterval(this._handle);
                this._handle = undefined;
            }
        }
```

I have tested this code by patching the diff using postinstall in package.json and it worked. Please let me know if this can be fixed this way. If not, please share your thoughts on why it's written the way it is right now.

sriramprabhu:
Fixes #743

### DIFF
--- a/AutoCollection/NativePerformance.ts
+++ b/AutoCollection/NativePerformance.ts
@@ -81,8 +81,10 @@ export class AutoCollectNativePerformance {
         if (this._isEnabled && AutoCollectNativePerformance._emitter) {
             // enable self
             AutoCollectNativePerformance._emitter.enable(true, collectionInterval);
-            this._handle = setInterval(() => this._trackNativeMetrics(), collectionInterval);
-            this._handle.unref();
+            if (!this._handle) {
+                this._handle = setInterval(() => this._trackNativeMetrics(), collectionInterval);
+                this._handle.unref();
+            }
         } else if (AutoCollectNativePerformance._emitter) {
             // disable self
             AutoCollectNativePerformance._emitter.enable(false);
--- a/Tests/AutoCollection/NativePerformance.tests.ts
+++ b/Tests/AutoCollection/NativePerformance.tests.ts
@@ -47,6 +47,19 @@ describe("AutoCollection/NativePerformance", () => {
                 assert.ok(disposeSpy.calledOnce, "dispose is called when second instance is constructed");
             });
 
+            it("Calling enable multiple times shoud not create multiple timers", () => {
+                var client = new TelemetryClient("1aa11111-bbbb-1ccc-8ddd-eeeeffff3333");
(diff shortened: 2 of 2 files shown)


hectorhdzg (approved):
Thanks for your contribution

PR #746, opened 2021-03-30
Add AAD support

hectorhdzg:


kryalama:
Draft looks good to me 👍

TimothyMothra:
I think we are leaning away from including those values in the Connection String.

Library/AuthHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+            timeBetweenRefreshAttemptsInMs
+        );
+    }
+
+    /**
+       * Applies the Bearer token to the request through the Authorization header.
+       */
+    public async addAuthorizationHeader(requestOptions: http.RequestOptions | https.RequestOptions): Promise<void> {
+        const token = await this._getToken({});
+        requestOptions.headers[azureCore.Constants.HeaderConstants.AUTHORIZATION] = `Bearer ${token}`;
```
MSNev, line 76: Does the header value need to be encoded?

Library/AuthHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private async _getToken(options: azureCore.GetTokenOptions): Promise<string | undefined> {
+        // We reset the cached token some before it expires,
+        // after that point, we retry the refresh of the token only if the token refresher is ready.
+        let token = this._tokenCache.getCachedToken();
+        if (!token && this._tokenRefresher.isReady()) {
+            token = await this._tokenRefresher.refresh(options);
```
MSNev, line 84: What happens if this fails, does the setCachedToken() handle null / undefined or invalid token value?

Library/AuthHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+
+class AuthHandler {
+
+    private _tokenCache: azureCore.AccessTokenCache;
+    private _tokenRefresher: azureCore.AccessTokenRefresher;
+
+    constructor(config: Config) {
+        let credential: azureCore.TokenCredential;
+
+        let authOptions = new AuthorizationOptions();
```
ramthi, line 47: We are using what comes from config directly. How about some basic validation?

Library/AuthHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+       */
+    public async addAuthorizationHeader(requestOptions: http.RequestOptions | https.RequestOptions): Promise<void> {
+        const token = await this._getToken({});
+        requestOptions.headers[azureCore.Constants.HeaderConstants.AUTHORIZATION] = `Bearer ${token}`;
+    }
+
+    private async _getToken(options: azureCore.GetTokenOptions): Promise<string | undefined> {
+        // We reset the cached token some before it expires,
+        // after that point, we retry the refresh of the token only if the token refresher is ready.
+        let token = this._tokenCache.getCachedToken();
```
ramthi, line 82: NIT: name this cachedToken

Library/AuthHandler.ts
ramthi: +1

Library/Config.ts
```diff
@@ -92,6 +101,16 @@ class Config {
             "*.core.eaglex.ic.gov"
         ];
 
+        this.isAuthRequired = csCode.authorization && csCode.authorization === "aad";
```
ramthi, line 104: There was a discussion on the Spec doc.  suggestion is to use AAD instead of aad.  Unless the connectionstring parser tolowers everything you need to update this.

Library/AuthHandler.ts
```diff
@@ -0,0 +1,91 @@
...
+    public isSecretAuth(): boolean {
+        return this.appId != "" && this.tenantId != "" && this.appKey != "";
+    }
+
+    public isManagedIdentityAuth(): boolean {
+        return this.appId != "";
+    }
+}
+
+class AuthHandler {
```
ramthi, line 39: Can you pl. fully spell this out as AuthorizationHandler ?

Library/QuickPulseSender.ts
```diff
@@ -84,6 +88,19 @@ class QuickPulseSender {
             additionalHeaders.forEach(header => options.headers[header.name] = header.value);
         }
 
+        if (this._authorizationHandler) {
+            try {
+                // Add bearer token
+                await this._authorizationHandler.addAuthorizationHeader(options);
+            }
+            catch (authError) {
```
ramthi, line 96: Will exceptions from the async methods dep in the stack like "_tokenRefresher.refresh(..)" bubbleup properly to report a meaningful error?  We may have to do some logging in those methods to ensure we dont lose critical diagnostic info.

Library/QuickPulseSender.ts
```diff
@@ -84,6 +88,19 @@ class QuickPulseSender {
...
+                let notice = `Failed to get AAD bearer token for the Application. Error:`;
+                Logging.info(QuickPulseSender.TAG, notice, authError);
```
ramthi, line 98: Shouldnt this be Logging.warn?  I think its critical enough to be a warning ...there will be data loss.

Library/Sender.ts
```diff
@@ -146,6 +165,8 @@ class Sender {
                             }
                             // store to disk in case of burst throttling
                         } else if (
+                            res.statusCode === 401 || // Unauthorized
+                            res.statusCode === 403 || // Forbidden
```
ramthi, line 169: So there is no retry when we hit any AAD auth error?  Is that what other langs are doing?  I thought we were going to retry!

ramthi (changes requested):

Library/Sender.ts
hectorhdzg: This specific code is the retry logic, envelopes would be put into disk, are you expecting something else?

Library/AuthHandler.ts
hectorhdzg: The token could be null or undefined, will add the check before adding the header

https://github.com/Azure/azure-sdk-for-js/blob/master/sdk/core/core-http/src/credentials/accessTokenCache.ts#L50

Library/QuickPulseSender.ts
hectorhdzg: Exceptions must bubble up, are you worried about a specific path?, I prefer to have single place handling the errors but we can add more handling if specific error paths are identified and logging could be useful

Library/QuickPulseSender.ts
hectorhdzg: This is for quickpulse calls, will apply to both ping and post requests so warnings could get super noisy, data loss is not critical here

Library/AuthHandler.ts
hectorhdzg: Not using config anymore, user need to provide TokenCredential

Library/AuthHandler.ts
hectorhdzg: This is the way other SDKs are adding the header, code is not ready on Breeze side so this may change in the future

Declarations/Contracts/Constants.ts
```diff
@@ -37,4 +37,4 @@ export interface ConnectionString {
     // type ConnectionString = { [key in ConnectionStringKey]?: string }
 }
 
-export type ConnectionStringKey = "authorization" | "instrumentationkey" | "ingestionendpoint" | "liveendpoint" | "location" | "endpointsuffix";
+export type ConnectionStringKey = "authorization" | "instrumentationkey" | "ingestionendpoint" | "liveendpoint" | "location"| "endpointsuffix";
```
kryalama, line 40: we may not need this "authorization" config in connection string anymore.

Library/AuthorizationHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+import azureCore = require("@azure/core-http");
+
+/**
+ * The automated token refresh will only start to happen at the
+ * expiration date minus the value of timeBetweenRefreshAttemptsInMs,
+ * which is by default 30 seconds.
+ */
+const timeBetweenRefreshAttemptsInMs = 30000;
+
+const applicationInsightsResource = "https://monitor.azure.com";
```
kryalama, line 12: since user is providing the token credential, is it a good idea to have them a option to provide "scope" too? If they donot provide the scope, we can use this default scope.

hectorhdzg: Looking at other Azure SDKs implementations, scopes are never exposed and I don't believe we want the user to control which resource to authenticate, this is 100% managed by Breeze, if scope change we will need to update the SDK

Library/AuthorizationHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (token) {
+            requestOptions.headers[azureCore.Constants.HeaderConstants.AUTHORIZATION] = `Bearer ${token}`;
+        }
+    }
+
+    private async _getToken(options: azureCore.GetTokenOptions): Promise<string | undefined> {
+        // We reset the cached token some before it expires,
+        // after that point, we retry the refresh of the token only if the token refresher is ready.
+        let cachedToken = this._tokenCache.getCachedToken();
+        if (!cachedToken && this._tokenRefresher.isReady()) {
```
lzchen, line 44: Shouldn't this be `OR`?

Library/AuthorizationHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+ * The automated token refresh will only start to happen at the
+ * expiration date minus the value of timeBetweenRefreshAttemptsInMs,
+ * which is by default 30 seconds.
+ */
+const timeBetweenRefreshAttemptsInMs = 30000;
+
+const applicationInsightsResource = "https://monitor.azure.com";
+
+
+class AuthorizationHandler {
```
lzchen, line 15: Should this be private?

Library/Sender.ts
```diff
@@ -97,6 +100,24 @@ class Sender {
             }
         };
 
+        let authHandler = this._getAuthorizationHandler ? this._getAuthorizationHandler() : null;
+        if (authHandler) {
+            try {
+                // Add bearer token
+                await authHandler.addAuthorizationHeader(options);
+            }
+            catch (authError) {
```
lzchen, line 109: I believe we want to do different things (retry, drop, etc.) based off the type of exception that is thrown when getting the token. [Python](https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/identity/azure-identity/README.md#error-handling) throws these types of exceptions (I believe Java is similar). For `CredentialUnavailableError ` we probably do not want to retry. For `ClientAuthenticationError` we should investigate which cases make sense to retry.

lzchen:
I think an example file of passing in a TokenCredential would be good.

Library/Sender.ts
```diff
@@ -136,6 +157,8 @@ class Sender {
                             }
                             // store to disk in case of burst throttling
                         } else if (
+                            res.statusCode === 401 || // Unauthorized
+                            res.statusCode === 403 || // Forbidden
```
lzchen, line 161: When are the cases where 403 is returned?

hectorhdzg: This is just a place holder, not sure which responses will be send by Breeze most likely will drop 403 check

Library/Sender.ts
hectorhdzg: If we do not retry telemetry will be lost on the spot in that case, even if customer configuration is wrong we should try to store the telemetry in disk, what about AAD outages, maybe CredentialUnavailableError is triggered in that case?

Library/Sender.ts
ramthi: Yes, just looked at Python PR.  Leighton is retrying if we hit 401. @lzchen can you comment?

lzchen: Yes 401 is a retry scenario.

MSNev (approved):

MSNev (approved):

ramthi (approved):

PR #747, opened 2021-04-01
Adding Breeze endpoint redirect support

hectorhdzg:


Library/Sender.ts
```diff
@@ -156,6 +146,27 @@ class Sender {
                             this._storeToDisk(payload);
                         }
                     }
+                    // Redirect handling
+                    if (res.statusCode === 301 || // Moved Permanently
```
MSNev, line 150: Krishna, found with Application Insights that by default XHR will follow redirects (and there is no way to disable), for fetch the default is also to "follow" redirects (but you can change the default with the "redirect" option).

In both of the default cases you won't get the redirect status code.

For XHR (may be the same for fetch), he had to "review" the response URL to see if it was different it "infer" that a redirect occurred and so all new requests get redirected to the new endpoint.

MSNev (approved):

Library/Sender.ts
hectorhdzg: Node.js http native libraries will not handle redirects automatically so no need to worry about this scenario here

Library/Sender.ts
```diff
@@ -205,8 +216,19 @@ class Sender {
...
     }
 
+    private _handleRedirect(location: string) {
+        this._numConsecutiveRedirects++;
+        // To prevent circular redirects
+        if (this._numConsecutiveRedirects < 10) {
+            this._redirectedHost = location;
+        }
+        else {
+            //TODO: Add ?redirect=false to avoid further redirection when 2.1 endpoint available
```
kryalama, line 226: still following up with Oleg on this if Breeze will support ?redirect=false.  Based on our last conversation, looks like Stamp Specific endpoints donot support ?redirec=false with AAD authentication enabled.

kryalama (approved):

PR #752, opened 2021-04-14
Adding 206 response handling from Breeze 

hectorhdzg:
Adding task to clean old files

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -7,8 +7,10 @@ import child_process = require("child_process");
 
 import Logging = require("./Logging");
 import Config = require("./Config")
+import Contracts = require("../Declarations/Contracts");
 import AutoCollectHttpDependencies = require("../AutoCollection/HttpDependencies");
 import Util = require("./Util");
+import { BreezeError } from "../Declarations/Contracts";
 
 class Sender {
     private static TAG = "Sender";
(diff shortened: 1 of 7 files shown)


Library/Sender.ts
```diff
@@ -83,148 +92,185 @@ class Sender {
...
-                options.headers["Content-Encoding"] = "gzip";
-                options.headers["Content-Length"] = buffer.length.toString();
-            }
+            let batch: string = "";
 
-            Logging.info(Sender.TAG, options);
+            envelopes.forEach(envelope => {
+                var payload: string = this._stringify(envelope);
+                if (typeof payload !== "string") {
+                    return;
```
MSNev, line 124: If there is a single "invalid" envelope that can't be stringified, (say in the middle of the payload), won't this case only half the batch to be sent?
Not that I can think of a normal case where JSON.stringify() would fail... Just looking for corner cases.

Library/Sender.ts
```diff
@@ -83,148 +92,185 @@ class Sender {
...
-                    Logging.info(Sender.TAG, notice, error)
-                }
-                this._onErrorHelper(error);
+                    this._onErrorHelper(error);
+
+                    if (typeof callback === "function") {
+                        var errorMessage = "error sending telemetry";
+                        if (error && (typeof error.toString === "function")) {
+                            errorMessage = error.toString();
+                        }
```
MSNev, line 235: On "trick" I'm using now is that to just use JSON.strinfigy() on the error -- so you still get some context.

Library/Sender.ts
```diff
@@ -447,12 +490,12 @@ class Sender {
...
                 //would require an external dependency
                 var fileName = new Date().getTime() + ".ai.json";
-                var fileFullPath = path.join(directory, fileName);
+                var fileFullPath = path.join(this._tempDir, fileName);
 
                 // Mode 600 is w/r for creator and no read access for others (only applies on *nix)
                 // For Windows, ACL rules are applied to the entire directory (see logic in _confirmDirExists and _applyACLRules)
                 Logging.info(Sender.TAG, "saving data to disk at: " + fileFullPath);
-                fs.writeFile(fileFullPath, payload, { mode: 0o600 }, (error) => this._onErrorHelper(error));
+                fs.writeFile(fileFullPath, JSON.stringify(envelopes), { mode: 0o600 }, (error) => this._onErrorHelper(error));
```
MSNev, line 498: 2 questions
- I'm assuming that we don't need to worry about back compat for loading previous version persisted items
- Based on the try/catch stringify check when sending requests do we want to do that here as well so entire batches don't get dropped.

MSNev: Yours also explicitly converting the string to utf8 prior to reading -- should ensure that it's written as such also.

hectorhdzg: We will not support old AppInsights files because the telemetry was stored differently, this change will be part of major version update of the SDK, older files will be automatically cleaned after expiration is reached

Library/Sender.ts
hectorhdzg: We want to drop the individual "wrong" event and continue with others, I'm trying to keep previous logic, this change is mainly focused on having the envelopes available in this class instead of payload like before, allowing us to filter them for retriable response codes.

MSNev (approved):

PR #778, opened 2021-06-15
Negative preaggregated metrics fix

### LINKED ISSUE #776: After upgrading to v2, negative server response times are recorded
With v1 of the library, we are seeing consistent response times in the 10-20ms range. After upgrading to v2, server response times as measured in app insights go negative and are wildly variable. I've attached a screenshot showing the negative response times after upgrading. Testing the server and looking at response times captured by the Azure Web App show no actual change in the response times of the server.

Screen Shot 2021-06-15 at 10 18 34 AM

After downgrading to v1, response times go back to the normal expected range. Node v14.15.1 and Express. We are using the promise based router for express (express-promise-router) which I could imagine might be affecting the measurement of response duration.

hectorhdzg:
Last interval execution time need to be part of specific pre aggregated counter instead of global one to avoid miscalculations generating negative numbers

Fixes #776

### DIFF
--- a/AutoCollection/PreAggregatedMetrics.ts
+++ b/AutoCollection/PreAggregatedMetrics.ts
@@ -23,8 +23,6 @@ class AutoCollectPreAggregatedMetrics {
     private _handle: NodeJS.Timer;
     private _isEnabled: boolean;
     private _isInitialized: boolean;
-    private _lastIntervalRequestExecutionTime: number = 0; // the sum of durations which took place during from app start until last interval
-    private _lastIntervalDependencyExecutionTime: number = 0;
 
     private static _dependencyCountersCollection: Array<AggregatedMetricCounter>;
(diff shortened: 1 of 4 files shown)


MSNev (approved):

AutoCollection/PreAggregatedMetrics.ts
```diff
@@ -172,8 +170,8 @@ class AutoCollectPreAggregatedMetrics {
             currentCounter.time = +new Date;
             var intervalRequests = (currentCounter.totalCount - currentCounter.lastTotalCount) || 0;
             var elapsedMs = currentCounter.time - currentCounter.lastTime;
```
ramthi, line 172: This method and countDependency/countRequet methods have 95% similar code and perhaps could be refactored to remove duplication, in future.

PR #792, opened 2021-07-14
Adding StatsBeat implementation

hectorhdzg:


AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+    public enable(isEnabled: boolean) {
+        this._isEnabled = isEnabled;
+        if (this._isEnabled && !this._isInitialized) {
+            this._getCustomProperties();
+            this._isInitialized = true;
+        }
+        if (isEnabled) {
+            if (!this._handle) {
+                this._handle = setInterval(() => {
+                    this.trackStatsBeatMetrics()
```
heyams, line 59: trackStatsbeatMetrics? let's treat statsbeat as one word ;)

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    public isEnabled() {
+        return this._isEnabled;
+    }
+
+    public setCodelessAttach() {
+        this._attach = Constants.StatsBeatAttach.codeless;
+    }
+
+    public addFeature(feature: Constants.StatsBeatFeature) {
```
heyams, line 83: StatsbeatFeature?

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+import AzureVirtualMachine = require("../Library/AzureVirtualMachine");
+import Config = require("../Library/Config");
+import Context = require("../Library/Context");
+import { MetricTelemetry } from "../Declarations/Contracts";
+
+const STATS_BEAT_LANGUAGE = "node";
+const STATS_BEAT_IKEY = "";
+
+
+export class StatsBeat {
```
heyams, line 13: Statsbeat?  let's rename everywhere that has StatsBeat.

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    // Counters
+    private _lastRequests: { totalRequestCount: number; time: number; };
+    private _totalRequestCount: number = 0;
+    private _totalSuccesfulRequestCount: number = 0;
+    private _totalFailedRequestCount: number = 0;
+    private _retryCount: number = 0;
+    private _exceptionCount: number = 0;
```
heyams, line 21: how does node.js handle multiple threading? these availables are needed to be thread-safe?

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+    private _throttleCount: number = 0;
+    private _intervalRequestExecutionTime: number = 0;
+    private _lastIntervalRequestExecutionTime: number = 0;
+
+    private _collectionInterval: number = 900000; // 15 minutes
+    private _client: TelemetryClient;
+    private _handle: NodeJS.Timer | null;
+    private _isEnabled: boolean;
```
heyams, line 29: enabled?

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    public countRetry() {
+        if (!this.isEnabled()) {
+            return;
+        }
+        this._retryCount++;
+    }
+
+    public trackStatsBeatMetrics() {
```
heyams, line 134: trackStatsbeatMetrics?

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+        this._trackRequestDuration();
+        this._trackRequestsCount();
+    }
+
+    private _trackRequestDuration() {
```
heyams, line 139: traceRequestDurationAvg? it for average of the request duration in each 15 mins period.

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+        var lastRequests = this._lastRequests;
+        var requests = {
+            totalRequestCount: this._totalRequestCount,
+            time: +new Date
+        };
+        var intervalRequests = (requests.totalRequestCount - lastRequests.totalRequestCount) || 0;
+        var elapsedMs = requests.time - lastRequests.time;
+        var averageRequestExecutionTime = ((this._intervalRequestExecutionTime - this._lastIntervalRequestExecutionTime) / intervalRequests) || 0; // default to 0 in case no requests in this interval
```
heyams, line 147: not sure i fully understand this..  can you do this: 
totalRequestDuration/totalRequestcount for each 15 mins?

heyams: you might need to track total request duration.

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+    public addFeature(feature: Constants.StatsBeatFeature) {
+        this._features |= feature;
+    }
+
+    public removeFeature(feature: Constants.StatsBeatFeature) {
+        this._features &= ~feature;
+    }
+
+    public addInstrumentation(instrumentation: Constants.StatsBeatInstrumentation) {
+        this._instrumentations |= instrumentation;
```
heyams, line 92: i would like to see some test for this to fully understand how you convert a list of strings to number (64-bit)?

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+                "attach": this._attach,
+                "instrumentation": this._instrumentations,
+                "feature": this._features,
+            }
+        };
+        this._client.trackMetric(statsBeat);
+    }
+
+    private _getCustomProperties() {
+        this._language = STATS_BEAT_LANGUAGE;
```
heyams, line 198: STATSBEAT_LANGUAGE

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+        this._ciIkey = this._config.instrumentationKey;
+        this._sdkVersion = Context.sdkVersion; // "node" or "node-nativeperf"
+        this._os = os.type();
+        this._runtimeVersion = process.version;
+        // Check resource provider
+        if (process.env.WEBSITE_SITE_NAME) { // Web apps
+            this._resourceProvider = Constants.StatsBeatResourceProvider.appsvc;
```
heyams, line 205: StatsbeatResourceProvider

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,218 @@
...
+        } else if (process.env.FUNCTIONS_WORKER_RUNTIME) { // Function apps
+            this._resourceProvider = Constants.StatsBeatResourceProvider.function;
+        } else if (this._config) {
+            let vm = new AzureVirtualMachine(this._config);
```
heyams, line 209: how do you schedule this to run every 15 mins?   Look like it's one time call?

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -4,6 +4,7 @@ import { Span } from "@opentelemetry/tracing";
 import { SpanKind } from "@opentelemetry/api";
 
 import TelemetryClient = require("../../Library/TelemetryClient");
+import { StatsBeatInstrumentation } from "../../Declarations/Constants";
```
heyams, line 7: StatsbeatInstrumentation

Declarations/Constants.ts
```diff
@@ -141,4 +141,45 @@ export const DependencyTypeName = {
     InProc: "InProc",
 }
 
-export const HeartBeatMetricName = "HeartBeat";
\ No newline at end of file
+export const HeartBeatMetricName = "HeartBeat";
+
+export const StatsBeatMetricName = "StatsBeat";
```
heyams, line 146: Statsbeat?  it's not metric name but rather telemetry name.

Declarations/Constants.ts
```diff
@@ -141,4 +141,45 @@ export const DependencyTypeName = {
...
+    EXCEPTION_COUNT: "Exception Count",
+}
+
+export enum StatsBeatFeature {
+    NONE = 0,
+    DISK_RETRY = 1,
+    AAD_HANDLING = 2,
+}
+
+export enum StatsBeatInstrumentation {
```
heyams, line 175: i would like to see how you can decode the long back to string on the kusto size? i guess you will another map for the string -> 2 to the power of x.
Please add a test for encoding and decoding so that i have an rough idea how it looks like in kusto.

heyams: you might also consider when it's exceeding 64 bits, a new enum will be required? java has BitSet which can hold more than 64 bits.

Library/AzureVirtualMachine.ts
```diff
@@ -0,0 +1,61 @@
...
+import Util = require("./Util");
+import AutoCollectHttpDependencies = require("../AutoCollection/HttpDependencies");
+
+const AIMS_URI = "http://169.254.169.254/metadata/instance/compute";
+const AIMS_API_VERSION = "api-version=2017-12-01";
+const AIMS_FORMAT = "format=json";
+
+class AzureVirtualMachine {
+
+    public isVM: boolean;
```
heyams, line 12: how do you keep this running every 15 mins when it's from VM?

Library/AzureVirtualMachine.ts
```diff
@@ -0,0 +1,61 @@
...
+        const requestOptions = {
+            method: 'GET',
+            [AutoCollectHttpDependencies.disableCollectionRequestOption]: true,
+            headers: {
+                "Metadata": "True",
+            }
+        };
+
+        const req = Util.makeRequest(config, metadataRequestUrl, requestOptions, (res) => {
+            if (res.statusCode === 200) {
```
heyams, line 40: you might want to log when it's not reachable? non-vm will lead to socketexception.  in that case, it's a good indicator that it's not from a vm.. you can shutdown the scheduler if applicable.

heyams:
how does the config look like for statsbeat? 
this is for Java: 

` "internal": {
	"statsbeat": {
		"instrumentationKey": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx", 
		"endpoint": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
		"intervalSeconds": 30 , 
		"featureIntervalSeconds": 30
	}
  },`

AutoCollection/StatsBeat.ts
```diff
@@ -191,7 +198,9 @@ export class StatsBeat {
                 "feature": this._features,
             }
         };
-        this._client.trackMetric(statsBeat);
+        let envelope = EnvelopeFactory.createEnvelope(statsBeat, Contracts.TelemetryType.Metric, null, null, this._statsBeatConfig);
+        envelope.name = Constants.StatsBeatMetricName;
+        this._sender.send([envelope]);
```
MSNev, line 203: As send is async (returning a promise) we should also handle this as async -- otherwise if the promise rejects it will cause a promise unhandledrejection to be thrown / caught by something -- not sure if this would be terminal for node environment

Library/AzureVirtualMachine.ts
hectorhdzg: That must be logged already in line 68 for latest commit

hectorhdzg:
@heyams for Statsbeat here there is no configuration file available, everything is set as attributes or constants in Statsbeat class

Library/AzureVirtualMachine.ts
hectorhdzg: I assumed VM info is never updated during runtime, I will update so meta data request is made each time Statsbeat or Heartbeat need the info

AutoCollection/StatsBeat.ts
hectorhdzg: This code must be handling all exceptions already so in theory it must not throw any error making the promise to be rejected, even so I added async and try and catch to be sure nothing goes wrong

AutoCollection/StatsBeat.ts
```diff
@@ -0,0 +1,236 @@
...
+        }
+        this._retryCount++;
+    }
+
+    public trackStatsbeatMetrics() {
+        try {
+            this._trackRequestDuration();
+            this._trackRequestsCount();
+        }
+        catch (error) {
```
MSNev, line 149: try / catch doesn't work for promise (async) request responses as when the promise rejects the exception is thrown to an event handler and not this code -- because the JS execution has long since moved on.

MSNev: You would really have to do something like this
```
this._trackRequestDuration().then(() => { }, () => {});
```

Which would effectively ignore the "success" and "reject" responses.

Declarations/Constants.ts
hectorhdzg: We have around 8 instrumentations currently so we must be good, I believe signed number used by bitwise operations in JS is only 32 bits

AutoCollection/StatsBeat.ts
hectorhdzg: No need to worry about multiple threads in Javascript, this values must be good and counts will be associated by specific Telemetry Clients

AutoCollection/StatsBeat.ts
hectorhdzg: There is no list of strings, I use an enumerator describing the instrumentation and a number as value, number is always a power of 2.

AutoCollection/StatsBeat.ts
hectorhdzg: this is a carbon copy of how we calculate average duration for customer requests for Live Metrics, it is exactly what you mentioned, totalRequestDuration/totalRequestcount for each 15 mins

AutoCollection/StatsBeat.ts
heyams: on the kusto side, you want to see the string right? number is not that intuitive. however, as long as you understand which number maps to which instrumentation, that's up to you. it's just not that clear to others.  Here is fine, but on the kusto end you might want to have another map to the string, and t would be nice if you can document it in the constants class to indicate which number maps to which instrumentation name.

AutoCollection/Statsbeat.ts
```diff
@@ -0,0 +1,263 @@
...
+        let properties = {
+            "os": this._os,
+            "rp": this._resourceProvider,
+            "cikey": this._cikey,
+            "runtimeVersion": this._runtimeVersion,
+            "language": this._language,
+            "version": this._sdkVersion,
+            "attach": this._attach,
+            "instrumentation": this._instrumentations,
+            "feature": this._features,
```
heyams, line 208: feature is sent every 24 hours.  it's not part of the network statsbeat. please check the design doc: https://bit.ly/3zEf3Vx

AutoCollection/Statsbeat.ts
```diff
@@ -0,0 +1,263 @@
...
+        this._cikey = this._config.instrumentationKey;
+        this._sdkVersion = Context.sdkVersion; // "node" or "node-nativeperf"
+        this._os = os.type();
+        this._runtimeVersion = process.version;
+    }
+
+    private _getResourceProvider(callback: () => void) {
+        // Check resource provider
+        let waiting: boolean = false;
+        if (process.env.WEBSITE_SITE_NAME) { // Web apps
```
heyams, line 235: where is rpId? I didn't find it.

AutoCollection/Statsbeat.ts
```diff
@@ -0,0 +1,263 @@
...
+            "instrumentation": this._instrumentations,
+            "feature": this._features,
+        }
+        for (let i = 0; i < this._statbeatMetrics.length; i++) {
+            let statsbeat: Contracts.MetricTelemetry = {
+                name: this._statbeatMetrics[i].name,
+                value: this._statbeatMetrics[i].value,
+                properties: properties
+            };
+            let envelope = EnvelopeFactory.createEnvelope(statsbeat, Contracts.TelemetryType.Metric, null, null, this._statsbeatConfig);
```
heyams, line 216: it looks like config is used to store statsbeat connection string only.  can you just pass the connection string to createEnvelop? unless you will parse connection string from the json config file later?

AutoCollection/diagnostic-channel/azure-coretracing.sub.ts
```diff
@@ -37,15 +38,22 @@ export const subscriber = (event: IStandardEvent<Span>) => {
 };
 
 export function enable(enabled: boolean, client: TelemetryClient) {
+    let statsbeat = client.getStatsbeat();
     if (enabled) {
         if (clients.length === 0) {
             channel.subscribe<any>("azure-coretracing", subscriber);
+            if (statsbeat) {
+                statsbeat.addInstrumentation(StatsbeatInstrumentation.AZURE_CORE_TRACING);
```
heyams, line 46: never mind about the other comment about instrumentation.. i think name is a good indicate what each number is for.

AutoCollection/Statsbeat.ts
```diff
@@ -0,0 +1,263 @@
...
+    private _cikey: string;
+    private _attach: string = Constants.StatsbeatAttach.sdk; // Default is SDK
+    private _features: number = Constants.StatsbeatFeature.NONE;
+    private _instrumentations: number = Constants.StatsbeatInstrumentation.NONE;
+
+    constructor(config?: Config) {
+        this._isInitialized = false;
+        this._statbeatMetrics = [];
+        this._config = config;
+        this._statsbeatConfig = new Config(STATSBEAT_CONNECTIONSTRING);
```
heyams, line 55: it would be nice if you can get statsbeatConnectionString from the input parameter Config, then you can add statsbeat test to the e2e testing suite :)

heyams: maybe add a todo for later if you like.

Tests/AutoCollection/Statsbeat.tests.ts
```diff
@@ -0,0 +1,209 @@
...
+        it("It adds correct properties to custom metric", (done) => {
+            const statsBeat: Statsbeat = new Statsbeat(config);
+            statsBeat.enable(true);
+            const spy = sandbox.spy(statsBeat["_sender"], "send");
+            statsBeat.countRequest(123, true);
+            statsBeat.trackStatsbeatMetrics().then(() => {
+                assert.equal(spy.callCount, 1, "should call sender");
+                let envelope = spy.args[0][0][0];
+                assert.equal(envelope.name, "Statsbeat");
+                assert.equal(envelope.iKey, "c4a29126-a7cb-47e5-b348-11414998b11e");
```
heyams, line 109: it will be better if you don't test against the prod statsbeat ikey.  make it configurable will help.

heyams: and it will minimize testing data infiltering with the customer's stats data.

AutoCollection/Statsbeat.ts
```diff
@@ -0,0 +1,263 @@
...
+    public async trackStatsbeatMetrics() {
+        this._getResourceProvider(async () => {
+            this._trackRequestDuration();
+            this._trackRequestsCount();
+            await this._sendStatsbeats();
+        });
+    }
+
+    private _trackRequestDuration() {
+        var lastRequests = this._lastRequests;
```
MSNev, line 159: nit: Technically, (but probably unlikely) someone could call trackStatsbeatMetic() (as it's public) before its really enabled which would mean that this._lastRequests will be ```undefined`` and then cause bad things to happen here

MSNev (approved):

AutoCollection/Statsbeat.ts
hectorhdzg: I created a task for that, current there is no way to configure in attach scenarios but it would be great to have something available

AutoCollection/Statsbeat.ts
hectorhdzg: I thought rpId only exist in "attach" Statsbeat, this PR is only adding network Statbeats will add attach and feature in different PR

AutoCollection/Statsbeat.ts
hectorhdzg: Feature statbeat is not part of this PR I'm planning to include in other PR adding attach one as well. can you confirm if attach is correct here in network stats?

AutoCollection/Statsbeat.ts
hectorhdzg: We are using same mechanism as other envelopes created in the SDK, I'm trying to avoid specific code paths just for Statbeat here, added connection string as static so it could be updated if we add JSON config later

AutoCollection/Statsbeat.ts
heyams: yes, rpid is part of attach statsbeat.

AutoCollection/Statsbeat.ts
heyams: then please don't add feature in the network statsbeat payload. feature is on its own.

hectorhdzg: Must be removed now

heyams (approved):
please exclude feature from the network payload.  other than that, it looks good to me.

Issue #432, opened 2018-10-04
HttpDependencyParser.ts uses non-compliant `url.parse()` instead of `url.URL()`


avoidwork:
Hi,

`HttpDependencyParser.ts` should be using `new URL()` to generate a correctly parsed URL object, specifically the `pathname`. The root issue is `url.parse()` is very flawed, and was replaced quite awhile ago by the WHATWG compliant `url.URL` constructor.

Example (purposefully flawed URL results in flawed parsing):

![screen shot 2018-10-04 at 6 31 51 am](https://user-images.githubusercontent.com/369698/46468756-5daac400-c79f-11e8-922d-0c68e4eaedcd.png)

OsvaldoRosado:
This will probably need special casing, since this SDK has backcompat to Node 0.10, and if I'm reading documentation correctly the URL constructor was introduced in Node 7.

In case anyone is looking to contribute a PR, `url.parse` is also used in at least `HttpRequestParser.ts`, `TelemetryClient.ts`, and `Util.ts`

avoidwork:
Hi,

As someone who authors a few things... I can honestly say that trying to support 0.10 today is going to hurt the project. You should be targeting current LTS and when one loses support that'd be a chance for an LTS branch on this project.

Anyone running old node in production or on a dev machine today is either a masochist or not the target audience. ;)

OsvaldoRosado:
I wouldn't be very surprised if we reduced the version support in our next major release, but this would need special casing regardless. I anticipate Node 6 would still fall into support by this SDK as it's a maintained LTS line until well into 2019 :smile:

avoidwork:
nice.

PR #793, opened 2021-07-16
Update to use WHATWG URL API instead of url.parse()

xiao-lix:
This PR includes: 
- Update to WHATWG URL API instead of `url.parse()` method.
- Addresses #432 
- URL API is stable on Node 8: https://nodejs.org/docs/latest-v8.x/api/url.html#url_url

Details: 
- [url.parse](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) is deprecated
- The [WHATWG URL API](https://nodejs.org/api/url.html#url_the_whatwg_url_api) is recommended

Note: 
`new URL(input[, base])` : [details](https://nodejs.org/api/url.html#url_new_url_input_base)
- When creating a new URL object, if the first parameter `input `is relative, then `base `is required. If `input `is absolute, the `base `is ignored.
- A `TypeError `will be thrown if the `input `or `base `are not valid URLs.

### DIFF
--- a/AutoCollection/HttpDependencyParser.ts
+++ b/AutoCollection/HttpDependencyParser.ts
@@ -1,10 +1,7 @@
 import http = require("http");
 import https = require("https");
 import url = require("url");
-
 import Contracts = require("../Declarations/Contracts");
-import TelemetryClient = require("../Library/TelemetryClient");
-import Logging = require("../Library/Logging");
 import Util = require("../Library/Util");
 import RequestResponseHeaders = require("../Library/RequestResponseHeaders");
 import RequestParser = require("./RequestParser");
@@ -46,7 +43,7 @@ class HttpDependencyParser extends RequestParser {
      * Gets a dependency data contract object for a completed ClientRequest.
      */
     public getDependencyTelemetry(baseTelemetry?: Contracts.Telemetry, dependencyId?: string): Contracts.DependencyTelemetry {
-        let urlObject = url.parse(this.url);
+        let urlObject = new url.URL(this.url);
         urlObject.search = undefined;
         urlObject.hash = undefined;
         let dependencyName = this.method.toUpperCase() + " " + urlObject.pathname;
@@ -116,17 +113,17 @@ class HttpDependencyParser extends RequestParser {
         if (typeof options === 'string') {
             if (options.indexOf("http://") === 0 || options.indexOf("https://") === 0) {
                 // protocol exists, parse normally
-                options = url.parse(options);
+                options = new url.URL(options);
             } else {
                 // protocol not found, insert http/https where appropriate
-                const parsed = url.parse(options);
-                if (parsed.host === "443") {
-                    options = url.parse("https://" + options);
+                const parsed = new url.URL("http://" + options);
+                if (parsed.port === "443") {
+                    options = new url.URL("https://" + options);
                 } else {
-                    options = url.parse("http://" + options)
+                    options = new url.URL("http://" + options);
                 }
             }
-        } else if (options && typeof (url as any).URL === 'function' && options instanceof (url as any).URL) {
+        } else if (options && typeof url.URL === 'function' && options instanceof URL) {
             return url.format(options);
         } else {
             // Avoid modifying the original options object.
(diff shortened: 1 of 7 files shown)


AutoCollection/HttpDependencyParser.ts
```diff
@@ -46,7 +43,7 @@ class HttpDependencyParser extends RequestParser {
      * Gets a dependency data contract object for a completed ClientRequest.
      */
     public getDependencyTelemetry(baseTelemetry?: Contracts.Telemetry, dependencyId?: string): Contracts.DependencyTelemetry {
-        let urlObject = url.parse(this.url);
+        let urlObject = new url.URL(this.url);
```
xiao-lix, line 46: When `this.url` is used, it is the full url (including protocol) so ignore `base`.

AutoCollection/HttpDependencyParser.ts
```diff
@@ -116,17 +113,17 @@ class HttpDependencyParser extends RequestParser {
...
             if (options.indexOf("http://") === 0 || options.indexOf("https://") === 0) {
                 // protocol exists, parse normally
-                options = url.parse(options);
+                options = new url.URL(options);
             } else {
                 // protocol not found, insert http/https where appropriate
-                const parsed = url.parse(options);
-                if (parsed.host === "443") {
-                    options = url.parse("https://" + options);
+                const parsed = new url.URL("http://" + options);
```
xiao-lix, line 119: The protocol "http://" on this line is in order to make the new URL work - it requires protocol to parse url:
`new URL(input[, base])`
- input <string> The absolute or relative input URL to parse. If input is relative, then base is required. If input is absolute, the base is ignored.

AutoCollection/HttpRequestParser.ts
```diff
@@ -167,13 +165,15 @@ class HttpRequestParser extends RequestParser {
         }
 
         var encrypted = (<any>request).connection ? ((<any>request).connection as any).encrypted : null;
-        var requestUrl = url.parse(request.url);
+
+        var protocol = (encrypted || request.headers["x-forwarded-proto"] == "https") ? "https" : "http";
```
xiao-lix, line 169: Update the sequence to get the protocol first, then URL API can be used - it requires protocol in url.

Library/TelemetryClient.ts
```diff
@@ -118,7 +118,14 @@ class TelemetryClient {
             // url.parse().host returns null for non-urls,
             // making this essentially a no-op in those cases
             // If this logic is moved, update jsdoc in DependencyTelemetry.target
-            telemetry.target = url.parse(telemetry.data).host;
+            // url.parse() is deprecated, update to use WHATWG URL API instead
+            try {
+                telemetry.target = new url.URL(telemetry.data).host;
```
xiao-lix, line 123: A `TypeError `will be thrown if the input or base are not valid URLs. 
To be compliant with previous behavior, when url is invalid, set `telemetry.target` as `null`.

Tests/AutoCollection/HttpDependencyParser.tests.ts
```diff
@@ -56,8 +56,8 @@ describe("AutoCollection/HttpDependencyParser", () => {
             assert.equal(dependencyTelemetry.dependencyTypeName, Contracts.RemoteDependencyDataConstants.TYPE_HTTP);
             assert.equal(dependencyTelemetry.success, true);
             assert.equal(dependencyTelemetry.name, "GET /search");
-            assert.equal(dependencyTelemetry.data, "https://a.bing.com:443/search");
-            assert.equal(dependencyTelemetry.target, "a.bing.com:443");
+            assert.equal(dependencyTelemetry.data, "https://a.bing.com/search");
+            assert.equal(dependencyTelemetry.target, "a.bing.com");
```
xiao-lix, line 60: Remove port because : 
When using URL API to create a new object, when port matches with the protocol, the port is dropped. 

For example: 
- new URL("**https**://a.bing.com:443/search").port returns ""      - port dropped since https matches with 443
- new URL("**http**://a.bing.com:443/search").port returns "443"  - port not dropped
- new URL("**http**://a.bing.com:80/search").port returns ""          - port dropped since http matches with 80
- new URL("**https**://a.bing.com:80/search").port returns "80"     - port not dropped
- new URL("**https**://a.bing.com:1234/search").port returns "1234"   - port not dropped
- new URL("**http**://a.bing.com:1234/search").port returns "1234"     - port not dropped 

image

Confirmed with Nev that : the default port for a protocol is often never sent @hectorhdzg can you confirm this? Thanks.

hectorhdzg (approved):
LGTM, thanks

PR #799, opened 2021-07-26
Adding Attach and Feature Statbeats

hectorhdzg:


AutoCollection/Statsbeat.ts
```diff
@@ -148,15 +164,52 @@ class Statsbeat {
...
+            this._trackRequestsCount(networkProperties);
+            await this._sendStatsbeats();
+        });
+    }
+
     public async trackStatsbeatMetrics() {
         this._getResourceProvider(async () => {
-            this._trackRequestDuration();
-            this._trackRequestsCount();
+            let attachProperties = {
```
hectorhdzg, line 187: @heyams the spec doesn't mention attach property to determine if "codeless" or "sdk" for this Statsbeat can you confirm if is by design?

AutoCollection/Statsbeat.ts
```diff
@@ -232,17 +275,23 @@ class Statsbeat {
     private _getResourceProvider(callback: () => void) {
         // Check resource provider
         let waiting: boolean = false;
+        this._resourceIdentifier = Constants.StatsbeatResourceProvider.unknown;
         if (process.env.WEBSITE_SITE_NAME) { // Web apps
             this._resourceProvider = Constants.StatsbeatResourceProvider.appsvc;
+            this._resourceIdentifier = process.env.WEBSITE_SITE_NAME;
```
lzchen, line 281: @heyams  Current specs says `WEBSITE_SITE_NAME/WEBSITE_HOME_STAMPNAME` for `rpid`. Does this mean the literal string (with the slash concatenating the two), or choose `WEBSITE_SITE_NAME` if available, otherwise `WEBSITE_HOME_STAMPNAME`?

heyams: I expect WEBSITE_SITE_NAME/WEBSITE_HOME_STAMPNAME.  if stampName is empty, it will be "{WEBSITE_SITE_NAME}/".  yes.

AutoCollection/Statsbeat.ts
heyams: attach statsbeats is different from attach type, which is codeless or sdk.  attach statsbeat is used to collect rp and rpid for attach rates later. hope that makes sense.  i could rename attach in the customer dimensions to sdkType? it's either codeless or sdk.

AutoCollection/Statsbeat.ts
```diff
@@ -232,17 +275,23 @@ class Statsbeat {
...
         } else if (process.env.FUNCTIONS_WORKER_RUNTIME) { // Function apps
             this._resourceProvider = Constants.StatsbeatResourceProvider.function;
+            if (process.env.WEBSITE_HOSTNAME) {
+                this._resourceIdentifier = process.env.WEBSITE_HOSTNAME;
+            }
         } else if (this._config) {
```
lzchen, line 287: What if the application is not in any of these rps? We probably shouldn't be sending this stats at all.

heyams: rp has a default case: "unknown".  like standalone apps.. it's not an attach scenario.

AutoCollection/Statsbeat.ts
```diff
@@ -148,15 +164,52 @@ class Statsbeat {
...
+            let attachProperties = {
+                "os": this._os,
+                "rp": this._resourceProvider,
+                "rpid": this._resourceIdentifier,
+                "cikey": this._cikey,
+                "runtimeVersion": this._runtimeVersion,
+                "language": this._language,
+                "version": this._sdkVersion
+            }
+            this._statbeatMetrics.push({ name: Constants.StatsbeatCounter.ATTACH, value: 1, properties: attachProperties });
```
heyams, line 196: attach statsbeat is every 15 mins.  feature is every 24 hours. you might want to have a separate method for attach and one for feature.

AutoCollection/Statsbeat.ts
lzchen: Right, so we should not be sending any of the "attach" metrics in this default case.

AutoCollection/Statsbeat.ts
lzchen: @heyams For applications that exit before 24 hours, do we send a statsbeat telemetry upon exit?

AutoCollection/Statsbeat.ts
hectorhdzg: You can see this exactly what the code is supposed to be doing here
https://github.com/microsoft/ApplicationInsights-E2E-Tests/blob/main/AppInsightsE2ETest/E2ETestCases/TestsConfiguration.cs#L90
It must only use setting value when env var is not present, it would be great if you can fix the issue in the code we are all using in this case

AutoCollection/Statsbeat.ts
hectorhdzg: @heyams  Why attach is being sent every 15 minutes?, are you expecting rpId to be updated that often?

AutoCollection/Statsbeat.ts
hectorhdzg: @heyams can you confirm we don't send attach Statbeat for unknown

AutoCollection/Statsbeat.ts
heyams: this is the same interval as the azure metadata service.

AutoCollection/Statsbeat.ts
heyams: I would like to track unknown as well.. i.e. non-attach vs. attach.  that's a good data trend to show.

AutoCollection/Statsbeat.ts
heyams: vm id can get changed frequently, which impacts rpId for vm. other attach scenarios is not an issue.

heyams: > @heyams For applications that exit before 24 hours, do we send a statsbeat telemetry upon exit?

it will be nice.  currently java doesn't do that.

AutoCollection/Statsbeat.ts
```diff
@@ -70,18 +73,31 @@ class Statsbeat {
                     time: +new Date
                 };
                 this._handle = setInterval(() => {
-                    this.trackStatsbeatMetrics().catch((error) => {
+                    this.trackShortIntervalStatsbeats().catch((error) => {
                         // Failed to send Statsbeat
                         Logging.info(Statsbeat.TAG, error);
                     });
                 }, Statsbeat.NETWORK_STATS_COLLECTION_INTERVAL);
```
heyams, line 80: can you rename this constant to be short_interval similar to 'LONG_STATS_COLLECTION_INTERVAL'?

heyams (approved):
One minor constant renaming.. other than that, looks good to me.

AutoCollection/Statsbeat.ts
lzchen: What do we populate for properties then? Do we just leave empty?

AutoCollection/Statsbeat.ts
lzchen: It's not only about "nice". If applications exit, there may be cases where we never get the telemetry. We MUST flush before exit.

lzchen: @hectorhdzg probably not blocking for this pr though

AutoCollection/Statsbeat.ts
```diff
@@ -148,15 +164,52 @@ class Statsbeat {
...
+                "version": this._sdkVersion,
+                "attach": this._attach,
+                "instrumentation": this._instrumentations,
+            }
+            this._trackRequestDuration(networkProperties);
+            this._trackRequestsCount(networkProperties);
+            let attachProperties = {
+                "os": this._os,
+                "rp": this._resourceProvider,
+                "rpid": this._resourceIdentifier,
```
lzchen, line 184: Curious, what will `undefined` serialize as? Empty string? This is for the case when `rp` == `unknown`

AutoCollection/Statsbeat.ts
```diff
@@ -148,15 +164,52 @@ class Statsbeat {
...
-            this._trackRequestsCount();
+            let featureProperties = {
+                "os": this._os,
+                "rp": this._resourceProvider,
+                "cikey": this._cikey,
+                "runtimeVersion": this._runtimeVersion,
+                "language": this._language,
+                "version": this._sdkVersion,
+                "attach": this._attach,
+                "feature": this._features,
```
lzchen, line 205: The "aad handling" is listed as a feature. Will stats be released as the next beta version?

lzchen:
Statsbeats metrics related to instrumentations have not been added yet correct? @hectorhdzg

AutoCollection/Statsbeat.ts
lzchen, line 205: @heyams @hectorhdzg Do we have a comprehensive list of all the `features` we will be tracking for each language? It would also be good to coordinate bit representations for common features across our sdks.

heyams: each language will have their own specific features..  tracking AAD enablement will be a common feature.. other than that, there is nothing else i can think of.

heyams: @hectorhdzg is aad handling the same as tracking aad enablement? i check your repo.. look like this is not added anywhere to the feature list. it only happens in tests.

lzchen: What about local storage? Standard metrics? Pre-aggregated metrics? Live metrics? All of these are/will be features that will be supported by all sdks so we should agree on a list.

AutoCollection/Statsbeat.ts
hectorhdzg: @lzchen we flush all telemetry on exit, but we do not create telemetry on exit, we don't want to do that, that will make sense with high priority telemetry, maybe something that actually let users know that the app stopped running, but that doesn't apply with Statsbeats

AutoCollection/Statsbeat.ts
hectorhdzg: We only have AAD feature in beta branch, I'm planning to update beta with latest changes in main for next release including Statsbeat implementation

hectorhdzg: @lzchen having a shared list of features will make easier the data consumption here, happy to add them once we agree on them

AutoCollection/Statsbeat.ts
hectorhdzg: this._resourceIdentifier defaults to "unknown" so it supposed to never be undefined, are you seeing something here?
I believe undefined properties are dropped when serializing JSON, at least they are when running this in the browser
JSON.stringify({"test":undefined})
"{}"

hectorhdzg:
@lzchen we send instrumentation property in all network Statsbeats, I don't believe there is an "Instrumentation" specific one.

AutoCollection/Statsbeat.ts
heyams: Statsbeat is designed only for Breeze endpoints. Live metrics is not tracked. Java hasn't implemented pre-aggregated metrics and standard metrics yet.  
I am planning to create a new set of statsbeat metrics for Local Storage.  As of now, AAD is the only common feature.  Let's get the basic Statsbeat done first and we can add the rest as Statsbeat evolves.. A list to agree upon at this moment is premature.  Please document it somewhere for future consideration.  (cc @ramthi)

AutoCollection/Statsbeat.ts
lzchen: So on non-attach scenario, the payload would look like: {"rp": "unknown"} with no "rpid" correct?

AutoCollection/Statsbeat.ts
hectorhdzg: Statbeats will be part of both releases, beta and regular one

AutoCollection/Statsbeat.ts
hectorhdzg: Actually rpid is also included, something like this {"rp": "unknown","rpid": "unknown"}

AutoCollection/Statsbeat.ts
lzchen: Interesting. Something to bring up at tomorrow's meeting.

lzchen (approved):

PR #801, opened 2021-07-28
feat: use @azure/logger

### LINKED ISSUE #706: [Help] Structured logging

### LINKED ISSUE #731: Use @azure/logger for logging

xiao-lix:
- Fixed conflicts in this PR #733 
- Replace other Loggings where internal log is used
- Fixed tests - instead of testing with Logging class, Channel and Sender are now using private `_logger` instead, tests should test against `_logger.warning` mthods, etc.

I don't have permission to push to jmealo's fork, create a new one and could: 
Closes #733 
Closes #731 
Closes #706


Library/CorrelationIdManager.ts
```diff
@@ -49,7 +45,7 @@ class CorrelationIdManager {
                 disableAppInsightsAutoCollection: true
             };
 
-            Logging.info(CorrelationIdManager.TAG, requestOptions);
```
hectorhdzg, line 52: Tags are really important in the logs as we can easily identify where the log is coming from, please add them as part of the message

xiao-lix: added

PR #805, opened 2021-08-05
Add category and endpoint in Network Statsbeats

hectorhdzg:


MSNev (approved):

AutoCollection/Statsbeat.ts
```diff
@@ -209,42 +201,64 @@ class Statsbeat {
...
-        if (this._throttleCount > 0) {
-            this._statbeatMetrics.push({ name: Constants.StatsbeatCounter.THROTTLE_COUNT, value: this._throttleCount, properties: properties });
-            this._throttleCount = 0; //Reset
+    private _trackRequestDuration(commonProperties: {}) {
+        for (let i = 0; i < this._networkStatsbeatCollection.length; i++) {
+            var currentCounter = this._networkStatsbeatCollection[i];
+            currentCounter.time = +new Date;
+            var intervalRequests = (currentCounter.totalRequestCount - currentCounter.lastRequestCount) || 0;
+            var elapsedMs = currentCounter.time - currentCounter.lastTime;
+            var averageRequestExecutionTime = ((currentCounter.intervalRequestExecutionTime - currentCounter.lastIntervalRequestExecutionTime) / intervalRequests) || 0;
```
lzchen, line 225: Could there be a divide by zero error thrown here?

AutoCollection/Statsbeat.ts
```diff
@@ -18,17 +19,7 @@ class Statsbeat {
...
-    private _totalRequestCount: number = 0;
-    private _totalSuccesfulRequestCount: number = 0;
-    private _totalFailedRequestCount: number = 0;
-    private _retryCount: number = 0;
-    private _exceptionCount: number = 0;
-    private _throttleCount: number = 0;
-    private _intervalRequestExecutionTime: number = 0;
-    private _lastIntervalRequestExecutionTime: number = 0;
-
+    private _networkStatsbeatCollection: Array<Network.NetworkStatsbeat>;
```
lzchen, line 22: I'm wondering if a dictionary can be used here? With (category, endpoint) as the key?

AutoCollection/Statsbeat.ts
```diff
@@ -209,42 +201,64 @@ class Statsbeat {
...
+            var currentCounter = this._networkStatsbeatCollection[i];
+            currentCounter.time = +new Date;
+            var intervalRequests = (currentCounter.totalRequestCount - currentCounter.lastRequestCount) || 0;
+            var elapsedMs = currentCounter.time - currentCounter.lastTime;
+            var averageRequestExecutionTime = ((currentCounter.intervalRequestExecutionTime - currentCounter.lastIntervalRequestExecutionTime) / intervalRequests) || 0;
+            currentCounter.lastIntervalRequestExecutionTime = currentCounter.intervalRequestExecutionTime; // reset
+            if (elapsedMs > 0 && intervalRequests > 0) {
+                // Add extra properties
+                let properties = Object.assign({ "category": this._networkStatsbeatCollection[i].category, "endpoint": this._networkStatsbeatCollection[i].endpoint }, commonProperties);
+                this._statbeatMetrics.push({ name: Constants.StatsbeatCounter.REQUEST_DURATION, value: averageRequestExecutionTime, properties: properties });
```
lzchen, line 230: I think part of the specs says that if the value of the stats is 0 for network stats, we don't send the metric? Might need a check before appending to `_statsbeatMetrics`

hectorhdzg: Check must be right there a few lines before, we don't send duration unless there was a request and elapsed milliseconds

AutoCollection/Statsbeat.ts
hectorhdzg: Normal size of this array would be less than 3, could be slightly bigger in case of redirects triggered from the backend, using some kind of hash table sound to much for this scenario

AutoCollection/Statsbeat.ts
hectorhdzg: Node.js doesn't throw an error with division by zero, it will return "Infinity" and metric will not be send because intervalRequests would be 0

AutoCollection/Statsbeat.ts
```diff
@@ -134,39 +121,44 @@ class Statsbeat {
...
         }
         else {
-            this._totalSuccesfulRequestCount++;
+            counter.totalSuccesfulRequestCount++;
         }
+
     }
 
-    public countException() {
+    public countException(category: number, endpoint: string) {
```
heyams, line 140: category and endpoint only matter when it's actually sending the exception count, right? I am not sure why incrementing the count care about these two input..

AutoCollection/Statsbeat.ts
```diff
@@ -134,39 +121,44 @@ class Statsbeat {
...
         if (!this.isEnabled()) {
             return;
         }
-        this._exceptionCount++;
+        let counter: Network.NetworkStatsbeat = this._getNetworkStatsbeatCounter(category, endpoint);
+        counter.exceptionCount++;
     }
 
-    public countThrottle() {
+    public countThrottle(category: number, endpoint: string) {
```
heyams, line 148: same here

AutoCollection/Statsbeat.ts
```diff
@@ -134,39 +121,44 @@ class Statsbeat {
...
         if (!this.isEnabled()) {
             return;
         }
-        this._throttleCount++;
+        let counter: Network.NetworkStatsbeat = this._getNetworkStatsbeatCounter(category, endpoint);
+        counter.throttleCount++;
     }
 
-    public countRetry() {
+    public countRetry(category: number, endpoint: string) {
```
heyams, line 156: same here

AutoCollection/Statsbeat.ts
hectorhdzg: Well but the counts are specific to a category/endpoint combination, Breeze could have 10 exceptions and Quickpulse 6, we don't want to mix counts in here, what are you proposing instead?

heyams: it makes sense to me to track count at the category level.. it's crazy to count at the endpoint level.. there are lots of stamp specific endpoints for breeze.  we want to know exception count to breeze not at a specific stamp endpoint.. for live metric, it's the same host name.

hectorhdzg: We don't expect customer to hit tons of endpoints, default would be something like this "dc.services.visualstudio.com", if redirect is triggered we will some other Statsbeat with endpoint "centralus-2.in.applicationinsights.azure.com", if customer provide custom endpointUrl during initialization we will get that as well, I'm just implementing what we discussed previously and I thought everyone agreed.

MSNev: As hector said, while breeze may have multiple servers (stamp) this endpoint is the one configured in the app not the stamp specific VM.
And I think that it does actually make sense to track exceptions (and all below) as the endpoint level as if we introduce a bug which causes an exception or the user is getting throttled etc this is all good information to help track down issues

heyams: i think we might have a misunderstanding. i think we just want the total exception count.  for debugging, exception detail will be available in the exception table.  this count won't give much info about the actual cause of the exception. 

dc.services.visualstudio.com
centralus-2.in.applicationinsights.azure.com
are all breeze endpoint.. 

I am saying we should aggregate the exception count at breeze not at the host.  It doesn't offer much benefit except sending more Exception Count metrics than necessary.

MSNev: It won't be sending ```more Exception Count metrics``` as an application will generally only have 1 endpoint that it's sending too (excluding redirect scenario, at which point it just changes to the new one).

What it does change is the cardinality of the "complete" collection as it this is an additional dimension for the metrics.

heyams: well, java agent supports live polling of the json config.. like azure sdk.. customers can change their connection string and sampling rate at runtime.. depends on how often the connection string gets updated.. this will trigger equivalent amount of the Count metrics and then multiplies by 6 (different network statsbeat metric count).  for regular customers who use single ikey throughout the lifetime of the app, category and endpoint won't get changed at all and there is no point to pass it in for each increment either.

lzchen (approved):

heyams (approved):
👍

heyams:
it will be nice to have a test for count per host.

PR #817, opened 2021-08-16
Respect existing trace headers when using Azure Function contexts

### LINKED ISSUE #816: API Management <> Azure Functions missing links
Hi,

I've set up the same App Insights instance for APIM and Azure Function but I cannot track requests end-to-end.

It shows 2 Clusters/Paths:
1) Example-APIM-Name -> Example-AzureFunction-Name
2) "Web" (Static Name) to MongoDB

There's no connection/Link between the AzureFunction Name and the static Web path making correlation extremely hard.

Any clue how to tell the AzureFunction to identify with it's actual name (Example-AzureFunction-Name) instead of "Web" and link itself to the correlation path behind the APIM?

f0o:
FIxes #816 

Disclaimer: I have very little experience with Typescript, patch was original written in plain JS and then retro-fitted for TS.

Original Patch:
```javascript
            var traceparent = null
            var tracestate = null
            if( "req" in context ) {
                if( "traceparent" in context.req.headers ) {
                    traceparent = new Traceparent(request.headers.traceparent);
                } else if( "request-id" in context.req.headers ) {
                    traceparent = new Traceparent(null, request.headers["request-id"]);
                }
                if( "tracestate" in context.req.headers ) {
                    tracestate = new Tracestate(request.headers.tracestate);
                }
            }
            if( !traceparent ) {
                traceparent = new Traceparent(traceContext.traceparent);
            }
            if( !tracestate ) {
                tracestate = new Tracestate(traceContext.tracestate);
            }
```


hectorhdzg:
Thanks @f0o for submitting a PR, looking at the code I can see you are adding pretty similar code to next if [statement](https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/AutoCollection/CorrelationContextManager.ts#L220) in that method, most likely this could be fixed by adding an extra check on top of just checking for traceContext to exist, I'm a little bit confused about what exactly is happening here, is traceContext not well formed and that is what is causing the issue?, I'm trying to setup some APIM environment to try it out

hectorhdzg:
Looks like the problem is some service in your system is not providing correct headers, Azure Functions requires traceparent headers to maintain correlation and apparently that is the one where it breaks, in this case traceContext have priority and that is the value we want to use, if not there, the request headers will be used in next if statement, this could potentially break scenarios where traceContext is correct.

f0o:
@hectorhdzg if by my services you mean Azure Monitor, Azure APIM and Azure WebTests then yeah sure. I'm however not responsible for Azure Internals so the only way to fix it to work properly with Azure Products is using above Patch.

PR #836, opened 2021-09-09
 Use Website instance ID when running in Azure

hectorhdzg:


MSNev (approved):

heyams:
where is the code it defaults to hostname when instance_id is not available?

hectorhdzg:
@heyams code is here
https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/Library/Context.ts#L47 it will be overwritten by this telemetry processor if enabled and running in Azure

heyams (approved):

Issue #837, opened 2021-09-10
"Role Name" does not show up in Live Metric in Azure Portal


PhakornKiong:
Reproducible Repo:
https://github.com/PhakornKiong/applicationInsightsRoleName

Node v12.22.1
"applicationinsights": "^2.1.6",
"express": "^4.17.1"

Description: "Role Name" always default to "Web" although `RoleName` field had been configured in accordance to the documentation [here](https://github.com/microsoft/ApplicationInsights-node.js#multiple-roles-for-multi-component-applications).

I've checked and confirmed that the setting of `RoleName` works for "Application Map" but not "Live Metrics", the setting was reflected in the logs as well.

Expected Behavior: "Role Name" should be the same with `RoleName` field.

![image](https://user-images.githubusercontent.com/30017432/132828908-fa63fccc-3196-46ba-bd35-22c4f0982e10.png)
![image](https://user-images.githubusercontent.com/30017432/132829083-52fa65ed-4ffa-474f-bf98-f6941d830611.png)
![image](https://user-images.githubusercontent.com/30017432/132829366-febf67c9-2cd8-459b-a56a-3520d0c360e7.png)

PR #838, opened 2021-09-10
applicationinsights.ts

PhakornKiong:
Fix https://github.com/microsoft/ApplicationInsights-node.js/issues/837

- pass defaultClient.context to QuickPulseClient

While tinkering with the debugger, I noticed that the constructor for `QuickPulseStateManager `  will use default context if one is not provided.
https://github.com/microsoft/ApplicationInsights-node.js/blob/a327c10f53cc4e93699fb5c2b196bc0906b7882d/Library/QuickPulseStateManager.ts#L35-L40

and

null was always passed to `QuickPulseStateManager`, hence the default context will always be used
https://github.com/microsoft/ApplicationInsights-node.js/blob/a327c10f53cc4e93699fb5c2b196bc0906b7882d/applicationinsights.ts#L363-L369

This explains #837 inability to get "Role Name" updated in the portal.

### DIFF
--- a/applicationinsights.ts
+++ b/applicationinsights.ts
@@ -362,7 +362,7 @@ export class Configuration {
 
         if (!liveMetricsClient && enable) {
             // No qps client exists. Create one and prepare it to be enabled at .start()
-            liveMetricsClient = new QuickPulseClient(defaultClient.config, null);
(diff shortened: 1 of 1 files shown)


hectorhdzg (approved):
Thanks for your contribution

PR #845, opened 2021-09-20
Fix issue loading SDK version

hectorhdzg:
Adding test to ensure version is updated during releases

### DIFF
--- a/Library/Context.ts
+++ b/Library/Context.ts
@@ -1,22 +1,22 @@
 ﻿import os = require("os");
-import http = require("http");
 import fs = require("fs");
 import path = require("path");
 
 import Contracts = require("../Declarations/Contracts");
+import { APPLICATION_INSIGHTS_SDK_VERSION } from "../Declarations/Constants";
 import Logging = require("./Logging");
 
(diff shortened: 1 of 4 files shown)


Library/Context.ts
```diff
@@ -54,21 +54,7 @@ class Context {
...
-                let packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
-                if (packageJson && typeof packageJson.version === "string") {
-                    Context.sdkVersion = packageJson.version;
-                }
-            } catch (exception) {
-                Logging.info("unable to read app version: ", exception);
-            }
-        }
-
+        Context.sdkVersion = APPLICATION_INSIGHTS_SDK_VERSION;
```
xiao-lix, line 57: Just curious - why do you change the version to hardcoded rather than extract from package json?

hectorhdzg: Looking at Statsbeat data we can tell package json logic fails a lot, this will ensure we send correct version all the times

xiao-lix (approved):

PR #869, opened 2021-11-12
Add namespace for custom metrics

### LINKED ISSUE #609: Allow to provide metric namespace
Hi,

Looking at the [documentation](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-custom-overview?toc=%2Fazure%2Fazure-monitor%2Ftoc.json#namespace) and Application Insights dashboard, it seems metrics could be grouped by namespace.

Except if I missed something, this SDK doesn't allow to provide a namespace, and all metrics ends up in the `azure.applicationinsights` namespace.

image

hectorhdzg:
Fixes #609


Tests/Library/EnvelopeFactoryTests.ts
```diff
@@ -235,4 +235,32 @@ describe("Library/EnvelopeFactory", () => {
...
         });
     });
+
+    describe("MetricData", () => {
+        let metricTelemetry: Contracts.MetricTelemetry;
+        beforeEach(() => {
+            metricTelemetry = {
+                name: "TestName",
+                value: 123,
+                namespace: "TestNamespace",
```
MSNev, line 245: Is this something we should be adding to the web code as well?
We only have these
export class DataPoint {

    /**
     * Name of the metric.
     */
    public name: string;

    /**
     * Metric type. Single measurement or the aggregated value.
     */
    public kind: DataPointType = DataPointType.Measurement;

    /**
     * Single value for measurement. Sum of individual measurements for the aggregation.
     */
    public value: number;

    /**
     * Metric weight of the aggregated metric. Should not be set for a measurement.
     */
    public count: number;

    /**
     * Minimum value of the aggregated metric. Should not be set for a measurement.
     */
    public min: number;

    /**
     * Maximum value of the aggregated metric. Should not be set for a measurement.
     */
    public max: number;

    /**
     * Standard deviation of the aggregated metric. Should not be set for a measurement.
     */
    public stdDev: number;
}

hectorhdzg: Yes, you should, "ns" is the actual field we put in the envelope, namespace is the property name of the class we expose in the API

MSNev: OK, will raise an issue on AI-JS

MSNev (approved):

PR #875, opened 2021-12-01
JSON configuration support

hectorhdzg:


AutoCollection/AutoCollector.ts
```diff
@@ -0,0 +1,171 @@
...
+        this._serverRequests = new AutoCollectHttpRequests(client);
+        this._clientRequests = new AutoCollectHttpDependencies(client);
+        if (!this._nativePerformance) {
+            this._nativePerformance = new AutoCollectNativePerformance(client);
+        }
+    }
+
+    public start() {
+        this._isStarted = true;
+        this._console.enable(this.isConsole, this.isConsoleLog);
```
xiao-lix, line 56: Should the enable methods below only be called if "_isStarted" is false? Otherwise, for this case: 

1. configured flag `false` in json config file
2. call "setxxx" api to config the flag `true`
3. call start

The flag should be `false` because json config always take the highest priority. But in this case, the flag is `true` because setxxx api reset the flag before start. 
A more detailed example: [readme](https://github.com/microsoft/ApplicationInsights-node.js#configure-settings-through-json-file)
If setxxx api takes higher priority, then the `scenario 2` [here](https://github.com/microsoft/ApplicationInsights-node.js#configure-settings-through-json-file) breaks.

AutoCollection/AutoCollector.ts
```diff
@@ -0,0 +1,171 @@
...
+    private _preAggregatedMetrics: AutoCollecPreAggregatedMetrics;
+    private _heartbeat: HeartBeat;
+    private _nativePerformance: AutoCollectNativePerformance;
+    private _serverRequests: AutoCollectHttpRequests;
+    private _clientRequests: AutoCollectHttpDependencies;
+    private _client: TelemetryClient;
+    private _isStarted = false;
+
+    public setup(client: TelemetryClient) {
+        this._client = client;
```
xiao-lix, line 40: Compared the old code, what if the client already setup? Do we not need the logging info anymore?

xiao-lix: Never mind... I get it, the appinsights.ts setup keeps the previous logic.

AutoCollection/AutoCollector.ts
hectorhdzg: Flag should be true, explicit API calls in customer code will always have priority over env variables and JSON configuration

Declarations/Interfaces.ts
```diff
@@ -28,87 +26,71 @@ export interface IJsonConfig {
...
     /**
      * Sets the state of pre aggregated metrics tracking (enabled by default)
      * if true pre aggregated metrics will be collected every minute and sent to Application Insights
      */
-    enableAutoCollectPreAggregatedMetrics?: boolean;
-
+    enableAutoCollectPreAggregatedMetrics: boolean;
     /**
      * Sets the state of request tracking (enabled by default)
      * if true HeartBeat metric data will be collected every 15 mintues and sent to Application Insights
```
MSNev, line 71: type: minutes

MSNev (approved):

PR #881, opened 2021-12-13
Removing old functionality that would be replaced by OpenTelemetry

hectorhdzg:
Removing static auto collectors
Allow collection to be associated to specific client


hectorhdzg:
Updated will be merged to "preview" branch not main one

MSNev (approved):

PR #885, opened 2022-01-18
Add path to http dependency when host is missing

MohammedFoysal:
Whenever the hostname is populated but the host is missing, use the hostname and add in the path. This fixes the issue where the path is missing from app insights detailed in this issue #886


hectorhdzg (approved):
LGTM, thanks for your contribution

MohammedFoysal:
@hectorhdzg Apologies for the delay, the CLA has now been signed and the PR is ready to go 👍

PR #891, opened 2022-01-21
Karlie/web snippet

Karlie-777:


AutoCollection/WebSnippet/fetchSnippet.js
```diff
@@ -0,0 +1,26 @@
+const path = require('path');
+const fs = require('fs');
+const https = require('https');
+
+let snippetUrl = 'https://raw.githubusercontent.com/microsoft/ApplicationInsights-JS/master/AISKU/snippet/snippet.js';
```
hectorhdzg, line 5: I thought you were creating an npm package to avoid this

AutoCollection/WebSnippet/snippet.js
```diff
@@ -0,0 +1,328 @@
+﻿(function (win, doc, snipConfig) {
```
hectorhdzg, line 1: Remove hardcoded snippet

AutoCollection/WebSnippet/snippet.min.js
```diff
@@ -0,0 +1,10 @@
...nerror=a,n.onreadystatechange=function(e,t){"loaded"!==n.readyState&&"complete"!==n.readyState||i(0,t)},n}();y.ld<0?l.getElementsByTagName("head")[0].appendChild(e):setTimeout(function(){l.getElementsByTagName(k)[0].parentNode.appendChild(e)},y.ld||0)}try{m.cookie=l.cookie}catch(p){}function t(e){for(;e.length;)!function(t){m[t]=function(){var e=arguments;g||m.queue.push(function(){m[t].apply(m,e)})}}(e.pop())}var n="track",r="TrackPage",o="TrackEvent";t([n+"Event",n+"PageView",n+"Exception",n+"Trace",n+"DependencyData",n+"Metric",n+"PageViewPerformance","start"+r,"stop"+r,"start"+o,"stop"+o,"addTelemetryInitializer","setAuthenticatedUserContext","clearAuthenticatedUserContext","flush"]),m.SeverityLevel={Verbose:0,Information:1,Warning:2,Error:3,Critical:4};var s=(d.extensionConfig||{}).ApplicationInsightsAnalytics||{};if(!0!==d[I]&&!0!==s[I]){var c="onerror";t(["_"+c]);var u=T[c];T[c]=function(e,t,n,a,i){var r=u&&u(e,t,n,a,i);return!0!==r&&m["_"+c]({message:e,url:t,lineNumber:n,columnNumber:a,error:i}),r},d.autoExceptionInstrumented=!0}return m}(y.cfg);function a(){y.onInit&&y.onInit(n)}(T[t]=n).queue&&0===n.queue.length?(n.queue.push(a),n.trackPageView({})):a()}(window,document,{
```
hectorhdzg, line 1: Same here, remove this one as well

Declarations/Interfaces.ts
```diff
@@ -134,6 +134,12 @@ export interface IBaseConfig {
     * Live Metrics custom host
     */
     quickPulseHost: string;
+    /**
+     * Enable web snippet auto html injection, default to false
+     */
+     enableAutoWebSnippetInjection: boolean;
+     isDebugWebSnippet: boolean;
```
hectorhdzg, line 141: What is the difference between debug and not debug one?

AutoCollection/WebSnippet/fetchSnippet.js
Karlie-777: Yes, I will update the PR soon! Thank you Hector!

Declarations/Interfaces.ts
Karlie-777: debug use js version and none debug use min.js version. I modified it from previous PR

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,147 @@
...
+            }
+            // Read response stream
+            let html = input.toString();
+            // Try to add script before HTML head closure
+            let index = html.indexOf("</head>");
+            if (index >= 0) {
+                let subStart = html.substring(0, index);
+                let subEnd = html.substring(index);
+                input = subStart + '<script type="text/javascript">' + WebSnippet._snippet + '</script>' + subEnd;
+                html = subStart + '<script type="text/javascript">' + WebSnippet._snippet + '</script>' + subEnd;
```
MSNev, line 127: It looks like you only need one of these `input` or `html`

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+    public static INSTANCE: WebSnippet;
+
+    private static _snippet: string;
+    private static _aiUrl: string;
+    private _isEnabled: boolean;
+    private _isInitialized: boolean;
+
+    constructor(client: TelemetryClient) {
+        if (!!WebSnippet.INSTANCE) {
+            throw new Error("Web snippet injection should be configured from the applicationInsights object");
```
ramthi, line 19: Not sure what this scenario is.  The error message should describe what they did wrong and what they should do. Is this to ensure we have a singleton object of WebSnippet?

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        WebSnippet.INSTANCE = this;
+        // AI URL used to validate if snippet already included
+        WebSnippet._aiUrl = " https://js.monitor.azure.com/scripts/b/ai.2";
+
+        let snippetPath = path.resolve(__dirname, "../../node_modules/applicationinsight-web-snippet/snippet/snippet.min.js");
+        if (client.config.isDebugWebSnippet) {
+            snippetPath = path.resolve(__dirname, "../../node_modules/applicationinsight-web-snippet/snippet/snippet.js");
+        }
```
ramthi, line 29:
```suggestion
        let snippetPath = "../../node_modules/applicationinsight-web-snippet/snippet";
        let snippetFile = "snippet.min.js" //default filename
        if (client.config.isDebugWebSnippet) {
            snippetFile = "snippet.js";
        }
        snippetPath = path.resolve(snippetPath, snippetFile); //no need to bring in __dirname since path.resolve will use current folder to resolve as long as your param is a relative folder.
```

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+            this._initialize();
+        }
+    }
+
+    public isInitialized() {
+        return this._isInitialized;
+    }
+
+    private _initialize() {
+        this._isInitialized = true;
```
ramthi, line 51: Ideally, you would set this to true after all the initialization work is down.  There is a bunch of code and if anything fails and we bail halfway through you don't want isInitialized lying!

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        fs.readFile(snippetPath, function (err, snippet) {
+            if (err) {
+                Logging.warn("Failed to load AI Web snippet. Ex:" + err);
+            }
+            WebSnippet._snippet = snippet.toString().replace("INSTRUMENTATION_KEY", client.config.instrumentationKey);
+        });
+    }
+
+    public enable(isEnabled: boolean) {
```
ramthi, line 38: When do users call this public method?  Seems like you want to support "disable" functionality as well!  If so, there isn't much disabling going on other than setting the flag.  Also, I don't see this._isEnabled used anywhere so if someone sets this to false, what exactly will happen?

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Validate response and try to inject Web snippet
+     */
+    public static ValidateInjection(response: http.ServerResponse, input: string | Buffer): boolean {
+        if (response && input) {
+            // insert snippet if only response returns 200
+            if (response.statusCode != 200) return false;
+            let contentType = response.getHeader('Content-Type');
```
ramthi, line 94: mixed use of single-quotes and double-quotes.  Can you make the usage consistent across the project?  Check with Hector on what the standard is.

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+    private _isInitialized: boolean;
+
+    constructor(client: TelemetryClient) {
+        if (!!WebSnippet.INSTANCE) {
+            throw new Error("Web snippet injection should be configured from the applicationInsights object");
+        }
+
+        WebSnippet.INSTANCE = this;
+        // AI URL used to validate if snippet already included
+        WebSnippet._aiUrl = " https://js.monitor.azure.com/scripts/b/ai.2";
```
ramthi, line 24: we should probably check for a pattern to see if our snippet is already there.  There could be multiple versions used, isn't it?  @hectorhdzg can you chime in?

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+            let html = input.toString();
+            // Try to add script before HTML head closure
+            let index = html.indexOf("</head>");
+            if (index >= 0) {
+                let subStart = html.substring(0, index);
+                let subEnd = html.substring(index);
+                input = subStart + '<script type="text/javascript">' + WebSnippet._snippet + '</script>' + subEnd;
+                // Set headers
+                if (hasContentHeader) {
+                    response.setHeader("Content-Length", input.length.toString());
```
ramthi, line 129: do you have to convert the length to string?  setHeader takes <any> for value.

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Inject Web snippet
+     */
+    public static InjectWebSnippet(response: http.ServerResponse, input: string | Buffer): string | Buffer {
+        try {
+            let hasContentHeader = !!response.getHeader('Content-Length');
```
ramthi, line 114: hasContentLengthHeader would be a good name

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,146 @@
...
+                let subEnd = html.substring(index);
+                input = subStart + '<script type="text/javascript">' + WebSnippet._snippet + '</script>' + subEnd;
+                // Set headers
+                if (hasContentHeader) {
+                    response.setHeader("Content-Length", input.length.toString());
+                }
+            }
+        }
+        catch (ex) {
+            Logging.info("Failed to change content-lenght headers for JS injection. Exception:" + ex);
```
ramthi, line 134:
```suggestion
            Logging.info("Failed to inject JS Snippet. Exception:" + ex);
```

ramthi (changes requested):

AutoCollection/WebSnippet.ts
Karlie-777: Yes! If the instance is not found, it means the auto injection config is set to false or some errors happening during init!

AutoCollection/WebSnippet.ts
Karlie-777: will update this with npm when snippet is published

AutoCollection/WebSnippet.ts
Karlie-777: This will be used by applicationInsights.ts during init or reset process. it should be enable/disable like all other configs

AutoCollection/WebSnippet.ts
Karlie-777: updated in helper methods

AutoCollection/WebSnippet.ts
Karlie-777: updated

AutoCollection/WebSnippet.ts
Karlie-777: removed version numbers

AutoCollection/WebSnippet.ts
Karlie-777: updated

AutoCollection/WebSnippet.ts
Karlie-777: follow the previous original pattern [here](https://github.com/microsoft/ApplicationInsights-node.js/blob/033d1f4871ba9a5c2b136f98cebc7a21dd91ff37/AutoCollection/Exceptions.ts#L39)

AutoCollection/WebSnippet/fetchSnippet.js
```diff
@@ -0,0 +1,26 @@
+const path = require('path');
```
hectorhdzg, line 1: Remove this file as snippet is already part of the changes and will be checked in until npm package is available

Library/JsonConfig.ts
```diff
@@ -18,6 +18,7 @@ const ENV_noDiagnosticChannel = "APPLICATION_INSIGHTS_NO_DIAGNOSTIC_CHANNEL"
 const ENV_noStatsbeat = "APPLICATION_INSIGHTS_NO_STATSBEAT";
 const ENV_noHttpAgentKeepAlive = "APPLICATION_INSIGHTS_NO_HTTP_AGENT_KEEP_ALIVE";
 const ENV_noPatchModules = "APPLICATION_INSIGHTS_NO_PATCH_MODULES";
+const ENV_webSnippetEnable = "APPINSIGHTS_WEB_SNIPPET_ENABLED";
```
hectorhdzg, line 21: We have "APPLICATION_INSIGHTS", "APPLICATIONINSIGHTS" and now APPINSIGHTS, maybe stick to some of the other two, I recommend to use "APPLICATIONINSIGHTS_WEB_SNIPPET_ENABLED"

Library/JsonConfig.ts
```diff
@@ -130,6 +134,12 @@ export class JsonConfig implements IJsonConfig {
             if (jsonConfig.noPatchModules != undefined) {
                 this.noPatchModules = jsonConfig.noPatchModules;
             }
+            if (jsonConfig.enableAutoWebSnippetInjection != undefined) {
+                this.enableAutoWebSnippetInjection = jsonConfig.enableAutoWebSnippetInjection;
+            }
+            if (jsonConfig.isDebugWebSnippet !== undefined) {
```
hectorhdzg, line 140: I will expect enableAutoWebSnippetInjection to control this one as well, it is weird user add env variable to turn the feature off and debug snippet is still added, add extra check to avoid this

applicationinsights.ts
```diff
@@ -307,6 +318,33 @@ export class Configuration {
         return Configuration;
     }
 
+    /**
+     * Sets the state of Web snippet injection
+     * @param value if true Web snippet will be tried to be injected in server response
+     * @returns {Configuration} this class
+     */
+    public static setWebSnippetInjection(value: boolean) {
```
hectorhdzg, line 326: Is cleaner to have a single method to enable this and extra parameter to determine if debug version or not

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,293 @@
...
+        WebSnippet._aiUrl = "https://js.monitor.azure.com/scripts/b/ai";
+        WebSnippet._aiDeprecatedUrl = "https://az416426.vo.msecnd.net/scripts/b/ai"
+
+        //TODO: replace the path with npm package exports
+        let snippetPath = path.resolve(__dirname, "../../AutoCollection/snippet/snippet.min.js");
+        if (client.config.isDebugWebSnippet) {
+            snippetPath = path.resolve(__dirname, "../../AutoCollection/snippet/snippet.js");
+        }
+
+        fs.readFile(snippetPath, function (err, snippet) {
```
hectorhdzg, line 38: Add try catch to all file operation to avoid app crashing in case of error

AutoCollection/WebSnippet.ts
```diff
@@ -0,0 +1,293 @@
...
+            }
+          
+        }
+
+    }
+
+    /**
+     * Validate response and try to inject Web snippet
+     */
+    public static ValidateInjection(response: http.ServerResponse, input: string | Buffer): boolean {
```
hectorhdzg, line 196: You already handle singleton using INSTANCE, do you need all these methods as static?

AutoCollection/WebSnippet.ts
ramthi: Would it hurt to move this to end of the method?  If not I suggest you move it :)

Karlie-777: I assume putting `isInitialized` flag in the beginning would prevent double monkey patching. for example, when user enables the feature but initialization hasn't been finished and then user enabled it again, if isInitialized` flag is still be false at this time, the process will start a new monkey patch again. I am not sure if there is a better way to handle this scenario.😊

hectorhdzg (approved):
LGTM, please merge with latest changes in develop branch

ramthi (approved):

ramthi:
Lets Ship it!!

PR #895, opened 2022-01-30
Ignore github files in npm package

orgads:


.npmignore
```diff
@@ -7,6 +7,8 @@ obj
 *.sln
 *.njsproj
 .vscode/
+.github
+applicationinsights.json
```
hectorhdzg, line 11: We want this file to be part of npm package, it is available for configuration of the SDK

orgads: Done

hectorhdzg (approved):

PR #911, opened 2022-02-10
Statsbeat do not count failed request when throttle

hectorhdzg:


MSNev (approved):

Library/Sender.ts
```diff
@@ -197,11 +205,8 @@ class Sender {
                                 }
                             } else if (this._isRetriable(res.statusCode)) {
                                 try {
-                                    if (this._statsbeat) {
+                                    if (this._statsbeat && res.statusCode != 429) {
```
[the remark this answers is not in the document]
hectorhdzg: So, we send Statsbeat metrics for throttle and retry, @heyams can you confirm?

Library/Sender.ts
```diff
@@ -185,6 +185,14 @@ class Sender {
                         let endTime = +new Date();
                         let duration = endTime - startTime;
                         this._numConsecutiveFailures = 0;
+                        if (this._statsbeat) {
+                            if (res.statusCode == 429) { // Throttle
```
heyams, line 189: 439 is throttle

heyams: 429 is too many requests. they're different.

Library/Sender.ts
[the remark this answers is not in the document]
heyams: can you only add 439 to throttling count in Statsbeat and then count 429 as failure? that's what we do in java.

heyams:
@hectorhdzg here is breeze's definition of these 2 response codes: 
```
 **429 - Too many requests**
  The client send too many telemetries to the endpoint within in the given time interval so it is throttled

* **439 - Too many requests and refresh cache**
  The client is throttled at one point so frequently that the endpoint would send it 439 status code saying that it is throttled and should not retry for a long time.
Then, the client would drop the current data from its buffers.
```

heyams (approved):

Library/Sender.ts
```diff
@@ -15,8 +15,8 @@ import Util = require("./Util");
 import { URL } from "url";
 import Logging = require("./Logging");
 import { FileAccessControl } from "./FileAccessControl";
-import { Envelope } from "../Declarations/Contracts";
 
+const throttleStatusCode = 439; //  - Too many requests and refresh cache
```
ramthi, line 19: 439 is Quota exceeded NOT too many requests.  
429 would be too many requests hence throttled.  Please see https://msazure.visualstudio.com/One/_git/Mgmt-AppInsights-Breeze?path=/Specifications/EndpointRequestResponse.md&_a=preview&version=GBmaster

Also, noticed that 439 is a legacy code.  They want new SDKs to use 402

heyams:
@hectorhdzg let's chat tomorrow if we want to send two count to breeze.. daily quota exceeded vs. monthly quota exceeded.

PR #927, opened 2022-03-23
Strip http[s] request credentials

### LINKED ISSUE #926: Credentials on basic authentication appear on the path
When using `axios`:

```javascript
axios({
  url: 'http://server/path',
  auth: {
    username: 'user',
    password: 's3cr37',
  }
});
```

The request dependency entry is logged as `http://user:s3cr37@server/path`, which is unsafe.

So far I found that `_getUrlFromRequestOptions` calls `url.format`, which includes the credentials if `auth` exists.

Is this intended? Is it possible to call `url.format` with `auth` stripped?

orgads:
Fixes #926

### DIFF
--- a/AutoCollection/HttpDependencyParser.ts
+++ b/AutoCollection/HttpDependencyParser.ts
@@ -137,7 +137,7 @@ class HttpDependencyParser extends RequestParser {
                 }
             }
         } else if (options && typeof url.URL === "function" && options instanceof url.URL) {
-            return url.format(options);
+            return url.format(options, { auth: false });
         } else {
             // Avoid modifying the original options object.
             let originalOptions = options;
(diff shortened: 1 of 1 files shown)


hectorhdzg:
Thanks for your contribution @orgads, we cannot change the URL for everyone like proposed change, people could be expecting URL to be in specific format and having telemetry processors on top modifying it, this could be potentially a breaking change, I understand the concern and my initial suggestion would be to try to update the URL using telemetry processors and remove sensitive information in there, other approach similar to what you are doing would be to expose the entire options object for url.format as configuration of the SDK so people can update the defaults if they explicitly want to, allowing them to change other properties like removing query parameters that are also place for sensitive data in some cases.

orgads:
I see your concerns. This change was intended to be RFC anyway. I stripped the data in telemetry processor in my application.

I think that a flag can be useful. Notice that options for `url.format` are not enough, since they are not used when passing an object (that's the case with axios).

PR #931, opened 2022-03-29
Updating usage example in README.md

napernik:
Updating Azure Function integration example:
- fixing compatibility with AZF which have function result mapped to the HTTP response
- making sure that tracked requests have correct start time (previously end time was registered as start time)


hectorhdzg (approved):
LGTM, thanks for your contribution.

Issue #938, opened 2022-04-01
"ApplicationInsights:Failed to load AI Web snippet" in webpack bundled node.js project


rcbevans:
Change #683 breaks webpack bundling of application insights.

```ts
        //TODO: replace the path with npm package exports
        let snippetPath = path.resolve(__dirname, "../../AutoCollection/snippet/snippet.min.js");
```

https://github.com/microsoft/ApplicationInsights-node.js/commit/c4b5d3201afe6d81bb609e5f0c6371938ab9dfce#r70249203.

When bundling application insights as part of a node.js based application, `__dirname` gets resolved at runtime to be the path on disk of the bundled application, NOT the path of `AutoCollection/WebSnippet.ts` in `node_modules`.  The resolved relative path of `../../AutoCollection/snippet/snippet.min.js` is incorrect causing the application to fail to start with error

```
ApplicationInsights:Failed to load AI Web snippet. Ex:Error: ENOENT: no such file or directory, open 'project/packages/AutoCollection/snippet/snippet.min.js' []
TypeError: Cannot read properties of undefined (reading 'toString')
    at ReadFileContext.eval [as callback] (webpack://project-server/./node_modules/applicationinsights/out/AutoCollection/WebSnippet.js?:26:47)
    at FSReqCallback.readFileAfterOpen [as oncomplete] (node:fs:314:13)
```

Workarounds are to exclude applicationinsights from bundling and ship alongside the bundled application, or pin applicationinsights to [version 2.2.2](https://www.npmjs.com/package/applicationinsights/v/2.2.2)

hectorhdzg:
We will be publishing version 2.3.1 with the fix today

rcbevans:
Thanks for the quick turn around on this one!

PR #939, opened 2022-04-02
fix injection error in bundle

Karlie-777:
Quick fix on error in bundling ApplicationInsights.
Should be completely fix when snippet NPM is published.


Karlie-777:
https://github.com/microsoft/ApplicationInsights-node.js/issues/938

hectorhdzg (approved):

PR #944, opened 2022-04-08
[Preview] Adding handlers and exporters for metrics/logs

hectorhdzg:
Removing unnecessary static classes/methods
Removed old contracts
Removed auto instrumentation code that will be handled by OpenTelemetry
Using swagger autogenerated classes for Breeze types
Changed test folder structure


hectorhdzg:
Several TODOs and enablement of tests will come after

Issue #930, opened 2022-03-29
.setAutoCollectDependencies(false) still collects dependencies


MatejSkrbis:
I've updated the application insights package from **1.8.10** to **2.2.2**

Since the update the dependencies are getting collected even when `.setAutoCollectDependencies(false)` is set on setup.
Is there any other option to turn this off in newer version that I missed?

This function worked correctly before the update.

hectorhdzg:
Can you describe your environment a little more? is possible your dependencies are coming from different place; can you take a look at SDKVersion field in the unexpected telemetry and share it?

MatejSkrbis:
I'm using node 14, windows server 2016.
The dependencies started coming the moment I deployed the code to upgrade insights package from **1.8.10** to **2.2.2**. There were no dependencies in insights before that. There were only exceptions and some custom events which we logged.

SDK field in insights says **node:2.2.2**
The **application_Version** is matching with the latest package version which we automatically increase on commit. So it must be from our latest update.

BenjaminBertran:
Hi,
I confirm this behavior if these calls are made on a client:
- setup(key)
- dispose
- again setup(key)

Looking at the code, the variable isEnabled in HttpDependencies is not used [here](
https://github.com/microsoft/ApplicationInsights-node.js/blob/c060888eb97d6a8d55366ebbfc49243f4f2794c4/AutoCollection/HttpDependencies.ts#L79).
... && _this.isEnabled) should be added.
In the meantime, "interceptors" should be removed on http(s).request in dispose() method.

Apokalypt:
I agree with Benjamin's comment and I am able to reproduce the problem with the calls he indicated.

The developer's choice about whether or not to track dependencies is not really taken into account since the variable containing this information is only read before initialization. If this variable changes over time, the [trackRequest ](https://github.com/microsoft/ApplicationInsights-node.js/blob/c060888eb97d6a8d55366ebbfc49243f4f2794c4/AutoCollection/HttpDependencies.ts#L140) function seems to continue tracking requests regardless of the developer's choice.

[...]

PR #946, opened 2022-04-12
Ensure dependencies are not tracked if disabled after initialization

hectorhdzg:
Related to #930

### DIFF
--- a/AutoCollection/HttpDependencies.ts
+++ b/AutoCollection/HttpDependencies.ts
@@ -60,48 +60,54 @@ class AutoCollectHttpDependencies {
         const originalHttpsRequest = https.request;
 
         const clientRequestPatch = (request: http.ClientRequest, options: string | URL | http.RequestOptions | https.RequestOptions) => {
-            var shouldCollect = !(<any>options)[AutoCollectHttpDependencies.disableCollectionRequestOption] &&
-                !(<any>request)[AutoCollectHttpDependencies.alreadyAutoCollectedFlag];
+            try {
+                var shouldCollect = !(<any>options)[AutoCollectHttpDependencies.disableCollectionRequestOption] &&
+                    !(<any>request)[AutoCollectHttpDependencies.alreadyAutoCollectedFlag];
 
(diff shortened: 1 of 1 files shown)


AutoCollection/HttpDependencies.ts
```diff
@@ -74,29 +74,32 @@ class AutoCollectHttpDependencies {
...
-                // If there is no context create one, this apply when no request is triggering the dependency
-                if (!CorrelationContextManager.getCurrentContext()) {
-                    // Create correlation context and wrap execution
-                    let operationId = null;
-                    if (CorrelationIdManager.w3cEnabled) {
-                        let traceparent = new Traceparent();
-                        operationId = traceparent.traceId;
+                if (this._isEnabled) {
+                    // Mark as auto collected
+                    (<any>request)[AutoCollectHttpDependencies.alreadyAutoCollectedFlag] = true;
```
MSNev, line 81: Note: for browsers there has been at least 1 environment which doesn't allow "adding" properties to the request, which throws. So you might want to wrap this in a try / catch to allow the telemetry to still be sent. Web also checks the URL

MSNev (approved):

AutoCollection/HttpDependencies.ts
```diff
@@ -60,48 +60,54 @@ class AutoCollectHttpDependencies {
...
-                        // Create correlation context and wrap execution
-                        let operationId = null;
-                        if (CorrelationIdManager.w3cEnabled) {
-                            let traceparent = new Traceparent();
-                            operationId = traceparent.traceId;
+                if (request && options && shouldCollect) {
+                    CorrelationContextManager.wrapEmitter(request);
+                    if (this._isEnabled) {
+                        // Mark as auto collected
+                        (<any>request)[AutoCollectHttpDependencies.alreadyAutoCollectedFlag] = true;
```
MSNev, line 82: It was really just this line I was thinking about wrapping -- but that could cause repetitive requests. So this works too.

MSNev (approved):

Issue #945, opened 2022-04-12
Need clarification about the dispose() method


Apokalypt:
After looking at the code to understand a bug mentioned in another issue I found some strange things about the operation of the [dispose()](https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/applicationinsights.ts#L382) method.
I would like to understand its precise objective :
- Should this function reset the various properties to a default value so that the various methods ( like [setAutoCollectExceptions](https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/applicationinsights.ts#L84) ) can be recalled to redefine the property values?
- Should this function stop all "watches" ( track request, track dependency calls, ... ) in order to allow a new instantiation of each property ( _console, _exceptions, _performance, ... ) without edge effects / duplicates?

I ask this because currently the description of the function suggests that it only resets the properties to default values when in fact it prepares for a new instantiation (it sets [defaultClient to null](https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/applicationinsights.ts#L384) and therefore it is necessary to call [setup()](https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/applicationinsights.ts#L84) for the monitoring to work again except that in this function we create a new instance for each property ( _console, _exceptions, _performance, ... ))

hectorhdzg:
Should stop all automatic tracking and remove all telemetry processors that were added, setup need to be called again after calling dispose, not sure what you mean by "watches" we patch a lot of code and dispose is not actually removing these patches, we should not generate telemetry though(#946), about the actual internal implementation, there is a bunch of static classes so dispose doesn't mean we are actually cleaning up everything from memory

Apokalypt:
> not sure what you mean by "watches"
I've called "watches" : tracking and telemetry processor ( I didn't know how to name both at the same time 😂 )

[...]

PR #947, opened 2022-04-13
fix(AutoCollectHttpRequests): Http requests aren't collect if we call dispose+setup after http.createServer

Apokalypt:
### **- What is the problem?**
Currently, every time an instance of AutoCollectHttpRequests is created, we overwrite the http(s).createServer function to add a handler to track HTTP(S) requests. As a result, if we call the constructor several times we end up with several handlers that can duplicate data.
The second problem is when we initialize the library once with setup() and then create the server with http(s).createServer. Here we track the requests correctly but if, for X reasons, we call dispose() and then setup(), the HTTP requests will not be tracked anymore because the code waits for a call to createServer to add the instance handler (despite the fact that the instance created at the first setup call already had one in place)

### **- How is it corrected?**
As the code allows to have only one instance of AutoCollectHttpRequests at the same time ( stored in the static variable INSTANCE ), it is not necessary to modify the code of http(s).createServer each time the constructor is called. It is necessary to do it only once and in the added handler, we call a method of the instance stored in the static variable to register the request.
Thus, after the call to dispose() the registrations will stop but will resume after the call to setup despite the change of instance.

__ __

_The fix is not optimized. In fact, this kind of handler should not be added when calling a constructor but rather at the very beginning of the library to optimize the chances that the library is well instantiated by the developers. However, after a discussion with @hectorhdzg ( #945 ) it turns out that this part of the code is going to change soon, so I didn't want to dwell on a code that will soon be completely overwritten._

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -13,7 +13,7 @@ import AutoCollectPerformance = require("./Performance");
 class AutoCollectHttpRequests {
 
     public static INSTANCE: AutoCollectHttpRequests;
-    public static HANDLER_READY: boolean = false;
+    private static HANDLER_READY: boolean = false;
 
     private static alreadyAutoCollectedFlag = "_appInsightsAutoCollected";
 
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -1,6 +1,5 @@
 import http = require("http");
 import https = require("https");
-import url = require("url");
 
 import Contracts = require("../Declarations/Contracts");
 import TelemetryClient = require("../Library/TelemetryClient");
@@ -14,6 +13,7 @@ import AutoCollectPerformance = require("./Performance");
 class AutoCollectHttpRequests {
 
     public static INSTANCE: AutoCollectHttpRequests;
+    public static HANDLER_READY: boolean = false;
 
     private static alreadyAutoCollectedFlag = "_appInsightsAutoCollected";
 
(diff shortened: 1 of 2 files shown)


AutoCollection/HttpRequests.ts
```diff
@@ -75,9 +75,38 @@ class AutoCollectHttpRequests {
...
+                onRequest(request, response);
+            }
+        });
+    }
+
     private _initialize() {
         this._isInitialized = true;
 
+        // Avoid the creation of multiple handler on http(s).createServer
+        if (AutoCollectHttpRequests.HANDLER_READY) {
```
hectorhdzg, line 104: You can use _isInitialized , no need to add new flag for the same thing, do not clean it on dispose

AutoCollection/HttpRequests.ts
```diff
@@ -14,6 +13,7 @@ import AutoCollectPerformance = require("./Performance");
 class AutoCollectHttpRequests {
 
     public static INSTANCE: AutoCollectHttpRequests;
+    public static HANDLER_READY: boolean = false;
```
hectorhdzg, line 16: I'm trying to remove overuse of static stuff in this SDK, no need to expose this one, just leave this one as private not static property here

Apokalypt: Unfortunately we need a static variable to know if the handler is set on the createServer function.

.setup() creates a new instance of AutoCollectHttpRequests even if an instance currently exists. This means that if there is no static variable to say that the handler has been created beforehand a second handler will be added to the createServer function. The problem that will occur, without a static variable to tell if the handler has been created, is if the user makes a call to .setup() then .dispose()+.setup() then .dispose()+.setup(), ... and after call http.createServer. With this scenario, these requests will be recorded several times :/

Apokalypt: What can be done is to change the behavior of the .setup() function.
- If _serverRequests is null we create an instance
- If _serverRequests is not null we update the _client variable

Do you think it would be better?

hectorhdzg: I see, leaving as static should be fine then, just leave it as private, also looks like http basic usage tests are [failing](https://github.com/microsoft/ApplicationInsights-node.js/runs/6040472929?check_suite_focus=true) in EndToEnd test class because of this change.

Apokalypt: Yes, I saw it. I'm going to watch it tonight to try to figure out what's going on

Apokalypt:
I don't understand why tests failed... They fail on a part that I did not modify and on top of that, they pass in local setup...

hectorhdzg (approved):
Thanks for your contribution

PR #952, opened 2022-04-21
iKey env variable deprecation

hectorhdzg:
Add warning when iKey is present and connection string is not available
Remove direct usage of iKey env variable for diagnostics

### DIFF
--- a/Bootstrap/DiagnosticLogger.ts
+++ b/Bootstrap/DiagnosticLogger.ts
@@ -18,14 +18,16 @@ export class DiagnosticLogger {
             language: "nodejs",
             operation: "Startup",
             siteName: process.env.WEBSITE_SITE_NAME,
-            ikey: process.env.APPINSIGHTS_INSTRUMENTATIONKEY,
+            ikey: "unknown",
             extensionVersion: process.env.ApplicationInsightsAgent_EXTENSION_VERSION,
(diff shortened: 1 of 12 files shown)


lzchen (approved):

Bootstrap/DiagnosticLogger.ts
```diff
@@ -18,14 +18,16 @@ export class DiagnosticLogger {
             language: "nodejs",
             operation: "Startup",
             siteName: process.env.WEBSITE_SITE_NAME,
-            ikey: process.env.APPINSIGHTS_INSTRUMENTATIONKEY,
+            ikey: "unknown",
```
MSNev, line 21: I thought that while the ikey usage has been deprecated, the documentation says it will continue to work.... At least until 2025...

MSNev: Found the code below for ENV_legacyInstrumentationKey

MSNev (approved):

Bootstrap/DiagnosticLogger.ts
hectorhdzg: Yeah we always populate this one using whatever is returned from Config, that one will read all environment variables with correct priority and log warnings if needed

PR #956, opened 2022-04-26
Turn off Statsbeat if not able to connect to endpoint

hectorhdzg:


MSNev (approved):

AutoCollection/Statsbeat.ts
```diff
@@ -54,7 +54,7 @@ class Statsbeat {
         let statsbeatConnectionString = this._getConnectionString(config);
         this._statsbeatConfig = new Config(statsbeatConnectionString);
         this._statsbeatConfig.samplingPercentage = 100; // Do not sample
-        this._sender = new Sender(this._statsbeatConfig, null, null, this._handleNetworkError);
+        this._sender = new Sender(this._statsbeatConfig, null, null, null, null, true, this._shutdownStatsbeat);
```
heyams, line 57: parameters are getting jammed. how do you handle that in node? what is the limit or keep growing?

Library/Sender.ts
```diff
@@ -59,6 +64,9 @@ class Sender {
         this._fileCleanupTimer = null;
         // tmpdir is /tmp for *nix and USERDIR/AppData/Local/Temp for Windows
         this._tempDir = path.join(os.tmpdir(), Sender.TEMPDIR_PREFIX + this._config.instrumentationKey);
+        this._isStatsbeatSender = isStatsbeatSender || false;
+        this._shutdownStatsbeat = shutdownStatsbeat;
+        this._failedToIngestCounter = 0;
```
heyams, line 69: line 36 is default to 0, right? line 69 might be redundant? at least that's what Java does.

AutoCollection/Statsbeat.ts
hectorhdzg: When the object starts doing stuff the moment is instantiated, we need to provide the settings in the constructor, it could receive a single "settings" object with everything instead of multiple required and optional parameters, that is what I prefer to do but that will be a breaking change so we cannot do it right now, customer could be instantiating this class directly so we cannot change the signature now.

Library/Sender.ts
hectorhdzg: Default value would be undefined in this case even if we set the type as number

Library/Sender.ts
```diff
@@ -327,31 +347,50 @@ class Sender {
...
+    }
+
+    private _logWarn(message?: any, ...optionalParams: any[]) {
+        if (!this._isStatsbeatSender) {
+            Logging.warn(Sender.TAG, message, optionalParams);
+        }
+    }
+
+    private _statsbeatFailedToIngest() {
+        this._failedToIngestCounter++;
```
heyams, line 363: should you only increment when _shutdownStatsbeat is not null? what if customer failed 3 times, then statsbeat failed, shutdown is not null, it will shutdown on 1st occurrence.

heyams (approved):

PR #969, opened 2022-05-16
[Preview] Automatic tracking of HTTP through OpenTelemetry

hectorhdzg:


src/library/handlers/metricHandler.ts
```diff
@@ -72,8 +69,8 @@ export class MetricHandler {
         this._nativePerformance.enable(this.isNativePerformance, this.disabledExtendedMetrics);
     }
 
-    public flush(options?: FlushOptions) {
-        this._batchProcessor.triggerSend(options ? options.isAppCrashing : false);
+    public flush(isAppCrashing?: boolean) {
```
MSNev, line 72: Personally, I like the previous implementation (passing a FlushOptions) options object as it allows for any future expansion without the need to add additional arguments -- even though you are only using the one today.

eg. in the future we might want to add things like timeout, retry, synchronous etc

test/unitTests/fixtures/server-key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
MSNev, line 1: This will probably cause an issue with secret scanning -- can this be auto generated as part of the build so it doesn't need to be checked in? or is this required to send it somewhere.

MSNev (approved):

src/library/handlers/traceHandler.ts
```diff
@@ -95,42 +101,46 @@ export class TraceHandler {
...
+    }
+
+    public shutdown() {
+        this._spanProcessor.shutdown();
+        this._exporter.shutdown();
     }
 
     // Support Legacy APIs
     public trackRequest(telemetry: Contracts.RequestTelemetry) {
+        let startTime = Date.now();
```
lzchen, line 115: Isn't the `startTime` supposed to be before the actual request has been made, not when `trackRequest` is called?

src/library/handlers/traceHandler.ts
```diff
@@ -145,42 +155,51 @@ export class TraceHandler {
...
         };
-        if (telemetry.dependencyTypeName.toLowerCase().indexOf("http") > -1) {
-            attributes[SemanticAttributes.HTTP_METHOD] = "http";
-            try {
-                let url = new URL(telemetry.data);
-                attributes[SemanticAttributes.HTTP_METHOD] = url.protocol;
-            } catch (error) {
-                // Ignore error
+        if (telemetry.dependencyTypeName) {
+            if (telemetry.dependencyTypeName.toLowerCase().indexOf("http") > -1) {
```
lzchen, line 162: This is kind of strange to me. Are we mapping from `DependencyTelemetry` (AppInsights) to OpenTelemetry spans and then back to `DependentTelemetry` when we export using the Azure monitor exporter?

src/library/handlers/traceHandler.ts
lzchen: Not a blocking comment, but maybe we can let the user pass in the `startTime` as part of the `RequestTelemetry`. This way, the `startTime` could potentially represent when the actual start time of the request was made.

lzchen (approved):

src/library/handlers/traceHandler.ts
hectorhdzg: That is correct, that would be the only way to allow a manual API right now

PR #971, opened 2022-05-26
Adding tests for LogHandler and MetricHandler

hectorhdzg:
Fixing issues found while testing the code


src/library/handlers/shared/batchProcessor.ts
```diff
@@ -80,6 +78,7 @@ export class BatchProcessor {
             this._buffer = [];
             clearTimeout(this._timeoutHandle);
             this._timeoutHandle = null;
+            resolve();
```
MSNev, line 81: Won't this resolve immediately, even if the exporter is waiting for a response?

I was thinking that the previous "resolve" was correct, but there should also be an `else { resolve(); }`.

MSNev (approved):

PR #981, opened 2022-06-10
[Preview] Add correlation in logs

hectorhdzg:


src/library/handlers/logHandler.ts
```diff
@@ -330,6 +322,16 @@ export class LogHandler {
...
+            tags[KnownContextTagKeys.AiInternalSdkVersion] = resourceManager.getInternalSdkVersion();
+        }
+        // Add Correlation headers
+        const spanContext = trace.getSpanContext(context.active());
+        if (spanContext) {
+            tags[KnownContextTagKeys.AiOperationId] = spanContext.traceId;
+            tags[KnownContextTagKeys.AiOperationParentId] = spanContext.spanId;
+        }
+        else{
+            tags[KnownContextTagKeys.AiOperationId] = this._idGenerator.generateTraceId();
```
lzchen, line 334: Python sets trace_Id to 0 if not present in SpanContext. Not sure if this what we want (because this behavior is not well-defined), but we should probably be consistent.

test/functionalTests/runner/taskExpectations.js
```diff
@@ -255,63 +247,63 @@ module.exports = {
...
-            telemetry.data.baseData.severityLevel === 3;
+            telemetry.data.baseData.severityLevel === "Error";
         }
     ),
     "WinstonWarn2": outputContract(
         "MessageData",
         (telemetry) => {
             return telemetry.data.baseData.message === "test warn" &&
-            telemetry.data.baseData.severityLevel === 2;
+            telemetry.data.baseData.severityLevel === "Warning";
```
MSNev, line 299: All these string might be better defined as constants somewhere, instead of repeating

hectorhdzg: The actual source code uses an enumerator "KnownSeverityLevel", I usually prefer to hardcode the values in the tests to ensure if the enumerator ever change the tests are affected, but I'm fine either way

src/library/handlers/logHandler.ts
MSNev: The JS Sdk generates a new one (like this) when the provided trace id is not provided or invalid (when it's being set).

MSNev (approved):

src/library/handlers/logHandler.ts
hectorhdzg: is there anything special that happens on Breeze side with the 0 value?

src/library/handlers/logHandler.ts
```diff
@@ -330,6 +322,16 @@ export class LogHandler {
             }
             const serviceInstanceId = attributes[SemanticResourceAttributes.SERVICE_INSTANCE_ID];
             tags[KnownContextTagKeys.AiCloudRoleInstance] = String(serviceInstanceId);
+            tags[KnownContextTagKeys.AiInternalSdkVersion] = resourceManager.getInternalSdkVersion();
+        }
+        // Add Correlation headers
+        const spanContext = trace.getSpanContext(context.active());
+        if (spanContext) {
+            tags[KnownContextTagKeys.AiOperationId] = spanContext.traceId;
```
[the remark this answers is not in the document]
hectorhdzg: We are not converting the values in the exporter as well; Breeze doesn't support Common Schema 4 so I don't believe we need to do it for now.

src/library/handlers/logHandler.ts
lzchen: I do not believe so, just know that this is how Opencensus is doing this. Perhaps generating a new one is the way to go but am not certain of the implications.

lzchen:
Approving contingent of the hex code change.

lzchen (approved):

PR #989, opened 2022-07-13
[Preview] Integrating with OpenTelemetry Metrics

hectorhdzg:
Using Azure Monitor Metric Exporter
Using Metrics SDK to handle perf counters, native metrics and hearbeat
Pre aggregated metrics and Statsbeat would be migrated in future PRs


src/autoCollection/metrics/heartBeat.ts
```diff
@@ -0,0 +1,91 @@
...
+        };
+        this._metricReader = new PeriodicExportingMetricReader(metricReaderOptions);
+        this._meterProvider.addMetricReader(this._metricReader);
+        this._meter = this._meterProvider.getMeter("ApplicationInsightsHeartBeatMeter");
+        this._metricGauge = this._meter.createObservableGauge(HeartBeatMetricName);
+        this._metricGauge.addCallback(this._trackHeartBeat);
+    }
+
+    public enable(isEnabled: boolean) {
+        // TODO: Allow enable/disable functionality?
```
MSNev, line 50: Yes, I believe this lets users enable / disable the metric

src/autoCollection/metrics/nativePerformance.ts
```diff
@@ -0,0 +1,192 @@
...
+     * Start instance of native metrics agent.
+     *
+     * @param {boolean} isEnabled
+     * @memberof AutoCollectNativePerformance
+     */
+    public enable(
+        isEnabled: boolean,
+        disabledMetrics: IDisabledExtendedMetrics = {}
+    ): void {
+        if (this._metricsAvailable == undefined && isEnabled && !this._isInitialized) {
```
MSNev, line 46: Probably, should check that it's not already enabled, just in case someone calls this twice.

A issue came up in the JS Sig Bug triage yesterday where this can occur and break stuff -- it got tagged as a P3 to fix at some point.

MSNev: Oh, I see your using the _isInitialized for that...

src/autoCollection/metrics/nativePerformance.ts
```diff
@@ -0,0 +1,192 @@
...
+            }
+        }
+        this._isEnabled = isEnabled;
+        this._disabledMetrics = disabledMetrics;// TODO: Use to filter out metrics in View
+        if (this._isEnabled && !this._isInitialized) {
+            this._isInitialized = true;
+        }
+
+        // Enable the emitter if we were able to construct one
+        if (this._isEnabled && this._emitter) {
```
MSNev, line 66: I guess from the comment above this could be entered more than once.

src/library/azureVirtualMachine.ts
```diff
@@ -17,7 +17,7 @@ export interface IVirtualMachineInfo {
 export class AzureVirtualMachine {
     private _TAG = "AzureVirtualMachine";
 
-    public getAzureComputeMetadata(config: Config, callback: (vm: IVirtualMachineInfo) => void) {
+    public async getAzureComputeMetadata(config: Config): Promise<IVirtualMachineInfo> {
```
MSNev, line 20: I don't think you need to tag as async because your not using await

MSNev (approved):
With suggestions

MSNev (approved):

src/library/azureVirtualMachine.ts
hectorhdzg: Changed to use await as well

PR #1002, opened 2022-08-11
[Preview] Adding support for Standard Metrics

hectorhdzg:
Added duration metrics for server/client in internal MetricsInstrumentation
Exceptions and traces standard metrics continue to work as before
Added SpanProcessor to add extra properties in spans


src/autoCollection/metrics/httpMetricsInstrumentation.ts
```diff
@@ -51,20 +65,28 @@ export class HttpMetricsInstrumentation extends InstrumentationBase {
     return [this._getHttpDefinition(), this._getHttpsDefinition()];
   }
 
-  private _htppRequestDone(metric: IHttpMetric) {
+  private _httpRequestDone(metric: IHttpStandardMetric) {
```
lzchen, line 68: Curious as to why it is named `IHttpStandardMetric` and not `ISpanStandardMetric`? We will probably need this for other standard metric signals from different types like db and messaging.

src/autoCollection/metrics/httpMetricsInstrumentation.ts
```diff
@@ -51,20 +65,28 @@ export class HttpMetricsInstrumentation extends InstrumentationBase {
...
     // Done could be called multiple times, only process metric once
     if (!metric.isProcessed) {
+      metric.isProcessed = true;
```
lzchen, line 71: Curious as to why `_httpRequestDone` is called multiple times for the same request?

src/library/handlers/metricHandler.ts
```diff
@@ -60,17 +57,27 @@ export class MetricHandler {
     private _isStarted = false;
     private _batchProcessor: BatchProcessor;
     private _performance: AutoCollectPerformance;
-    private _preAggregatedMetrics: AutoCollectPreAggregatedMetrics;
+    private _preAggregatedMetrics: AutoCollectStandardMetrics;
```
lzchen, line 60: `_standardMetrics` ?

src/library/handlers/logHandler.ts
```diff
@@ -144,6 +164,20 @@ export class LogHandler {
...
+                let exceptionDimensions: IMetricExceptionDimensions = {
+                    cloudRoleInstance: envelope.tags[KnownContextTagKeys.AiCloudRoleInstance],
+                    cloudRoleName: envelope.tags[KnownContextTagKeys.AiCloudRole]
+                };
+                this._metricHandler.getStandardMetricsCollector().countTrace(exceptionDimensions);
+                // Mark envelope as processed
+                const exceptionData: TelemetryExceptionData = (envelope.data as any).baseData;
+                exceptionData.properties = {
+                    ...exceptionData.properties,
+                    "_MS.ProcessedByMetricExtractors": "(Name:'Exceptions', Ver:'1.1')"
```
lzchen, line 178: Doesn't this property need to be set BEFORE the envelope is called `send`?

src/library/handlers/logHandler.ts
```diff
@@ -144,6 +164,20 @@ export class LogHandler {
         try {
             const envelope = this._exceptionToEnvelope(telemetry, this.config.instrumentationKey);
             this._batchProcessor.send(envelope);
+
+            if (this._metricHandler?.isStandardMetricsEnabled) {
+                let exceptionDimensions: IMetricExceptionDimensions = {
+                    cloudRoleInstance: envelope.tags[KnownContextTagKeys.AiCloudRoleInstance],
+                    cloudRoleName: envelope.tags[KnownContextTagKeys.AiCloudRole]
+                };
+                this._metricHandler.getStandardMetricsCollector().countTrace(exceptionDimensions);
```
lzchen, line 173: `countException`?

src/library/handlers/logHandler.ts
```diff
@@ -144,6 +164,20 @@ export class LogHandler {
         try {
             const envelope = this._exceptionToEnvelope(telemetry, this.config.instrumentationKey);
             this._batchProcessor.send(envelope);
+
+            if (this._metricHandler?.isStandardMetricsEnabled) {
+                let exceptionDimensions: IMetricExceptionDimensions = {
+                    cloudRoleInstance: envelope.tags[KnownContextTagKeys.AiCloudRoleInstance],
+                    cloudRoleName: envelope.tags[KnownContextTagKeys.AiCloudRole]
```
lzchen, line 171: Another field `client/type` might be needed. Curious did you look into what this dimension represents?

src/autoCollection/metrics/httpMetricsInstrumentation.ts
```diff
@@ -2,24 +2,27 @@
...
+import { SpanKind } from '@opentelemetry/api';
+import { SemanticAttributes } from '@opentelemetry/semantic-conventions';
 
 
-export class HttpMetricsInstrumentation extends InstrumentationBase {
+export class HttpMetricsInstrumentation extends InstrumentationBase<Http> {
 
-  private static _instance: HttpMetricsInstrumentation;
   private _nodeVersion: string;
   public totalRequestCount: number = 0;
```
lzchen, line 27: Why do we need to keep track of these totals? Are you planning to replace [performance counters](https://github.com/microsoft/ApplicationInsights-node.js/blob/develop/AutoCollection/Performance.ts) with this instrumentation?

hectorhdzg: Perfomance counters are currently calculated using these, is still not very clear how perf counters would be handled in OpenTelemetry so having this functionality in the meantime

src/library/handlers/logHandler.ts
hectorhdzg: I was not aware of the existence of that dimension, currently only enabling what we had before

src/autoCollection/metrics/httpMetricsInstrumentation.ts
hectorhdzg: It could be called multiple times for specific cases, like request timeout, abort, etc.

hectorhdzg:
Exporter related changes: https://github.com/Azure/azure-sdk-for-js/pull/22886

PR #1004, opened 2022-08-19
[Preview] Refactor Metric handlers

hectorhdzg:
Create specific classes receiving a Meter to create an update Instruments, same design as OpenTelemetry packages like HTTP instrumentation or hostMetrics.
Create specific handlers for Standard, Live Metrics and PerfCounters using Views to filter instruments.


hectorhdzg:
Will continue working with Live Metrics in different PR, adding views and handlers tests

PR #1007, opened 2022-08-25
Add statusCode and exceptionType Fields to Network Statsbeat

JacksonWeber:
This adds the statusCode and exceptionType fields to the appropriate network StatsBeat metrics per the spec https://github.com/microsoft/Telemetry-Collection-Spec/blob/main/ApplicationInsights/statsbeat.md

### DIFF
--- a/AutoCollection/Statsbeat.ts
+++ b/AutoCollection/Statsbeat.ts
@@ -44,8 +44,6 @@ class Statsbeat {
     private _attach: string = Constants.StatsbeatAttach.sdk; // Default is SDK
     private _feature: number = Constants.StatsbeatFeature.NONE;
     private _instrumentation: number = Constants.StatsbeatInstrumentation.NONE;
-    private _exceptionType: string;
-    private _statusCode: number;
 
     constructor(config: Config, context?: Context) {
         this._isInitialized = false;
@@ -298,7 +296,6 @@ class Statsbeat {
                 commonProperties
             );
             if (currentCounter.totalSuccesfulRequestCount > 0) {
-                properties = Object.assign({ ...properties, statusCode: this._statusCode });
                 this._statbeatMetrics.push({
                         name: Constants.StatsbeatCounter.REQUEST_SUCCESS,
                         value: currentCounter.totalSuccesfulRequestCount,
--- a/AutoCollection/Statsbeat.ts
+++ b/AutoCollection/Statsbeat.ts
@@ -126,6 +126,9 @@ class Statsbeat {
         counter.totalRequestCount++;
         counter.intervalRequestExecutionTime += duration;
         if (success === false) {
+            if (!statusCode) {
+                return;
+            }
             let currentStatusCounter = counter.totalFailedRequestCount.find((statusCounter) => statusCode === statusCounter.statusCode);
             if (currentStatusCounter) {
                 currentStatusCounter.count++;
--- a/AutoCollection/Statsbeat.ts
+++ b/AutoCollection/Statsbeat.ts
@@ -140,16 +140,16 @@ class Statsbeat {
         }
     }
 
-    public countException(endpoint: number, host: string, exceptionType: string) {
+    public countException(endpoint: number, host: string, exceptionType: Error) {
         if (!this.isEnabled()) {
             return;
         }
         let counter: Network.NetworkStatsbeat = this._getNetworkStatsbeatCounter(endpoint, host);
-        let currentErrorCounter = counter.exceptionCount.find((exceptionCounter) => exceptionType === exceptionCounter.exceptionType);
+        let currentErrorCounter = counter.exceptionCount.find((exceptionCounter) => exceptionType.name === exceptionCounter.exceptionType);
         if (currentErrorCounter) {
             currentErrorCounter.count++;
         } else {
-            counter.exceptionCount.push({ exceptionType: exceptionType, count: 1 });
+            counter.exceptionCount.push({ exceptionType: exceptionType.name, count: 1 });
         }
     }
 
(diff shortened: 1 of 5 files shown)


AutoCollection/Statsbeat.ts
```diff
@@ -286,42 +306,45 @@ class Statsbeat {
...
-                this._statbeatMetrics.push({
-                    name: Constants.StatsbeatCounter.REQUEST_FAILURE,
-                    value: currentCounter.totalFailedRequestCount,
-                    properties: properties
+            if (currentCounter.totalFailedRequestCount.length > 0) {
+                currentCounter.totalFailedRequestCount.forEach((currentCounter) => {
+                    this._statbeatMetrics.push({
+                        name: Constants.StatsbeatCounter.REQUEST_FAILURE,
+                        value: currentCounter.count,
+                        properties: properties
```
hectorhdzg, line 311: Where are you using currentCounter.statusCode?, I can see you have an accurate count but I'm not sure where are you actually adding it to the Statsbeat properties

Tests/AutoCollection/Statsbeat.tests.ts
```diff
@@ -171,17 +171,17 @@ describe("AutoCollection/Statsbeat", () => {
...
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRetry(0, "test", 206);
+            statsBeat.countRetry(0, "test", 206);
+            statsBeat.countThrottle(0, "test", 206);
+            statsBeat.countException(0, "test", "Statsbeat Exception");
```
hectorhdzg: Add assert checking this exception type

Tests/AutoCollection/Statsbeat.tests.ts
```diff
@@ -171,17 +171,17 @@ describe("AutoCollection/Statsbeat", () => {
...
-            statsBeat.countException(0, "test");
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, true, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRequest(0, "test", 1, false, 200);
+            statsBeat.countRetry(0, "test", 206);
+            statsBeat.countRetry(0, "test", 206);
```
hectorhdzg, line 182: Same here we want to validate you count multiple failed requests, retry and throttle and multiple Statsbeat are generated with the properties part of them

Tests/AutoCollection/Statsbeat.tests.ts
```diff
@@ -171,17 +171,17 @@ describe("AutoCollection/Statsbeat", () => {
...
-            statsBeat.countRequest(0, "test", 1, true);
-            statsBeat.countRequest(0, "test", 1, true);
-            statsBeat.countRequest(0, "test", 1, false);
-            statsBeat.countRequest(0, "test", 1, false);
-            statsBeat.countRequest(0, "test", 1, false);
-            statsBeat.countRetry(0, "test");
-            statsBeat.countRetry(0, "test");
-            statsBeat.countThrottle(0, "test");
-            statsBeat.countException(0, "test");
+            statsBeat.countRequest(0, "test", 1, true, 200);
```
hectorhdzg, line 174: StatusCode should be optional parameter if is not used all the time, like in this case

JacksonWeber: If we make this parameter optional are we ok with creating entries in the request counter like `{ statusCode: undefined, count: <currentCount>`?

hectorhdzg: no need to do that, successful count is only a number

Tests/AutoCollection/Statsbeat.tests.ts
JacksonWeber: Added some new tests that should ensure we're creating multiple metrics per statusCode now.

AutoCollection/Statsbeat.ts
JacksonWeber: Should be fixed now. Added it to the properties in each counter where it's relevant per the spec.

Tests/AutoCollection/Statsbeat.tests.ts
JacksonWeber: Added tests to determine the exceptionType field is populated correctly on the metric.

AutoCollection/Statsbeat.ts
```diff
@@ -118,43 +118,63 @@ class Statsbeat {
...
 
-    public countRequest(endpoint: number, host: string, duration: number, success: boolean) {
+    public countRequest(endpoint: number, host: string, duration: number, success: boolean, statusCode?: number) {
         if (!this.isEnabled()) {
             return;
         }
         let counter: Network.NetworkStatsbeat = this._getNetworkStatsbeatCounter(endpoint, host);
         counter.totalRequestCount++;
         counter.intervalRequestExecutionTime += duration;
         if (success === false) {
```
hectorhdzg, line 128: Add check for statusCode here as well or you will create a null statusCode counter if this code path is ever triggered

hectorhdzg (approved):
LGTM

PR #1008, opened 2022-09-07
Add error when ikey/connection string not present in TelemetryClient

hectorhdzg:


hectorhdzg:
Previous code throwing error while creating Config was causing issues with agent, preventing better logging for diagnostics.

MSNev (approved):

PR #1015, opened 2022-09-22
[Preview] Add custom metrics functionality

hectorhdzg:
Removed old metrics exporter
Disabling Statsbeat metrics


src/autoCollection/metrics/statsbeat/statsbeat.ts
```diff
@@ -362,7 +362,7 @@ export class Statsbeat {
                 metrics: [statsbeat],
                 properties: this._statbeatMetrics[i].properties,
             };
-            this._metricHandler.trackStatsbeatMetric(metricTelemetry);
+            // this._metricHandler.trackStatsbeatMetric(metricTelemetry);
```
JacksonWeber, line 365: Is this just being left in as a placeholder for exporting statsbeat as a custom metric?

hectorhdzg: That is correct, removing old code so AzureMonitorExporter is used instead

JacksonWeber (approved):
LGTM

PR #1019, opened 2022-10-07
[Preview] Allow configuration of instrumentations

hectorhdzg:
Use HTTPInstrumentation to get Standard Metrics now
Removed obsolete configs
Refactor log exporter as the only exporter available in here


JacksonWeber (approved):
LGTM

PR #1020, opened 2022-10-11
[Task]15814831: add prefix class

Karlie-777:
Add prefix helper class that can be re-used by any other components


hectorhdzg (approved):
LGTM

PR #1027, opened 2022-10-27
Update the comment for the default DistributedTracingModes to be accurate

travism-lgm:
The readme.md file indicates DistributedTracingModes.AI_AND_W3C is the default in the 'Configuration' section. 

The default value for CorrelationIdManager.w3cEnabled = true

But the comment in applicationinsights.ts indicates DistributedTracingModes.AI is the default. It looks like this comment is not accurate, so a quick PR to fix this.


hectorhdzg (approved):
LGTM thanks for fixing this

PR #1039, opened 2022-11-23
Address Negative Average Duration

JacksonWeber:
This PR sets `averageRequestExecutionTime` to 0 when given a negative `totalRequestExecutionTime` assuming that in this case something is wrong with the input `executionTime`.

This PR also adds support for more EU regions as defined by: 
https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/statsbeat/eudb.md

And shortens the `shortHost` name to be within spec:
https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/pull/139


hectorhdzg (approved):
LGTM

PR #1040, opened 2022-11-23
Migrate from @azure/core-http to @azure/core-rest-pipeline

jeremymeng:
`@azure/core-rest-pipeline` is the modern and performant version of JavaScript azure core package. And we are planning to archive `@azure/core-http`.


hectorhdzg (approved):
LGTM

PR #1044, opened 2022-12-02
Automatically handle request/dependency correlation in Azure Functions

hectorhdzg:


JacksonWeber (approved):
LGTM

AutoCollection/AzureFunctionsHook.ts
```diff
@@ -0,0 +1,76 @@
...
+        this._client = client;
+        try {
+            this._functionsCoreModule = require('@azure/functions-core');
+        }
+        catch (error) {
+            Logging.info("AutoCollectAzureFunctions failed to load, not running in Azure Functions");
+        }
+    }
+
+    public enable(isEnabled: boolean) {
```
MSNev, line 26: Do you want to add a check to block "double" enabling?

i.e. someone calls enable(true) more than once.

MSNev (approved):

AutoCollection/AzureFunctionsHook.ts
hectorhdzg: Should be handled already by _addPreInvocationHook

Ayzrian:
Hey guys, @MSNev , @JacksonWeber, @hectorhdzg  this PR introduces a critical bug. Starting from version 2.4.0 when this PR was merged the Async functions stopped to work because the AzureFunctionsHook calls originalCallback but does not handle a case when this callback can return promise. 

Feedback from me as an Azure customer. My function app async function basically stopped working properly because of that issue that was very hard to track down, which can impact other Azure customers and production applications as well.

Ayzrian:
Please see the test evidence here https://github.com/Azure/azure-functions-nodejs-worker/issues/652#issuecomment-1382897947

hectorhdzg:
Thanks @Ayzrian please update applicationinsights to previous version, we will investigate and fix on our side, created following issue to track it #1063

PR #1058, opened 2023-01-11
[Beta] Send Feature/Implementation Statsbeat and Refactor LogsExporter

JacksonWeber:
<h3>This PR Covers the Following</h3>

- Updating the LogsExporter to use the OpenTelemetry based Network Statsbeat
- Sends Feature/Implementation and Attach statsbeat to the Azure Monitor Exporter
- Supports Statsbeat in EU regions spec update
- Supports Statsbeat host short name spec update
- Fixes the average duration calculation and batched observable statsbeats

And corresponds with https://github.com/Azure/azure-sdk-for-js/pull/24470 which will need to be merged before classes like the `AzureMonitorStatsbeatExporter` will be available.

### DIFF
--- a/src/applicationInsightsClient.ts
+++ b/src/applicationInsightsClient.ts
@@ -110,6 +110,7 @@ export class ApplicationInsightsClient {
         if (!this._config?.disableOfflineStorage) {
             this._addFeature(StatsbeatFeature.DISK_RETRY);
         }
+        this._addFeature(StatsbeatFeature.DISTRO);
     }
 
     private _addFeature(feature: number) {
--- a/src/applicationInsightsClient.ts
+++ b/src/applicationInsightsClient.ts
@@ -6,7 +6,6 @@ import { LogHandler } from "./logs";
 import { MetricHandler } from "./metrics";
 import { TraceHandler } from "./traces";
 import { ResourceManager } from "./shared";
-import { StatsbeatFeature, StatsbeatInstrumentation } from "./metrics/statsbeat/types";
 
 export class ApplicationInsightsClient {
     private _config: ApplicationInsightsConfig;
@@ -14,9 +13,6 @@ export class ApplicationInsightsClient {
     private _traceHandler: TraceHandler;
     private _metricHandler: MetricHandler;
     private _logHandler: LogHandler;
-
-    private _instrumentation: StatsbeatInstrumentation;
-    private _feature: StatsbeatFeature;
 
     /**
      * Constructs a new client of the client
--- a/src/applicationInsightsClient.ts
+++ b/src/applicationInsightsClient.ts
@@ -31,11 +31,13 @@ export class ApplicationInsightsClient {
         }
         this._getStatsbeatInstrumentations();
         this._getStatsbeatFeatures();
+        process.env.STATSBEAT_INSTRUMENTATIONS = this._statsbeatInstrumentations.toString();
+        process.env.STATSBEAT_FEATURES = this._statsbeatFeatures.toString();
 
         // Statsbeat enable/disable is handled from within the Statsbeat class
         this._statsbeat = new Statsbeat(this._config);
         this._metricHandler = new MetricHandler(this._config);
-        this._traceHandler = new TraceHandler(this._config, this._metricHandler, this._statsbeatInstrumentations, this._statsbeatFeatures);
+        this._traceHandler = new TraceHandler(this._config, this._metricHandler);
         this._logHandler = new LogHandler(this._config, this._metricHandler);
     }
 
--- a/src/applicationInsightsClient.ts
+++ b/src/applicationInsightsClient.ts
@@ -26,7 +26,7 @@ export class ApplicationInsightsClient {
             );
         }
 
-        if (!config.getDisableStatsbeat()) {
+        if (!this._config.getDisableStatsbeat()) {
             this._statsbeat = new Statsbeat(this._config);
             this._statsbeat.setFeatureStatsbeat();
         }
--- a/src/applicationInsightsClient.ts
+++ b/src/applicationInsightsClient.ts
@@ -25,9 +25,15 @@ export class ApplicationInsightsClient {
                 "Connection String not found, please provide it before starting Application Insights SDK."
             );
         }
+
+        if (!config.getDisableStatsbeat()) {
+            this._statsbeat = new Statsbeat(this._config);
+            this._statsbeat.setFeatureStatsbeat();
+        }
+
         this._metricHandler = new MetricHandler(this._config);
         this._traceHandler = new TraceHandler(this._config, this._metricHandler);
-        this._logHandler = new LogHandler(this._config, this._metricHandler);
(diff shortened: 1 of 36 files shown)


src/applicationInsightsClient.ts
```diff
@@ -25,12 +29,13 @@ export class ApplicationInsightsClient {
...
-            this._statsbeat.enable(true);
-        }
+        this._getStatsbeatInstrumentations();
+        this._getStatsbeatFeatures();
+
+        // Statsbeat enable/disable is handled from within the Statsbeat class
+        this._statsbeat = new Statsbeat(this._config);
         this._metricHandler = new MetricHandler(this._config);
-        this._traceHandler = new TraceHandler(this._config, this._metricHandler);
+        this._traceHandler = new TraceHandler(this._config, this._metricHandler, this._statsbeatInstrumentations, this._statsbeatFeatures);
```
JacksonWeber, line 38: @hectorhdzg Would passing the Instrumentations and Features to the Metrics exporters as well make sense? Seems like this would create a situation where both the traceExporter and the metricsExporter create multiple feature/instrumentation statsbeats when this kind of statsbeat is intended to be created once per instance of the application. All depends on if it's possible for a standalone metric to be created, as in that case we'd want to make sure the feature/instrumentation statsbeat is still generated.

src/applicationInsightsClient.ts
```diff
@@ -25,10 +29,15 @@ export class ApplicationInsightsClient {
                 "Connection String not found, please provide it before starting Application Insights SDK."
             );
         }
-        if (!this._config.getDisableStatsbeat()) {
-            this._statsbeat = new Statsbeat(this._config);
-            this._statsbeat.enable(true);
-        }
+        this._getStatsbeatInstrumentations();
```
hectorhdzg, line 32: Let's move all Statsbeat logic to Statsbeat class, it would be easier to find like that

src/applicationInsightsClient.ts
```diff
@@ -25,10 +29,15 @@ export class ApplicationInsightsClient {
...
+        this._getStatsbeatFeatures();
+        process.env.AZURE_MONITOR_STATSBEAT_FEATURES = JSON.stringify({
```
hectorhdzg, line 34: Use constants for this env variable and try/catch all JSON.stringify calls

src/declarations/constants.ts
```diff
@@ -2,3 +2,21 @@ export const APPLICATION_INSIGHTS_SDK_VERSION = "3.0.0-beta.0";
 export const DEFAULT_BREEZE_ENDPOINT = "https://dc.services.visualstudio.com";
 export const DEFAULT_LIVEMETRICS_ENDPOINT = "https://rt.services.visualstudio.com";
 export const DEFAULT_LIVEMETRICS_HOST = "rt.services.visualstudio.com";
+export const NON_EU_CONNECTION_STRING =
```
hectorhdzg, line 5: I'm planning to get rid of declarations folder entirely, please move Statsbeat related types to metrics/statsbeat/types.ts

src/logs/exporters/logExporter.ts
```diff
@@ -78,9 +83,13 @@ export class LogExporter {
                     }, this._batchSendRetryIntervalMs);
                     this._retryTimer.unref();
                 }
+                this._statsbeatMetrics?.countSuccess(duration);
```
hectorhdzg, line 86: It doesn't look like you actually instantiate _statsbeatMetrics anywhere, please unit tests for this as well

src/logs/exporters/logExporter.ts
```diff
@@ -103,6 +114,9 @@ export class LogExporter {
                 return await this._persist(envelopes);
             }
             // Failed -- not retriable
+            if (statusCode) {
+                this._statsbeatMetrics.countFailure(duration, statusCode);
```
hectorhdzg, line 118: If this._statsbeatMetrics is null this will throw an error

src/applicationInsightsClient.ts
JacksonWeber: I moved this logic to the `statsbeat.ts` file and added the try/catch logic to handle any parsing errors from JSON.stringify. I included a const to give the value of the environment variable more visibility in the Statsbeat file, were you looking for me to change where the `process.env` call is made to set the environment variable? Just want to make sure I'm conforming to best practices here.

src/metrics/statsbeat/statsbeat.ts
```diff
@@ -2,39 +2,57 @@ import * as os from "os";
...
     private _isVM: boolean | undefined;
-    private _statbeatMetrics: Array<{ name: string; value: number; properties: unknown }>;
     private _azureVm: AzureVirtualMachine;
+    
+    private _networkStatsbeatMeter: Meter;
+    private _networkStatsbeatMeterProvider: MeterProvider;
+    private _networkAzureExporter: AzureMonitorStatsbeatExporter;
+    private _networkMetricReader: PeriodicExportingMetricReader;
+
+    private _longIntervalStatsbeatMeter: Meter;
```
hectorhdzg, line 52: Long Interval Statbeat Metrics should be handled by the Exporter now right?, is this going to cause duplicated attach and feature metrics?

src/applicationInsightsClient.ts
```diff
@@ -14,6 +15,9 @@ export class ApplicationInsightsClient {
     private _metricHandler: MetricHandler;
     private _logHandler: LogHandler;
 
+    private _instrumentation: StatsbeatInstrumentation;
```
hectorhdzg, line 18: Please remove, looks like this code is not needed anymore

src/applicationInsightsClient.ts
```diff
@@ -25,10 +29,6 @@ export class ApplicationInsightsClient {
                 "Connection String not found, please provide it before starting Application Insights SDK."
             );
         }
-        if (!this._config.getDisableStatsbeat()) {
```
hectorhdzg, line 28: Looks like you don't instantiate and enable Statsbeat anymore in normal execution by customer, unless I'm missing something, please validate E2E of this PR before merging

src/metrics/statsbeat/statsbeat.ts
JacksonWeber: The case I was considering when leaving it in was if a customer produces only logs, I don't see the azure-montior-exporter ever being instantiated. If this isn't possible or the likelihood of it is so low that it's not relevant it should absolutely be removed since you're correct that it would generate duplicate long-interval statsbeats for any customer that produces more than only logs.

src/applicationInsightsClient.ts
JacksonWeber: It was instantiating Statsbeat, this was handled by the logHandler and the statsbeat class would either proceed with constructing the class or skip construction based on the value of `config.getDisableStatsbeat()` however, this move did have the unintended consequence of leaving the setFeatures and setInstrumentations calls until after the azure-monitor-exporter had already been created, therefore always passing it an empty object as an env variable. Refactoring to ensure the Statsbeat class is instantiated before any call to the exporter.

hectorhdzg (approved):
LGTM

PR #1070, opened 2023-01-25
Add back Azure Functions correlation context hook

hectorhdzg:


MSNev (approved):

AutoCollection/AzureFunctionsHook.ts
```diff
@@ -34,84 +43,110 @@ export class AzureFunctionsHook {
...
+                        if (this._isHttpTrigger(ctx)) {
+                            const request: HttpRequest = postInvocationContext.inputs[0];
+                            if (request) {
+                                const startTime: number = postInvocationContext.hookData.appInsightsStartTime || Date.now();
+                                const extractedContext: CorrelationContext | undefined = postInvocationContext.hookData.appInsightsExtractedContext;
+                                if (!extractedContext) {
+                                    this._createIncomingRequestTelemetry(ctx, request, startTime, null);
+                                }
+                                CorrelationContextManager.runWithContext(extractedContext, () => {
+                                    this._createIncomingRequestTelemetry(ctx, request, startTime, extractedContext.operation.parentId);
```
ejizba, line 103: Can you clarify why `_createIncomingRequestTelemetry` is called twice one right after the other? Maybe the second one is supposed to be in an `else` case?

ejizba (approved):
🎉

PR #1071, opened 2023-01-27
[Beta] Updating Azure Function Hook logic

hectorhdzg:
Adding azure/funciton-core types
Removing auto generation of server spans and only support context propagation for now
Updating OpenTelemetry dependencies


JacksonWeber (approved):
LGTM

src/traces/azureFunctionsHook.ts
```diff
@@ -13,19 +13,17 @@ export class AzureFunctionsHook {
     constructor() {
         try {
             this._functionsCoreModule = require("@azure/functions-core");
+            // Only v3 of Azure Functions library is supported right now. See matrix of versions here:
+            // https://github.com/Azure/azure-functions-nodejs-library
+            const funcProgModel = this._functionsCoreModule.getProgrammingModel();
```
MSNev, line 18: Can this return null?

MSNev (approved):

src/traces/azureFunctionsHook.ts
hectorhdzg: It shouldn't but if that happens the catch will take care of it. I don't believe we want to do anything if we cannot get the programming model.

Issue #1085, opened 2023-02-27
`applicationinsights@3.0.0-beta.1` fails on flush


altinokdarici:
`ApplicationInsightsClient.flush()` call fails with `ApplicationInsights:Failed to flush telemetry` `TypeError: Cannot read properties of undefined (reading 'flush')` when one of the metric handlers is undefined because of configuration.

## Repro 
```ts
const config = new ApplicationInsightsConfig();
config.connectionString = '<myConnectionString>';
config.enableAutoCollectHeartbeat = false; // This will cause the failure

const client = new ApplicationInsightsClient(config);
client.start();

await client.flush(); // This line will fail on console.
```

## Explanation

`enableAutoCollectHeartbeat = false;` will cause  `this._heartbeatHandler` not being initialized in https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/src/metrics/metricHandler.ts#LL29C3-L31C10

It will fail on `flush` function call because ` this._heartbeatHandler` is undefined https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/src/metrics/metricHandler.ts#LL70C3-L74C10

PR #1086, opened 2023-02-27
[Beta] Do not flush undefined metricHandlers

altinokdarici:
- Fix: #1085


altinokdarici:
@hectorhdzg I found that there is an issue #1085. I was trying to fix but some CI builds have failed.

hectorhdzg (approved):

Issue #1050, opened 2022-12-28
Bug: Trace messages being dropped when the message is too long


paulsmithkc:
It used to be that app insights would truncate trace messages over 32768 characters. But now it seems that these messages are silently being dropped instead.

https://learn.microsoft.com/en-us/answers/questions/543396/how-to-increase-the-size-of-messages-logged-in-app.html

**Steps to reproduce**

1. Connect to application insights
2. Log a json object which is significantly larger than 32K

**Expected Behavior**

Either the logger throws an exception indicating that the message is too long, or writes a truncated message to AI.

**Actual Behavior**

Trace event is silently dropped, with no sign that it ever existed.

PR #1087, opened 2023-02-27
Truncate Trace Messages When Input Exceeds Maximum Length

JacksonWeber:
Truncates telemetry messages under the 32768 character limit as referenced here: https://learn.microsoft.com/en-us/answers/questions/543396/how-to-increase-the-size-of-messages-logged-in-app

and follows max lengths defined:
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/applicationinsights/data-plane/Monitor.Exporters/preview/v2.1/swagger.json
https://learn.microsoft.com/en-us/azure/azure-monitor/app/data-model

Fixes: #1050


Library/EnvelopeFactory.ts
```diff
@@ -153,15 +172,13 @@ class EnvelopeFactory {
...
     private static createDependencyData(telemetry: Contracts.DependencyTelemetry & Contracts.Identified): Contracts.Data<Contracts.RemoteDependencyData> {
         var remoteDependency = new Contracts.RemoteDependencyData();
-        if (typeof telemetry.name === "string") {
-            remoteDependency.name = telemetry.name.length > 1024 ? telemetry.name.slice(0, 1021) + "..." : telemetry.name;
-        }
-        remoteDependency.data = telemetry.data;
-        remoteDependency.target = telemetry.target;
+        remoteDependency.name = telemetry.name?.substring(0, MAX_NAME_LENGTH);
+        remoteDependency.data = telemetry.data?.substring(0, MAX_PROPERTY_LENGTH);
+        remoteDependency.target = telemetry.target?.substring(0, MAX_NAME_LENGTH);
```
hectorhdzg, line 177: I believe the constants lose benefit when the name cannot be associated with all the things that will use it, like here, you use max name for target field, is way cleaner and easier to read if we have the actual number here instead of having constants, but is up to you I don't strongly oppose to current solution

Library/EnvelopeFactory.ts
```diff
@@ -135,10 +137,27 @@ class EnvelopeFactory {
         }
     }
 
+    private static truncateProperties(telemetry: Telemetry) {
+        if (telemetry.properties) {
+            let properties: {[key: string]: any} = {};
```
hectorhdzg, line 142: You need to ensure this code doesn't throw an error with weird properties, this is code executed for all telemetry events, we need to be extra cautious here

hectorhdzg (approved):
Added some comments, but changes look good

Library/EnvelopeFactory.ts
JacksonWeber: That makes sense. I wanted to avoid magic numbers, but since we're just following the specifications of another API I agree that just using the value could improve readability here.

Library/EnvelopeFactory.ts
JacksonWeber: Added some extra error handling here. The call to serialize the values was already handled by the util function, but makes sense to wrap the whole method in case of seriously unexpected input.

PR #1093, opened 2023-03-01
[Beta] Adding `suppressOverrideMessage` parameter to `updateLogLevel`

altinokdarici:
## Issue
Fixes: #1094 

Currently, we are seeing confusing message when we call `updateLogLevel`. Although this is not an error, it looks like an error. OTel NodeJS SDK has a way to suppress this message however it's not consumable through this application insights sdk.

Screenshot 2023-02-28 at 6 16 35 PM

## Fix 
- Adding `suppressOverrideMessage` parameter to `updateLogLevel`

### DIFF
--- a/src/shared/logging/logger.ts
+++ b/src/shared/logging/logger.ts
@@ -45,9 +45,14 @@ export class Logger implements DiagLogger {
         this.updateLogLevel(this._diagLevel);
     }
 
+    /**
+     * Set the global LogLevel. If a global diag logger is already set, this will override it.
+     * @param logLevel - The DiagLogLevel used to filter logs sent to the logger.
+     * @param suppressOverrideMessage - Setting that suppress the warning message normally emitted when registering a logger when another logger is already registered.
+     */
     public updateLogLevel(logLevel: DiagLogLevel, suppressOverrideMessage?: boolean) {
         this._diagLevel = logLevel;
-        
+
         // Set OpenTelemetry Logger
         diag.setLogger(this, {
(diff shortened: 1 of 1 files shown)


src/shared/logging/logger.ts
```diff
@@ -45,10 +45,14 @@ export class Logger implements DiagLogger {
         this.updateLogLevel(this._diagLevel);
     }
```
hectorhdzg, line 47: Lets add some documentation about this method

/**
   * Set the global LogLevel. If a global diag logger is already set, this will override it.
   * @param logLevel - The DiagLogLevel used to filter logs sent to the logger.
   * @param suppressOverrideMessage - Setting that suppress the warning message normally emitted when registering a logger when another logger is already registered.
   */

src/shared/logging/logger.ts
```diff
@@ -45,10 +45,14 @@ export class Logger implements DiagLogger {
...
 
-    public updateLogLevel(logLevel: DiagLogLevel) {
+    public updateLogLevel(logLevel: DiagLogLevel, suppressOverrideMessage?: boolean) {
```
hectorhdzg, line 48:
```suggestion
    public updateLogLevel(logLevel: DiagLogLevel, suppressOverrideMessage: boolean = true) {
```

altinokdarici: Just to make sure, this will change the default behaviour. are we good with this?

hectorhdzg (approved):
Thanks for your contribution

Issue #1120, opened 2023-04-10
Issue logging with Bunyan


skhilliard:
We recently updated from version 1.8.10 to 2.5.0 and are having some issues with how logs are being handled with the Bunyan monkey  patching.  In our code, we log errors and warnings using a pattern such as this

```
                        this.log.error(
                            err,
                            `Error - url: ${req.url} userId: ${!!req.claims ? req.claims.userId : ''} userName: ${
                                !!req.claims ? req.claims.userName : ''
                            }`
                        );

                        this.log.warn(
                            err,
                            `Warning - url: ${req.url} userId: ${!!req.claims ? req.claims.userId : ''} userName: ${
                                !!req.claims ? req.claims.userName : ''
                            }`
                        );
```

With the 1.8.10 version of application insights, these would all show up in `traces` with varying severity levels with its `message` property containing a property of `err` that contains details about the error object we send in for the first parameter.  In addition, ther would be a `msg` property that contained our formatted message we use for the second parameter.  For example:

```
Error - url: /api/encounters/0/site/57d1d8185f672f881e8a3dd8/query userId: 590b6052b9c9960d98a9e3a9 userName: khilliard@tsysevolved.onmicrosoft.com
```

However, with the new 2.5.0 version, nothing shows up in `traces` anymore.  They are all in `exceptions` and we do not get the formatted string that we provide to the log.error or log.warn messages.  Losing this information in our logs is very problematic.

JacksonWeber:
@skhilliard looks to be related to #978. Would it be possible given how you want to use Bunyan to record the contextual information in a companion `log.info` call or to extend the native `Error` object to include a message field to capture that information? If that's not a viable solution, please let me know.

skhilliard:
@JacksonWeber Unfortunately, that'd mean updating hundreds (maybe thousands) of locations in the code.   I'd prefer not to go that route.

JacksonWeber:
@skhilliard I see. I'll investigate creating a customer-set flag to determine how errors are logged from bunyan.

skhilliard:
> @skhilliard I see. I'll investigate creating a customer-set flag to determine how errors are logged from bunyan.

Thanks!  Just as a reference:  https://github.com/trentm/node-bunyan#log-method-api

The way the code is currently particularly "breaks" cases like the 6th one listed.

JacksonWeber:
[...]

PR #1124, opened 2023-04-17
Include traces as well as exceptions for bunyan to avoid loss of information in messages

skhilliard:
Please note that the only "real" change is to AutoCollection/diagnostic-channel/bunyan.sub.ts.  The other changes (update to gitignore and including output files) were just convenient for us to use this fork for our code until a final solution is implemented in this code base.

For issue: https://github.com/microsoft/ApplicationInsights-node.js/issues/1120


skhilliard:
> This PR wont work for the us as it would send all errors/exceptions as traces as well effectively creating duplicate telemetry for everyone using Bunyan.

I can understand that.....as a temporary measure for us, I think we may update the code to also check the severity level rather than create duplicate entries....this is what I am considering.

```
const subscriber = (event: IStandardEvent<bunyan.IBunyanData>) => {
    let message = event.data.result as string;
    clients.forEach((client) => {
        const AIlevel = bunyanToAILevelMap[event.data.level];
        try {
            // Try to parse message as Bunyan log is JSON
            //   Only log as exception for level of Critical
            let log: any = JSON.parse(message);
            if (log.err && AIlevel === SeverityLevel.Critical) {
                let bunyanError = new Error(log.err.message);
                bunyanError.name = log.err.name;
                bunyanError.stack = log.err.stack;
                client.trackException({ exception: bunyanError });
                return;
            }
        }
        catch (err) {
            // Ignore error
        }
        client.trackTrace({ message: message, severity: AIlevel });
    });
};
```

hectorhdzg:
Config was added in 2.6.0 to address this, thank you.

PR #1125, opened 2023-04-21
Add Azure cloudRoleName and cloudRoleInstance in all telemetry

hectorhdzg:
Inlcudes Live Metrics telemetry as well, adding by default and removing explicit code to enable the feature


JacksonWeber (approved):
LGTM

PR #1146, opened 2023-05-18
[Beta] Read Resource attributes from env

hectorhdzg:
This is needed for AKS integration and new OTel Resources event


JacksonWeber (approved):
LGTM

PR #1151, opened 2023-05-23
[Beta] Add OTLP Trace and Metric Exporters

hectorhdzg:
Remove Heartbeat functionality as well as this will be replaced by OTEL resource event


JacksonWeber (approved):
LGTM

PR #1159, opened 2023-06-09
[Beta] Update default cloud role name and instance ID

hectorhdzg:
Use Azure env variables only in Agent code, use OpenTelemetry defaults


JacksonWeber (approved):
LGTM

PR #1175, opened 2023-08-02
[Shim] Add Auto-Collection Methods, TelemetryClient, and Config

JacksonWeber:
* Add auto-collection methods
* Add global applicationinsights to handle shim import compatibility
* Fix initialization of `setup()` and `start()` so customers can use `client.config` before running `start()`
* Implement `config` properties to set corresponding values on the Azure Monitor Client
* Add tests for the `client.config`


src/shim/util/connectionStringParser.ts
```diff
@@ -0,0 +1,51 @@
+import { ConnectionString, ConnectionStringKey } from "../types";
+import Constants = require("../../../src/declarations/constants");
+
+class ConnectionStringParser {
```
hectorhdzg, line 4: This class is also available in OTel Exporter, why do we want to parse the connection string multiple times?, it would be better to not support very custom scenarios where this applies

src/shim/util/configHelper.ts
```diff
@@ -0,0 +1,81 @@
+import { HttpInstrumentationConfig } from "@opentelemetry/instrumentation-http";
+import { ApplicationInsightsOptions, ExtendedMetricType } from "../../types";
+import * as http from "http";
+
+class ConfigHelper {
```
hectorhdzg, line 5: It feels you don't need a class here, everything is static and there are no attributes in it, just export the methods directly instead like we do in some other utils

src/shim/telemetryClient.ts
```diff
@@ -14,29 +14,41 @@ import { Logger } from "./logging";
...
     private _idGenerator: IdGenerator;
     public context: Context;
     public commonProperties: { [key: string]: string }; // TODO: Add setter so Resources are updated
+    public config: IConfig;
 
     /**
      * Constructs a new client of the client
      * @param setupString the Connection String or Instrumentation Key to use (read from environment variable if not specified)
      */
     constructor(input?: string | ApplicationInsightsOptions) {
```
hectorhdzg, line 47: Remove ApplicationInsightsOptions parameter from here and leave same structure as old Application Insights

src/shim/telemetryClient.ts
```diff
@@ -46,14 +58,278 @@ export class TelemetryClient {
...
+            Logger.getInstance().warn("The webInstrumentation configuration options are not supported by the shim.");
+        }
+    }
+    
+    /**
+     * Starts automatic collection of telemetry. Prior to calling start no telemetry will be collected
+     * @param input Set of options to configure the Azure Monitor Client
+     */
+    public start(input?: ApplicationInsightsOptions) {
+        if (_setupCalled) {
```
hectorhdzg, line 258: I'm confused why this flag exists, start is not a method that existed before so should be only called in shim-applicationinsights.ts

src/shim/telemetryClient.ts
```diff
@@ -436,4 +712,16 @@ export class TelemetryClient {
         const record = this._telemetryToLogRecord(telemetry, baseType, baseData);
         return record;
     }
+
+    public trackNodeHttpRequestSync(telemetry: Contracts.NodeHttpRequestTelemetry) {
```
hectorhdzg, line 645: If customer try to directly NodeClient your code will fail, these methods should be in a class called NodeClient

src/shim/telemetryClient.ts
```diff
@@ -436,4 +712,16 @@ export class TelemetryClient {
...
+        Logger.getInstance().warn("trackNodeHttpRequestSync is not implemented and is a no-op.");
+    }
+
+    public trackNodeHttpRequest(telemetry: Contracts.NodeHttpRequestTelemetry) {
+        Logger.getInstance().warn("trackNodeHttpRequest is not implemented and is a no-op.");
```
hectorhdzg, line 650: Why are these not supported? what exactly these were doing before?

src/shim/telemetryClient.ts
JacksonWeber: Made some updates to this, but the reasoning is that in the non-shim scenario we want to treat the creation of the `AzureMonitorOTelClient` as part of the constructor of the `TelemetryClient` class, but in the case of the shim we want to separate that logic out between the `setup()` and `start()` methods in the `ApplicationInsights` class. Issue here was that when I refactored, I removed the check from the constructor. Hence why the check didn't make sense. Excellent catch!

src/shim/util/connectionStringParser.ts
JacksonWeber: Given we're not supporting use cases that are this custom, I agree that this functionality seems extraneous. Removed this and the custom endpointUrl constructing.

src/shim/telemetryClient.ts
JacksonWeber: That will break some things like the functional tests and the configuration example in the distro's README. Are we planning to release the shim after some changes to the distro's user configuration?

src/shim/telemetryClient.ts
JacksonWeber: They were some helper functions that called methods (`trackRequestSync` and `trackRequest`) inside of the `AutoCollectHttpRequests` class. Given that OpenTelemetry tracks requests synchronously by default I figured it'd be less confusing to just say these methods weren't supported and have users unify on calling the single `trackRequest` method. 

I think it'd also be fair to implement them all calling the `TelemetryClient.trackRequest()` or `trackDependency` methods and just note that these will all be synchronous.

src/shim/telemetryClient.ts
```diff
@@ -307,7 +517,7 @@ export class TelemetryClient {
             contextObjects?: { [name: string]: any }
         ) => boolean
     ) {
-        throw new Error("Not implemented");
+        Logger.getInstance().warn("addTelemetryProcessor is not supported via the ApplicationInsights shim. Please upgrade to the ApplicationInsights SDK beta.");
```
hectorhdzg, line 520: Incorrect message here, AI Shim is the same as AI SDK beta right now, telemetry processor are not supported anymore

hectorhdzg (approved):
LGTM

src/shim/telemetryClient.ts
JacksonWeber: I'll update the message here to make it more explicit to users the expected alternative method to call.

PR #1185, opened 2023-08-14
Fix issue with older version of Node.js runtime

hectorhdzg:


JacksonWeber (approved):

package.json
```diff
@@ -61,8 +61,8 @@
   },
   "dependencies": {
     "@azure/core-auth": "^1.5.0",
-    "@azure/core-rest-pipeline": "1.12.0",
-    "@azure/core-util": "1.4.0",
```
JamieMagee, line 65: I don't think you need an explicit dependency on this package. It was just to pin the version used by other dependencies?

Can you try using npm overrides instead[^1]?

[^1]: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

hectorhdzg: Cool, never used that one before, thanks for sharing

JamieMagee (approved):

PR #1193, opened 2023-08-22
[Shim] Add Support for Context.tags

JacksonWeber:
* Added support for context.tags via a SpanProcessor to affix the tags to all exported spans.
* Added tests for tags.
* Fix issues with `enableAutoCollectRequests` and `enableAutoCollectDependencies`


src/shared/util/attributeSpanProcessor.ts
```diff
@@ -0,0 +1,16 @@
+import { BatchSpanProcessor, Span } from "@opentelemetry/sdk-trace-base";
+import { SpanExporter } from "@opentelemetry/sdk-trace-node";
+import { context } from "@opentelemetry/api";
+
+export class AttributeSpanProcessor extends BatchSpanProcessor {
```
hectorhdzg, line 5: You should be implementing SpanProcessor from @opentelemetry/sdk-trace-base
https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/monitor/monitor-opentelemetry/src/traces/spanProcessor.ts

src/shared/util/attributeLogRecordProcessor.ts
```diff
@@ -0,0 +1,14 @@
+import { BatchLogRecordProcessor, LogRecord, LogRecordExporter } from "@opentelemetry/sdk-logs";
+
+export class AttributeLogProcessor extends BatchLogRecordProcessor {
+    private _attributes: { [key: string]: string };
+    constructor(exporter: LogRecordExporter, attributes: { [key: string]: string }) {
+        super(exporter);
+        this._attributes = attributes;
+    }
+    // Override onEmit to apply log record attributes before exporting
+    onEmit(record: LogRecord) {
```
hectorhdzg, line 10: Same here, you need to implement the interface LogRecordProcessor not extending and overriding Batch one, then you need to add the LogProcessor using global LoggerProvider

hectorhdzg (approved):
LGTM

PR #1194, opened 2023-08-23
[Beta] Consume latest Azure Monitor OpenTelemetry

hectorhdzg:
Align with names used in Azure SDK projects
ApplicationInsightsClient to be used when manually using the SDK to not conflict with shim TelemetryClient
 Added LogApi class for manual ApplicationInsights event APIs


src/shim/applicationinsights.ts
```diff
@@ -42,15 +31,7 @@ let _options: ApplicationInsightsOptions;
  * and start the SDK.
  */
 export function setup(setupString?: string) {
-    _setupCalled = true;
-    // Save the setup string and create a config to modify with other functions in this file
-    _setupString = setupString;
-    if (!_options) {
-        _options = { azureMonitorExporterConfig: { connectionString: _setupString } };
-    } else {
-        Logger.getInstance().info("Cannot run applicationinsights.setup() more than once.");
```
JacksonWeber, line 51: Do we want to preserve this warning in the case the customer attempts to call setup multiple times with different connection strings?

hectorhdzg: nothing happens if customer call this method multiple times

JacksonWeber: Makes sense. My thought was just that we don't want to give the impression that calling setup multiple times allows for usage of multiple clients, however you're right that we need to worry about being able to setup multiple clients at all first.

src/shared/util/configHelper.ts
```diff
@@ -77,42 +77,20 @@ export function setAutoCollectConsole(options: ApplicationInsightsOptions, value
...
-    options.logInstrumentations = {
-        ...options.logInstrumentations,
+    options.logInstrumentationOptions = {
+        ...options.logInstrumentationOptions,
         winston: { enabled: value },
         bunyan: { enabled: value },
     }
 }
 
 export function enableAutoCollectConsole(options: ApplicationInsightsOptions, value: boolean) {
```
JacksonWeber, line 87: Doesn't look like this function is necessary anymore as `enableAutoCollectConsole()` runs inside of the config.ts file.

JacksonWeber: As a follow up if have all of these functions used/written in the config.ts file we may be able to eliminate this helper file.

JacksonWeber (approved):
LGTM

PR #1196, opened 2023-08-24
Avoid request to be blocked once the "dispose" method has been called

### LINKED ISSUE #1195: [BUG] Calling the global "dispose" method prevents the express server from running
### Context
I'm currently developing an API that will, potentially, call the dispose method in order to stop all collectors temporarily.
However, once the dispose method has been called, the server will no longer respond to requests.

### Code to reproduce
```Javascript
const ai = require('applicationinsights');
const express = require('express');

const app = express();
app.post('/log/start', (req, res) => {
    ai.setup('<valid-key>')
        .setAutoDependencyCorrelation(true)
        .setAutoCollectRequests(true)
        .setAutoCollectPerformance(false)
        .setAutoCollectExceptions(true)
        .setAutoCollectDependencies(true)
        .setUseDiskRetryCaching(true)
        .setSendLiveMetrics(false)
        .setDistributedTracingMode(ai.DistributedTracingModes.AI_AND_W3C)
        .setAutoCollectConsole(true, true);
    ai.start();

    res.status(200).send('OK');
});
app.post('/log/stop', (req, res) => {
    ai.dispose();

    res.status(200).send('OK');
});
app.listen(3000, () => console.log('Example app listening on port 3000!'));
```
### Steps to reproduce
1. Create a project with the code above (install dependencies required)
2. Start the server
3. Call the route '/log/stop'
4. Try to call the route '/log/start' -> You will never have an answer

### Environment
- NodeJS : v18.16.0
- NPM : 9.8.1
- applicationinsights : 2.7.3

Apokalypt:
When the dispose method is called, the static INSTANCE property is set to null, which means there is no query handler (neither the original nor the custom handler).
So I've added a control to call the default handler if the INSTANCE property is set to null

Fix #1195

### DIFF
--- a/AutoCollection/HttpRequests.ts
+++ b/AutoCollection/HttpRequests.ts
@@ -122,9 +122,7 @@ class AutoCollectHttpRequests {
                 if (request && shouldCollect && AutoCollectHttpRequests.INSTANCE) {
                     AutoCollectHttpRequests.INSTANCE._registerRequest(request, response, onRequest)
                 } else {
-                    if (typeof onRequest === "function") {
-                        onRequest(request, response);
-                    }
+                    onRequest(request, response);
                 }
             }
         };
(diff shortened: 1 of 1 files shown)


hectorhdzg (approved):
LGTM, thanks for your contribution

PR #1198, opened 2023-08-30
[Shim] Add Active Span Generation

JacksonWeber:
* Add active span generation so that if `.runWithContext()` is invoked with the `appInsights.getCurrentContext()` value, the callback passed to `.runwithContext()` will not be run with a null context.
* Add error handling for correlation context methods.


src/shim/correlationContextManager.ts
```diff
@@ -6,6 +6,9 @@ import { Span } from "@opentelemetry/sdk-trace-base";
 import { ICorrelationContext, ITraceparent, ITracestate, HttpRequest, ICustomProperties } from "./types";
 import { Logger } from "../shared/logging";
 import * as azureFunctionsTypes from "@azure/functions";
+import { Util } from "../shared/util";
+
+const CONTEXT_NAME = "ApplicationInsights-Context";
```
hectorhdzg, line 11: Can you enable correlation context tests as well now?

hectorhdzg (approved):
LGTM

PR #1200, opened 2023-08-31
Decode ai_authUser URI

JacksonWeber:
Should decode the ai_authUser cookie since it's passed from the JavaScript SDK as a URI.

### DIFF
--- a/AutoCollection/HttpRequestParser.ts
+++ b/AutoCollection/HttpRequestParser.ts
@@ -10,7 +10,8 @@ import CorrelationIdManager = require("../Library/CorrelationIdManager");
 import Tracestate = require("../Library/Tracestate");
 import Traceparent = require("../Library/Traceparent");
 import { HttpRequest } from "../Library/Functions";
-
+import Logging = require("../Library/Logging");
+import { HttpRequestCookieNames } from "../Declarations/Constants";
 
 /**
  * Helper class to read data from the request/response objects and convert them into the telemetry contract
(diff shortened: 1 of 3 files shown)


Tests/AutoCollection/HttpRequestParser.tests.ts
```diff
@@ -202,7 +202,7 @@ describe("AutoCollection/HttpRequestParser", () => {
             headers: {
                 host: "bing.com",
                 "x-forwarded-for": "123.123.123.123",
-                "cookie": "ai_user=cookieUser|time;ai_session=cookieSession|time;ai_authUser=cookieAuthUser|time",
+                "cookie": `ai_user=cookieUser|time;ai_session=cookieSession|time;ai_authUser=${encodeURI("cookieAuthUser|time")}`,
```
hectorhdzg, line 205: Can you add an actual test for encoded vs decoded?, what happen to the scenario where value is not a URI?

AutoCollection/HttpRequestParser.ts
```diff
@@ -247,6 +247,10 @@ class HttpRequestParser extends RequestParser {
     private _getId(name: string) {
         var cookie = (this.rawHeaders && this.rawHeaders["cookie"] &&
             typeof this.rawHeaders["cookie"] === "string" && this.rawHeaders["cookie"]) || "";
+
+        if (name === "ai_authUser") {
+            cookie = decodeURI(cookie);
```
hectorhdzg, line 252: Can this code throw an exception? if yes we should try/catch

AutoCollection/HttpRequestParser.ts
```diff
@@ -247,6 +247,10 @@ class HttpRequestParser extends RequestParser {
     private _getId(name: string) {
         var cookie = (this.rawHeaders && this.rawHeaders["cookie"] &&
             typeof this.rawHeaders["cookie"] === "string" && this.rawHeaders["cookie"]) || "";
+
+        if (name === "ai_authUser") {
```
hectorhdzg, line 251: Do we have some kind of enum or something to avoid magic strings?

Tests/AutoCollection/HttpRequestParser.tests.ts
JacksonWeber: Sure - I had assumed based on Nev's comments that we'd always receive an encoded URI here. The good news is that the `encodeURI()` method if passed a standard string will just return that same string. To throw an exception here someone would have to pass in a string containing a "%" without two hex digits following it. In that case I believe it's safe to assume that we've received a malformed URI and should catch that exception.

hectorhdzg (approved):

PR #1203, opened 2023-09-06
[Shim] Fix JSON Config

JacksonWeber:
* Add remaining JSON config values.
* Fix JSON config file path and stop us from overwriting defined env vars with undefined values from the JSON file.


src/shim/config.ts
```diff
@@ -288,6 +288,12 @@ class Config implements IConfig {
             };
         }
 
+        if (ShimJsonConfig.getInstance().noHttpAgentKeepAlive === true) {
```
hectorhdzg, line 291: This config was never related to OTLP exporters, we should not be doing this

src/logs/diagnostic-channel/initialization.ts
```diff
@@ -4,14 +4,17 @@
...
 
 const TAG = "DiagnosticChannel";
 let isInitialized = false;
 
 export function enablePublishers() {
     // Only register monkey patchs once
-    if (!isInitialized) {
+    if (!isInitialized && (ShimJsonConfig.getInstance().noDiagnosticChannel !== true)) {
         isInitialized = true;
+        const individualOptOuts: string = ShimJsonConfig.getInstance().noPatchModules;
```
hectorhdzg, line 16: Why are you adding the check here? you should be checking for config options in the subscriber code not here in publishers side

src/logs/diagnostic-channel/initialization.ts
```diff
@@ -4,14 +4,17 @@
 // Don't reference modules from these directly. Use only for types.
 import * as DiagChannelPublishers from "diagnostic-channel-publishers";
 import { Logger } from "../../shared/logging";
+import { ShimJsonConfig } from "../../shim/jsonConfig";
```
hectorhdzg, line 7: JSON config should never used outside of our Internal config class, value could be overridden using JS code so this ignores that

src/shim/jsonConfig.ts
```diff
@@ -119,8 +119,39 @@ export class ShimJsonConfig implements IJsonConfig {
         }
         try {
             const jsonConfig: IJsonConfig = JSON.parse(jsonString);
-            this.connectionString = jsonConfig.connectionString;
-            this.enableAutoCollectExceptions = jsonConfig.enableAutoCollectExceptions;
+            if (jsonConfig.connectionString !== undefined) {
```
hectorhdzg, line 122: Why you check for undefined?, when we merge undefined should be discarded

src/shim/config.ts
```diff
@@ -338,31 +344,6 @@ class Config implements IConfig {
...
-    * Not empty
-    * Every character is a hex character [0-9a-f]
-    * 32 characters are separated into 5 sections via 4 dashes
-    * First section has 8 characters
-    * Second section has 4 characters
-    * Third section has 4 characters
-    * Fourth section has 4 characters
-    * Fifth section has 12 characters
-    */
-    private static _validateInstrumentationKey(iKey: string): boolean {
```
hectorhdzg, line 356: We should add this code in Exporter side, please create a task for it

src/logs/diagnostic-channel/initialization.ts
JacksonWeber: Rational was that we used to disable diagnostic channels on the publisher side, however I agree it makes sense to disable them at a subscriber level. Will update that.

src/shim/jsonConfig.ts
JacksonWeber: Checking for undefined here on values that can be configured by environment variable earlier in the file. I found in testing that they were being overridden without the undefined checks.

src/logs/diagnostic-channel/initialization.ts
JacksonWeber: noDiagnosticChannel and noPatchModules cannot be configured via `defaultClient.config.noDiagnosticChannel` or `defaultClient.config.noPatchModules`, only via env var or JSON config, so I don't believe this could be overridden before being read here. Were you referring to another way it could be overridden?

src/logs/diagnostic-channel/initialization.ts
JacksonWeber: Refactored this to use the publisher's `enable` functions (specifically not calling them if noDiagnosticChannel is set to true) in the same way as `noPatchModules`.

hectorhdzg: Not seeing those changes here, is there a commit missing?

JacksonWeber: Changed such that we still do this check on the publisher side (line 26 is where we do the check for noDiagnosticChannel). It takes advantage of the same `enable()` methods each of the subscribers has. If we don't want to do this check for noDiagnosticChannel here we could move the logic for both noDiagnosticChannel and noPatchModules to each individual subscriber, but after looking into moving the code to the subscribers it seems like it'd make this logic more complicated than necessary.

hectorhdzg (approved):
LGTM

Issue #1169, opened 2023-07-06
[Beta] - Nested objects get serialised as "[object Object]"


mderriey:
Hi there 👋

I hope you're well.

## Issue

We've upgraded from `3.0.0-beta.5` to `3.0.0-beta.6` recently and we noticed that nested objects get serialised as "[object Object]".

See the screenshot below.
For reference, we deployed the package upgrade sometime during the 3rd of July.
No traces before the 3rd of July show `customDimensions.variables` with a value of `[object Object]`

![image](https://github.com/microsoft/ApplicationInsights-node.js/assets/6102639/ff7d8ee9-b04e-4f57-b035-948ae69e2b32)

## Setup

We don't interact with the SDK or the trace handler directly, all our logging is done through winston, and we have the appropriate log instrumentation enabled.

Here's the whole config, in case it helps:

```typescript
import { ExpressInstrumentation } from '@opentelemetry/instrumentation-express'
import { GraphQLInstrumentation } from '@opentelemetry/instrumentation-graphql'
import { WinstonInstrumentation } from '@opentelemetry/instrumentation-winston'
import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
import { Resource } from '@opentelemetry/resources'
import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
import { ApplicationInsightsClient, ApplicationInsightsConfig } from 'applicationinsights'
import { TediousInstrumentation } from '@opentelemetry/instrumentation-tedious'
import { DataloaderInstrumentation } from '@opentelemetry/instrumentation-dataloader'

// Default attributes set on the OpenTelemetry traces
const resource = Resource.EMPTY
resource.attributes[SemanticResourceAttributes.SERVICE_NAME] = '<redacted>'
resource.attributes[SemanticResourceAttributes.SERVICE_NAMESPACE] = '<redacted>'

const config = new ApplicationInsightsConfig()
config.resource = resource

/**
 * Disabling this for now as it shows 1000s of failed requests on the App Insights dashboard.
 * See https://github.com/microsoft/ApplicationInsights-node.js/issues/1123
 */
config.enableAutoCollectStandardMetrics = false

/**
 * - Disable the built-in HTTP instrumentation and use the official one (see below)
 * - Enable the Azure SDK instrumentation
 * - Enable winston instrumentation (which sends logs to App Insights)
 */
config.instrumentations.http = { enabled: false }
[...]

PR #1206, opened 2023-09-12
[Beta] Stringify objects when building a `LogRecord`

mderriey:
Related to #1169

After being invited by @JacksonWeber to open a PR [in this comment](https://github.com/microsoft/ApplicationInsights-node.js/issues/1169#issuecomment-1714469041), here it is.

The one change from the code snippet I posted in that issue is we're using `Util.getInstance().stringify` instead of `JSON.stringify` as I saw it's used a few lined below that code block.


mderriey:
I couldn't `npm install` as I don't have Python installed, which is why I opened this as a draft for now.

Could you please approve the workflow that would run tests to make sure it's all green before I mark as ready to review?

JacksonWeber:
@mderriey It looks like the workflow ran ok! And the PR looks good. Thank you for your contribution and approving!

JacksonWeber (approved):
LGTM

PR #1209, opened 2023-09-20
[Beta] Rename Options interface to AzureMonitorOpenTelemetryOptions

hectorhdzg:
This change will make all public sample code for @azure/monitor-opentelemetry to apply to this package as well


JacksonWeber (approved):
LGTM

PR #1211, opened 2023-09-20
[Shim] Resolve Shim TODOs and Add Global App Insights Tests

JacksonWeber:
* Resolve TODOs
* Add Global App Insights tests
* Add more tests for Telemetry Client
* Add custom metrics for manual metrics collection & tests


src/shim/telemetryClient.ts
```diff
@@ -118,12 +120,14 @@ export class TelemetryClient {
      * telemetry bandwidth by aggregating multiple measurements and sending the resulting average at intervals.
      * @param telemetry      Object encapsulating tracking options
      */
-    public trackMetric(telemetry: Contracts.MetricTelemetry): void {
-        // TODO : Create custom metric
-        // let meter = this.client.getMetricHandler().getCustomMetricsHandler().getMeter();
-        // let metricName = "";
-        // let options: MetricOptions = {};
-        // meter.createHistogram(metricName, options)
+    public trackMetric(telemetry: Contracts.MetricPointTelemetry): void {
```
hectorhdzg, line 123: Is there any attribute that can be passed for the Metrics in this API?

src/shim/telemetryClient.ts
```diff
@@ -224,12 +228,13 @@ export class TelemetryClient {
...
 
     /**
+     * @deprecated
      * Automatically populate telemetry properties like RoleName when running in Azure
      *
      * @param value if true properties will be populated
      */
     public setAutoPopulateAzureProperties() {
-        // TODO: Config is only used during initialization of ResourceManager so it cannot be set after.
+        // NO-OP
```
hectorhdzg, line 237: We want to support this, it should read some env variables and populate cloud role and role name, if you can create a task to follow up on this it would be great

src/shim/shim-config.ts
```diff
@@ -168,7 +171,7 @@ class Config implements IConfig {
             ...options.instrumentationOptions,
             http: {
                 ...options.instrumentationOptions?.http,
-                ignoreOutgoingUrls: this.correlationHeaderExcludedDomains, /// TODO: Deprecated configuration
```
JacksonWeber, line 171: Created a task to move this logic to the distro using ignoreOutgoingRequests to filter URLs.

src/shim/telemetryClient.ts
JacksonWeber: Updated typing and pass any custom properties via the `Histogram.record()` method's `attributes` parameter.

src/shim/telemetryClient.ts
```diff
@@ -139,7 +143,6 @@ export class TelemetryClient {
         const startTime = telemetry.time || new Date();
         const endTime = startTime.getTime() + telemetry.duration;
 
-        // TODO: Change resourceManager if ID is provided?
```
JacksonWeber, line 142: Don't believe this is necessary. If a request id is passed to `trackRequest` it will be tracked via the customProperties (attributes).

src/shim/telemetryClient.ts
```diff
@@ -191,7 +194,6 @@ export class TelemetryClient {
         };
         if (telemetry.dependencyTypeName) {
             if (telemetry.dependencyTypeName.toLowerCase().indexOf("http") > -1) {
-                attributes[SemanticAttributes.HTTP_METHOD] = "HTTP"; // TODO: Dependency doesn't expose method in any property
```
JacksonWeber, line 194: This setting is not exposed in telemetry, so we can remove it.

src/declarations/contracts/telemetryTypes/metricTelemetry.ts
```diff
@@ -7,7 +7,7 @@ import { Telemetry } from "./telemetry";
  */
 export interface MetricTelemetry extends Telemetry {
     /** List of metrics. Only one metric in the list is currently supported by Application Insights storage. If multiple data points were sent only the first one will be used. */
-    metrics: MetricPointTelemetry[];
+    metrics?: MetricPointTelemetry[];
```
JacksonWeber, line 10: Made metrics optional here and updated the typing accepted by trackMetric to `MetricTelemetry & MetricPointTelemetry` such that customers can pass MetricPointTelemetry including additional `properties`. However, this means that we need to make `metrics` optional since coming from AppInsights 2.x customers are not expecting to pass a MetricPointTelemetry array.

src/shim/shim-config.ts
hectorhdzg: Looks like correlation headers config is deprecated according to the comment, why you want to move that to distro?

src/shim/telemetryClient.ts
hectorhdzg: If you remove this, how can Exporter know the type?

src/shim/shim-config.ts
JacksonWeber: We already do processing there in the trace handler for filtering using `ignoredOutgoingRequestHook`. I'm also assuming that if we wanted to block dependency calls to the default excluded domains like `"*.core.windows.net"` in the distro as well it'd make sense to unify that work there. 
However, if we're not planning to filter any domains out in the distro, I can implement this logic here.

hectorhdzg: ignoredOutgoingRequestHook can also be provided by customers or added here in this project, this config is about correlation headers not blocking dependencies, I think you got the concept wrong, current behavior in shim is incorrect

JacksonWeber: I just realized what my confusion here was. I assumed the comment was referring to the `ignoreOutoingUrls` being deprecated (since it is a deprecated config option), while the comment really refers to `correlationHeaderExcludedDomains` being deprecated. Makes far more sense now, thanks!

src/shim/telemetryClient.ts
JacksonWeber: Ah, my mistake. Tested dependency auto-collection which wasn't affected by this change. Looks like manual dependency telemetry registers the type properly and sets `HTTP` correctly.

src/shim/telemetryClient.ts
JacksonWeber: As a follow up to this, I don't see any Semantic Resource Attributes in the exporter that would correspond to a request ID.

src/shim/shim-config.ts
```diff
@@ -377,6 +362,9 @@ class Config implements IConfig {
         ) {
             Logger.getInstance().warn("The webInstrumentation configuration options are not supported by the shim.");
         }
+        if (this.correlationHeaderExcludedDomains) {
```
hectorhdzg, line 365: Why is this deprecated? I know incorrect behavior was present but is not possible to control which requests to propagate the header?

src/shim/telemetryClient.ts
hectorhdzg: We allow customers to send dependency Type before, we should keep support for that, that is what the TODO was there for, not all dependencies are HTTP

JacksonWeber: Created task to track this and investigate what values we need to support for dependency type.

src/shim/shim-config.ts
JacksonWeber: Removed deprecation and created a task to track this.

hectorhdzg (approved):

PR #1212, opened 2023-09-20
[Beta] 3.0.0 beta.9 release

JacksonWeber:


package.json
```diff
@@ -3,7 +3,7 @@
   "author": "Microsoft Application Insights Team",
   "license": "MIT",
   "bugs": "https://github.com/microsoft/ApplicationInsights-node.js/issues",
-  "version": "3.0.0-beta.8",
+  "version": "3.0.0-beta.9",
```
hectorhdzg, line 6: Are we not setting AZURE_MONITOR_DISTRO_VERSION with this? we should because there will be no way to differentiate data coming from this package in Statsbeat

JacksonWeber: Updated the constant

hectorhdzg (approved):
LGTM

PR #1214, opened 2023-09-25
Fix issue with JSON config precedence in App Services

hectorhdzg:
Remove duplicated logic of configuration in global application insights object


JacksonWeber (approved):
LGTM

PR #1227, opened 2023-10-12
[Beta] Removed duplicate types

hectorhdzg:


test/unitTests/logs/api.tests.ts
```diff
@@ -21,7 +20,7 @@ describe("logs/API", () => {
 
     before(() => {
         sandbox = sinon.createSandbox();
-        nock(DEFAULT_BREEZE_ENDPOINT)
+        nock("https://dc.services.visualstudio.com")
```
JacksonWeber, line 23: If this constant is a dupe can we pull it from the exporter?

hectorhdzg: Constant are not exposed in the exporter, that would require API review by Azure SDK architects

JacksonWeber (approved):
LGTM

PR #1236, opened 2023-10-25
[Beta] Fix SDK Already Exists Check for Linux

JacksonWeber:
Based on investigation, the current working directory is not captured by the Node 8.9+ check of the appInstance value in Linux environments. Therefore, we miss the check intended to send a diagnostic log indicating that the SDK had already attached. This is the fix for the beta.


hectorhdzg (approved):
LGTM

PR #1243, opened 2023-11-15
Add Check for Invalid iKey

JacksonWeber:
If an invalid iKey is passed to application insights we should disable statsbeat.

### DIFF
--- a/AutoCollection/Statsbeat.ts
+++ b/AutoCollection/Statsbeat.ts
@@ -426,7 +426,7 @@ class Statsbeat {
     }
 
     private _shutdownStatsbeat() {
-        this.enable(false);// Disable Statsbeat as is it failed 3 times cosnecutively during initialization, is possible SDK is running in private or restricted network
+        this.enable(false);// Disable Statsbeat as is it failed 3 times consecutively during initialization, is possible SDK is running in private or restricted network
     }
 
     private _getConnectionString(config: Config): string {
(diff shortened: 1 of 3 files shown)


hectorhdzg (approved):
LGTM

d00rsfan:
It began to throw exception!
Which is i think wrong behaviour for minor version change, if excheption not catched, app will die!
`Error: Instrumentation key not found, please provide a connection string before starting Application Insights SDK.
    at NodeClient.TelemetryClient`

JacksonWeber:
> It began to throw exception! Which is i think wrong behaviour for minor version change (2.9.1->2.9.2), if exception not catched, app will die! `Error: Instrumentation key not found, please provide a connection string before starting Application Insights SDK. at NodeClient.TelemetryClient`

Can I see what you're using to initialize the SDK on your side? I don't believe this change should be throwing that error for you, but happy to take a look.

d00rsfan:
If version locked to "applicationinsights": "2.9.1", this example works forever.

JacksonWeber:
> Thank you for reply! I have reproduced it. Was confused that uncaught exception happened after 70sec of uptime. No reason to create project. It has only 1 file and 1 dependency in package.json: This programm will die in 70sec
> 
> ```
> import appInsights from 'applicationinsights';
> 
> appInsights.setup('foobar').start();
> 
> let sec = 10;
> setInterval(() => {
>   console.log(`uptime=${sec}seconds`);
>   sec+=10;
> }, 10000);
> console.log('started');
> ```
> 
> ......... uptime=70seconds ApplicationInsights:Instrumentation key was invalid, please check the iKey [] /path/to/thisproject/node_modules/applicationinsights/out/Library/Sender.js:199 _this._shutdownStatsbeat(); ^
> 
> TypeError: _this._shutdownStatsbeat is not a function at IncomingMessage. (/path/to/thisproject/node_modules/applicationinsights/out/Library/Sender.js:199:47) at IncomingMessage.emit (node:events:530:35) at endReadableNT (node:internal/streams/readable:1696:12) at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
> 
> Node.js v20.11.0

I see. https://github.com/microsoft/ApplicationInsights-node.js/pull/1259 should address this issue and will be included in the next release.

PR #1250, opened 2023-12-07
Add New Features to Statsbeat

JacksonWeber:
Add tracking for native metrics and live metrics to statsbeat collection.

### DIFF
--- a/AutoCollection/NativePerformance.ts
+++ b/AutoCollection/NativePerformance.ts
@@ -54,10 +54,6 @@ export class AutoCollectNativePerformance {
                 const NativeMetricsEmitters = require("applicationinsights-native-metrics");
                 AutoCollectNativePerformance._emitter = new NativeMetricsEmitters();
                 AutoCollectNativePerformance._metricsAvailable = true;
-                // Should only set statsbeat feature if it's the first time calling enabled
-                if (!this._isEnabled) {
-                    this._statsbeat.addFeature(Constants.StatsbeatFeature.NATIVE_METRICS);
-                }
                 Logging.info("Native metrics module successfully loaded!");
             } catch (err) {
                 // Package not available. Never try again
@@ -74,6 +70,7 @@ export class AutoCollectNativePerformance {
 
         // Enable the emitter if we were able to construct one
         if (this._isEnabled && AutoCollectNativePerformance._emitter) {
+            this._statsbeat.addFeature(Constants.StatsbeatFeature.NATIVE_METRICS);
             // enable self
             AutoCollectNativePerformance._emitter.enable(true, collectionInterval);
             if (!this._handle) {
@@ -81,6 +78,7 @@ export class AutoCollectNativePerformance {
                 this._handle.unref();
             }
         } else if (AutoCollectNativePerformance._emitter) {
+            this._statsbeat.removeFeature(Constants.StatsbeatFeature.NATIVE_METRICS);
             // disable self
             AutoCollectNativePerformance._emitter.enable(false);
             if (this._handle) {
@@ -97,7 +95,6 @@ export class AutoCollectNativePerformance {
      */
     public dispose(): void {
         this.enable(false);
-        this._statsbeat.removeFeature(Constants.StatsbeatFeature.NATIVE_METRICS);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


AutoCollection/NativePerformance.ts
```diff
@@ -51,6 +54,10 @@ export class AutoCollectNativePerformance {
                 const NativeMetricsEmitters = require("applicationinsights-native-metrics");
                 AutoCollectNativePerformance._emitter = new NativeMetricsEmitters();
                 AutoCollectNativePerformance._metricsAvailable = true;
+                // Should only set statsbeat feature if it's the first time calling enabled
+                if (!this._isEnabled) {
```
ramthi, line 58: isn't isEnabled the config that says if we should collectNative metrics or not?  If you want to send stats only after its enabled in config and if its the first time we are initializing it, it should be checking isInitialized seems like the right thing to do.  That said, AFAICT there is place where we are setting isInitialized=true after initialization!

JacksonWeber: The problem begins where if a user does something like enabling `NativePerformance`, thus setting initialized and enabled to true, then `dispose()` (which only sets enabled to false, but does not set initialized to false) and then calls enable again we end up in a state where initialized is true, but `this._isEnabled` is false. In that case we would want to add the native metrics feature to statsbeat, even though initialization has already happened.

JacksonWeber: I agree that the logic above is confusing though since it's counter intuitive to see a check for `!this._isEnabled` setting the feature on statsbeat. Let me take a look at this logic and see if I can make it a bit easier to read.

Library/QuickPulseStateManager.ts
```diff
@@ -33,12 +35,14 @@ class QuickPulseStateManager {
     private _collectors: { enable: (enable: boolean) => void }[] = [];
     private _redirectedHost: string = null;
     private _pollingIntervalHint: number = -1;
+    private _statsbeat: Statsbeat;
 
-    constructor(config: Config, context?: Context, getAuthorizationHandler?: (config: Config) => AuthorizationHandler) {
-        this.config = config;
+    constructor(client: TelemetryClient, context?: Context, getAuthorizationHandler?: (config: Config) => AuthorizationHandler) {
```
hectorhdzg, line 40: This could be considered a breaking change, new parameters should always be added as optional in the end

Declarations/Constants.ts
```diff
@@ -162,6 +162,8 @@ export enum StatsbeatFeature {
     DISK_RETRY = 1,
     AAD_HANDLING = 2,
     WEB_SNIPPET = 4,
+    NATIVE_METRICS = 8,
```
hectorhdzg, line 165: Native metrics will no be in distro, maybe add it as a large number

AutoCollection/NativePerformance.ts
```diff
@@ -35,6 +37,7 @@ export class AutoCollectNativePerformance {
         }
         AutoCollectNativePerformance.INSTANCE = this;
         this._client = client;
+        this._statsbeat = this._client.getStatsbeat();
```
hectorhdzg, line 40: This could be null and will throw an error

Declarations/Constants.ts
JacksonWeber: Is there any advantage to adding it as a large number over adding it as 8 now and skipping over that value in the distro?

AutoCollection/NativePerformance.ts
```diff
@@ -67,13 +70,15 @@ export class AutoCollectNativePerformance {
 
         // Enable the emitter if we were able to construct one
         if (this._isEnabled && AutoCollectNativePerformance._emitter) {
+            this._statsbeat.addFeature(Constants.StatsbeatFeature.NATIVE_METRICS);
```
hectorhdzg, line 73: Same as before object could be undefined

hectorhdzg (approved):
LGTM

Declarations/Constants.ts
heyams: bit is reserved for a feature, not for which sdk it is for.. sdkVersion will tell you which sdk it is used. skipping it make it hard to understand.

PR #1259, opened 2024-01-10
Fix error when `_shutdownStatsbeat` gets called while being undefined

### LINKED ISSUE #1258: Incorrect `shutdownStatsbeat` optional parameter handling (TypeError: _this._shutdownStatsbeat is not a function)
I've bumped into this issue when using `react-native-windows`.
After a clean project install, I get this error from `Sender.ts` as I try to run it.
```
TypeError: _this._shutdownStatsbeat is not a function
```
 
After some digging, I found out that the `shutdownStatsbeat` parameter is optional, however, there is a case when this callback gets called without a check if it exists or not.

This issue needs to be fixed in this repo.

mkayander:
This PR fixes #1258 issue, a `_shutdownStatsbeat` function should not be called if it is undefined.

The constructor parameter is optional, however, the according class property is not. 
There is a nullability check in one of the methods, but not in the `res.on("end", ...` handler. This causes errors when the callback is not provided.

I've made the class property type nullable as well for convenience, but unfortunately, TS strict mode is turned off in this project. As a result, that kind of nullability issues are not caught by TS.

### DIFF
--- a/Library/Sender.ts
+++ b/Library/Sender.ts
@@ -34,7 +34,7 @@ class Sender {
 
     private _config: Config;
     private _isStatsbeatSender: boolean;
-    private _shutdownStatsbeat: () => void;
+    private _shutdownStatsbeat?: () => void;
     private _failedToIngestCounter: number;
     private _statsbeatHasReachedIngestionAtLeastOnce: boolean;
     private _statsbeat: Statsbeat;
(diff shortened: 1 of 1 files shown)


jonthysell (approved):

JacksonWeber (commented):
LGTM

JacksonWeber (approved):

JacksonWeber:
Looks good, and makes sense that we need to fix the case where that function could be undefined. Thanks for the contribution!

PR #1260, opened 2024-01-18
Clean Up Shim

JacksonWeber:
Cleaning the shim of options no longer supported by the distro API after being updated. 

Including
* OTLP exporter configurations for traces/metrics/logs
* Extended metrics
* Log Instrumentation Options (will be merged into the instrumentation options object)
* Enabling auto collection of external loggers
* Native metrics
* MaxBatchIntervalMs
* Enabling auto collection of performance counters

Resolves bug where if enableAutoCollectDependencies/enableAutoCollectRequests was set to false, all other instrumentation options would be false.

Cleans up the unused `configHelper` class.


README.md
```diff
@@ -199,7 +199,15 @@ The following configurations are set using either environment variables, setting
 | httpAgent/httpsAgent | Not supported in the shim or Azure Monitor OpenTelemetry. |
 | enableWebInstrumentation | Not currently supported in the shim, but is in Azure Monitor OpenTelemetry as `enableBrowserSdkLoader`. |
 | webInstrumentationConnectionString | Not supported in the shim, but is in Azure Monitor OpenTelemetry as `browserSdkLoaderConnectionString`. |
-| webInstrumentationConfig | Not currently supported by the shim, but is in Azure Monitor OpenTelemetry as `browserSdkLoaderConfig`. |
+| webInstrumentationConfig | Not currently supported by the shim, or Azure Monitor OpenTelemetry. |
+| enableAutoCollectPerformance | Not supported in the shim. |
+| enableAutoCollectConsole | Not supported in the shim. |
```
hectorhdzg, line 204: Should be supported in shim

README.md
```diff
@@ -199,7 +199,15 @@ The following configurations are set using either environment variables, setting
...
+| enableAutoCollectExternalLoggers | Not supported in the shim. |
```
hectorhdzg, line 205: Should be supported in shim

README.md
```diff
@@ -199,7 +199,15 @@ The following configurations are set using either environment variables, setting
...
+| enableAutoCollectExceptions | Not supported in the shim. |
```
hectorhdzg, line 206: Do you have a task to follow up on this?, it would be good to check if we can support this

README.md
```diff
@@ -199,7 +199,15 @@ The following configurations are set using either environment variables, setting
 | httpAgent/httpsAgent | Not supported in the shim or Azure Monitor OpenTelemetry. |
 | enableWebInstrumentation | Not currently supported in the shim, but is in Azure Monitor OpenTelemetry as `enableBrowserSdkLoader`. |
 | webInstrumentationConnectionString | Not supported in the shim, but is in Azure Monitor OpenTelemetry as `browserSdkLoaderConnectionString`. |
-| webInstrumentationConfig | Not currently supported by the shim, but is in Azure Monitor OpenTelemetry as `browserSdkLoaderConfig`. |
+| webInstrumentationConfig | Not currently supported by the shim, or Azure Monitor OpenTelemetry. |
+| enableAutoCollectPerformance | Not supported in the shim. |
```
hectorhdzg, line 203: I though we supported perf counters in shim, is the code not there?

src/shim/shim-config.ts
```diff
@@ -138,15 +136,6 @@ class Config implements IConfig {
                 redis4: { enabled: true },
                 postgreSql: { enabled: true },
             },
-            logInstrumentationOptions: {
-                console: { enabled: false },
-                winston: { enabled: true },
-                bunyan: { enabled: true }
-            },
-            otlpTraceExporterConfig: {},
```
hectorhdzg, line 146: Why removing OTLP configurations here? is it being initialized somewhere else not sure I follow

src/shim/shim-config.ts
```diff
@@ -160,23 +149,13 @@ class Config implements IConfig {
         if (this.aadTokenCredential) {
             options.azureMonitorExporterOptions.credential = this.aadTokenCredential;
         }
-        if (typeof (this.enableAutoCollectConsole) === "boolean") {
-            const setting: boolean = this.enableAutoCollectConsole;
-            options.logInstrumentationOptions = {
```
hectorhdzg, line 165: Console OTel instrumentation still doesn't exist, so we cannot remove this

src/shim/shim-config.ts
```diff
@@ -196,16 +176,6 @@ class Config implements IConfig {
                 };
             }
         }
-        if (typeof (this.enableAutoCollectPerformance) === "boolean") {
-            options.enableAutoCollectPerformance = this.enableAutoCollectPerformance;
-        }
-        if (typeof (this.enableAutoCollectExternalLoggers) === "boolean") {
-            options.logInstrumentationOptions = {
-                ...options.logInstrumentationOptions,
-                winston: { enabled: this.enableAutoCollectExternalLoggers },
```
hectorhdzg, line 205: Same here winston still not collecting logs in OpenTelemetry instrumentation

src/shim/shim-config.ts
```diff
@@ -245,47 +210,6 @@ class Config implements IConfig {
                 process.env["APPLICATION_INSIGHTS_NO_STANDARD_METRICS"] = "disable";
             }
         }
-        // NATIVE METRICS
-        if (typeof (this.enableAutoCollectExtendedMetrics) === "boolean") {
```
hectorhdzg, line 249: If you are removing native metrics please remove any related code, including all types, only updating the config is not enough

README.md
JacksonWeber: I see, performance counters appear to be supported in the shim code, although not via the distro's API any longer.

README.md
JacksonWeber: I see performance counters will be, and Console + Winston are not supported in OTel yet, so I'll keep those instrumentations here. But I believe we need to kill `webInstrumentationConfig` since configuring the browser SDK loader is no longer supported in the distro.

README.md
JacksonWeber: It is, API changed on the distro side, but we have local code in the beta.

README.md
hectorhdzg: Yeah but this shim readme not distro

README.md
hectorhdzg: Yeah I didn't say anything about webInstrumentationConfig, we should deprecate that one

JacksonWeber: It's in the highlighted bit in the UI on Github on my side, sorry for the confusion!

src/shim/shim-config.ts
JacksonWeber: Ah, this was a reaction to the API change in the distro, looks like I can modify the internal config of the distro from within `main.ts` so I'll just transition what used to be configurations passed to the distro via its API to modifying the internal config.

PR #1262, opened 2024-01-24
[Shim/Beta] Extend Shim to Support New Distro Features

JacksonWeber:
Extend the shim to support
* Live Metrics
* Browser SDK Loader
* Use distro bunyan

Disables Node 14 support in pipelines for now.
Updated tests to account for changes to OpenTelemetry packages.
Updates OpenTelemetry, distro, and exporter depdendencies.
Updated InstrumentationOptions type to extended that found in the distro and removed LogInstrumentationOptions.


src/shim/shim-config.ts
```diff
@@ -325,6 +329,18 @@ class Config implements IConfig {
             }
         }
 
+        if (this.enableSendLiveMetrics === false) {
```
hectorhdzg, line 332: This kind of check would be broken the moment we change default values, we need to improve the logic here

src/shim/shim-config.ts
```diff
@@ -203,8 +204,11 @@ class Config implements IConfig {
             options.logInstrumentationOptions = {
```
hectorhdzg, line 204: Is a good time to remove logInstrumentationOptions  entirely and only use instrumentationOptions, this what will be supported in distro in the future

src/shim/shim-config.ts
JacksonWeber: Updated logic to first check if the value is a boolean.

src/shim/shim-config.ts
JacksonWeber: Removed the other `logInstrumentationOptions` and cleaned up the tests.

hectorhdzg (approved):
LGTM

Issue #1230, opened 2023-10-16
EEXIST error when using functions_worker_process_count


ejizba:
In Azure Functions, users can have multiple processes on the same instance if they use the `functions_worker_process_count` setting (docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#functions_worker_process_count)). Unfortunately that can lead to the following uncaught exception:

> Error while saving data to disk: [object Error]{ stack: 'Error: EEXIST: file already exists, mkdir 'C:\\local\\Temp\\appInsights-node...

I don't know the best behavior here, but at minimum this error should be handled more gracefully instead of throwing an uncaught exception which can crash the Node.js process.

JacksonWeber:
@ejizba Just wanted to confirm with you that the solution implemented in the above PRs works for you. Thanks.

ejizba:
@JacksonWeber this was a customer incident and I don't know exactly how to repro this issue myself. I tried on an old version of the appinsights package but I didn't see the error. Do have any tips for how to hit this code path? When is that file used?

JacksonWeber:
@ejizba IIRC the only way I could figure out to create the situation required to reproduce this problem was to create an Azure Function with access to multiple threads, and then have a sufficiently resource intensive process run on that function. I wasn't able to reproduce, but the logic behind the PR was to create a naming convention for created cache files that utilizes the PID such that no two processes could attempt to operate on the same file at once, even in the situation the customer described.

ejizba:
Okay I reviewed the PRs and they look good to me. This was only a single incident and it was arguably not even their main problem (high CPU). I think it's safe to close for now and we can just open a new issue if we ever see it happen again.

PR #1267, opened 2024-01-29
Add Exception Handling to File Name for Telemetry Caching

JacksonWeber:
https://github.com/microsoft/ApplicationInsights-node.js/issues/1230

Append the process ID to the file name created for holding disk cached telemetry. This should resolve the issue with multiple Azure Functions cores attempting to read/write/delete the same file when functions are scaled to use multiple cores.

Extended this logic outside of Azure Functions so that in any case where the SDK could be run concurrently we create distinct cache files.

### DIFF
--- a/Library/FileSystemHelper.ts
+++ b/Library/FileSystemHelper.ts
@@ -51,11 +51,10 @@ export const getShallowDirectorySize = async (directory: string): Promise<number
                 totalSize += fileStats.size;
             }
         }
-        return totalSize;
     } catch {
         Logging.warn(`Failed to get directory size for ${directory}`);
-        return totalSize;
     }
+    return totalSize;
 };
 
 /**
@@ -68,15 +67,14 @@ export const getShallowDirectorySizeSync = (directory: string): number => {
         for (let i = 0; i < files.length; i++) {
             totalSize += fs.statSync(path.join(directory, files[i])).size;
         }
-        return totalSize;
     } catch {
         Logging.warn(`Failed to get directory size synchronously for ${directory}`)
-        return totalSize;
     }
+    return totalSize
 }
 
 /**
-* Computes the size (in bytes) of a file asynchronously.
+* Computes the size (in bytes) of a file asynchronously. Returns -1 if the file does not exist.
 */
 export const getShallowFileSize = async (filePath: string): Promise<number> => {
     try {
@@ -86,6 +84,7 @@ export const getShallowFileSize = async (filePath: string): Promise<number> => {
         }
     } catch {
         Logging.warn(`Failed to get file size for ${filePath}`);
+        return -1;
     }
 }
 
--- a/Library/FileSystemHelper.ts
+++ b/Library/FileSystemHelper.ts
@@ -1,6 +1,8 @@
 import * as fs from "fs";
 import path = require("path");
 import { promisify } from "util";
+import Logging = require("./Logging");
+import Util = require("./Util");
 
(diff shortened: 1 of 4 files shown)


Library/FileSystemHelper.ts
```diff
@@ -39,38 +40,52 @@ export const confirmDirExists = async (directory: string): Promise<void> => {
...
+            if (fileStats.isFile()) {
+                totalSize += fileStats.size;
+            }
         }
+        return totalSize;
+    } catch {
+        Logging.warn(`Failed to get directory size for ${directory}`);
+        return totalSize;
     }
-    return totalSize;
```
hectorhdzg, line 52: You can keep this one and remove the other two returns

Library/FileSystemHelper.ts
```diff
@@ -39,38 +40,52 @@ export const confirmDirExists = async (directory: string): Promise<void> => {
...
+        let files = fs.readdirSync(directory);
+        for (let i = 0; i < files.length; i++) {
+            totalSize += fs.statSync(path.join(directory, files[i])).size;
+        }
+        return totalSize;
+    } catch {
+        Logging.warn(`Failed to get directory size synchronously for ${directory}`)
+        return totalSize;
     }
-    return totalSize;
```
hectorhdzg, line 64: Same here

Library/FileSystemHelper.ts
```diff
@@ -39,38 +40,52 @@ export const confirmDirExists = async (directory: string): Promise<void> => {
...
-    if (fileStats.isFile()) {
-        return fileStats.size;
+    try {
+        const fileStats = await statAsync(filePath);
+        if (fileStats.isFile()) {
+            return fileStats.size;
+        }
+    } catch {
+        Logging.warn(`Failed to get file size for ${filePath}`);
     }
```
hectorhdzg, line 89: Should you return something here in failure case?

Library/Sender.ts
```diff
@@ -428,7 +429,7 @@ class Sender {
         try {
              //create file - file name for now is the timestamp, a better approach would be a UUID but that
             //would require an external dependency
-            var fileName = new Date().getTime() + ".ai.json";
+            var fileName = `${new Date().getTime()}.${this._processId}.ai.json`;
```
hectorhdzg, line 432: How many digiths is the processId? I wonder if this could cause issues if the file name gets very long.

hectorhdzg: We may just drop the date now right?, maybe use a UUID like the comment below?

JacksonWeber: Looks like the max length for both Windows and Linux should be 255 characters for a file path. I'll add some tests for long file names, but assuming the above holds true. I'm comfortable using up a max of < 8% of our file name length for the PID.

JacksonWeber: > We may just drop the date now right?, maybe use a UUID like the comment below?

Are you suggesting we just drop the PID & Date approach and transfer entirely to using the UUID package?

Library/FileSystemHelper.ts
JacksonWeber: I'll return -1 if the file does not exist and update the function desc.

Library/Sender.ts
JacksonWeber: Is relying on the external dependency no longer a concern?

hectorhdzg (approved):
LGTM

PR #1268, opened 2024-01-29
[Beta] Detect Distro Already Loaded in Agent

JacksonWeber:
Should detect if the distro package is already present in the RP file system when attempting to attach the agent.

### DIFF
--- a/src/agent/agentLoader.ts
+++ b/src/agent/agentLoader.ts
@@ -236,24 +236,28 @@ export class AgentLoader {
     private _sdkAlreadyExists(): boolean {
         try {
             // appInstance should either resolve to user SDK or crash. If it resolves to attach SDK, user probably modified their NODE_PATH
-            let appInstance: string;
+            let sdkAppInstance: string;
+            let distroAppInstance: string;
             try {
                 // Node 8.9+ Windows
                 if (this._isWindows) {
-                    appInstance = (require.resolve as any)("applicationinsights", { paths: [process.cwd()] });
+                    sdkAppInstance = (require.resolve as any)("applicationinsights", { paths: [process.cwd()] });
+                    distroAppInstance = (require.resolve as any)("@azure/monitor-opentelemetry", { paths: [process.cwd()] });
                 }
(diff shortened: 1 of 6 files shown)


src/agent/agentLoader.ts
```diff
@@ -236,24 +236,28 @@ export class AgentLoader {
     private _sdkAlreadyExists(): boolean {
         try {
             // appInstance should either resolve to user SDK or crash. If it resolves to attach SDK, user probably modified their NODE_PATH
-            let appInstance: string;
+            let sdkAppInstance: string;
```
hectorhdzg, line 239:
```suggestion
            let shimInstance: string;
```

src/agent/agentLoader.ts
```diff
@@ -236,24 +236,28 @@ export class AgentLoader {
...
-                appInstance = require.resolve(`${process.cwd()}/node_modules/applicationinsights`);
+                sdkAppInstance = require.resolve(`${process.cwd()}/node_modules/applicationinsights`);
+                distroAppInstance = require.resolve(`${process.cwd()}/node_modules/@azure/monitor-opentelemetry`);
             }
             // If loaded instance is in Azure machine home path do not attach the SDK, this means customer already instrumented their app
-            if (appInstance.indexOf("home") > -1) {
+            if (sdkAppInstance.indexOf("home") > -1 || distroAppInstance.indexOf("home") > -1) {
                 const diagnosticLog: IDiagnosticLog = {
-                    message: `Azure Monitor Application Insights Distro already exists. Module is already installed in this application; not re-attaching. Location: ${appInstance}`,
+                    message: `Azure Monitor Application Insights Distro already exists. Module is already installed in this application; not re-attaching. Location: ${sdkAppInstance || distroAppInstance}`,
```
hectorhdzg, line 260: It would be good to differentiate the message depending which package was detected.

JacksonWeber: I'll make that more clear.

hectorhdzg (approved):
LGTM

Issue #1269, opened 2024-01-30
[3.0.0-beta.11] Incorrect types for `useAzureMonitor`?


mderriey:
Hi 👋

We've updated to 3.0.0-beta.11 to take advantage of errors being correctly serialised.

However, we're facing an issue with types.
If we understand correctly, this library extends [the `InstrumentionOptions` type from `@azure/opentelemetry-monitor](https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/src/types.ts#L39-L44) so we can specify instrumentations from both the OTel distro and this library in a single object.

However, this type isn't used, so this code doesn't compile with TypeScript:

```typescript
const resource = Resource.EMPTY
resource.attributes[SemanticResourceAttributes.SERVICE_NAME] = 'service-name'
resource.attributes[SemanticResourceAttributes.SERVICE_NAMESPACE] = 'service-namespace'

useAzureMonitor({
  resource,
  instrumentationOptions: {
    azureSdk: { enabled: false },
    http: { enabled: false },
    mongoDb: { enabled: false },
    mySql: { enabled: false },
    postgreSql: { enabled: false },
    redis: { enabled: false },
    redis4: { enabled: false },
    winston: { enabled: true },
    console: { enabled: false },
    bunyan: { enabled: false },
  }
})
```

The TypeScript error is:

```
TS2353: Object literal may only specify known properties, and winston does not exist in type InstrumentationOptions
```

We could only fix it by juggling types as below:

```typescript
import type { InstrumentationOptions as ApplicationInsightsInstrumentationOptions } from 'applicationinsights/out/src/types'
import type { InstrumentationOptions as OpenTelemetryInstrumentationOptions } from '@azure/monitor-opentelemetry'

// Default attributes set on the OpenTelemetry traces
const resource = Resource.EMPTY
resource.attributes[SemanticResourceAttributes.SERVICE_NAME] = 'service-name'
resource.attributes[SemanticResourceAttributes.SERVICE_NAMESPACE] = 'service-namespace'

useAzureMonitor({
  resource,
  instrumentationOptions: {
    azureSdk: { enabled: false },
    http: { enabled: false },
    mongoDb: { enabled: false },
    mySql: { enabled: false },
    postgreSql: { enabled: false },
    redis: { enabled: false },
    redis4: { enabled: false },
    winston: { enabled: true },
    console: { enabled: false },
    bunyan: { enabled: false },
[...]

PR #1270, opened 2024-01-30
[Beta] Update Instrumentation Options Type in Docs and Export from Index

JacksonWeber:
`AzureMonitorOpenTelemetryOptions` extends `DistroOptions` which contains the `InstrumentationOptions` interface, which we cannot modify. Therefore we need to cast `DistroInstrumentationOptions` passed to the shim config by the client to `InstrumentationOptions` which includes the console and winston instrumentation options.

#1269

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ npm install applicationinsights@beta
 
 
 ```typescript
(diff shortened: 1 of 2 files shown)


hectorhdzg (approved):
LGTM

PR #1271, opened 2024-02-01
[Beta] Update Prefix Version in Auto-Attach Scenarios

JacksonWeber:
We should include the "i" prefix for integrated-auto attach senarios. Our new prefixes will look something like: "awi_" which would correspond to running in App Service, on Windows, and with an attach type of "integrated auto" (auto-attach).


src/agent/aksLoader.ts
```diff
@@ -50,7 +50,7 @@ export class AKSLoader extends AgentLoader {
                     }
                 )
             );
-            process.env[AZURE_MONITOR_AGENT_PREFIX] = this._getVersionPrefix(AgentResourceProviderType.aks);
+            process.env[AZURE_MONITOR_AUTO_ATTACH] = "true";
```
hectorhdzg, line 53: You can add this in AgentLoader class

hectorhdzg (approved):
LGTM

PR #1275, opened 2024-02-05
[Beta] Add Note Regarding Our Support for Node Versions

JacksonWeber:
Note that we will only support versions of node supported by OpenTelemetry in the beta README.


hectorhdzg (approved):
LGTM

PR #1276, opened 2024-02-11
Update Statsbeat Network Metric Names

JacksonWeber:
Updated the network statsbeat metric names to match those sent by the exporter.


hectorhdzg (approved):
LGTM

PR #1278, opened 2024-02-14
Update SDK Version Prefix

JacksonWeber:
* Updates the prefix to be sent in manual attach situations.
* Updates the prefix to include a third bit either `m` or `i` indicating manual or integrated auto attach type respectively.


hectorhdzg (approved):
LGTM

PR #1280, opened 2024-02-16
Revert Linux App Service SDK Already Exists Changes

JacksonWeber:
- Revert bad check for Linux App Service SDK already attached. (#1235)
- Pin version of the web-snippet-loader package.


hectorhdzg (approved):
LGTM

PR #1282, opened 2024-02-17
[Beta] Fix Auto Exception Collection

JacksonWeber:
Exceptions were not being captured before process exit. This PR unifies how we handle `uncaughtException` and `unhandledRejection` events between all Node.js versions, as the `uncaughtExceptionMonitor` event listener was behaving unpredictably and not respecting the fulfillment of the `.forceFlush()` function before exiting program execution.


hectorhdzg (approved):
LGTM

PR #1292, opened 2024-02-29
[Beta] Update Warning Tests and Add Console Logging by Default

JacksonWeber:
* Fixes our console warning tests to be updated using the `diag` logger from OTel.
* Automatically sets the default logger in the shim to the console.

### DIFF
--- a/src/shim/applicationinsights.ts
+++ b/src/shim/applicationinsights.ts
@@ -9,6 +9,8 @@ import { CorrelationContextManager } from "./correlationContextManager";
 import { ICorrelationContext, HttpRequest, DistributedTracingModes } from "./types";
 import { TelemetryClient } from "./telemetryClient";
 import * as Contracts from "../declarations/contracts";
+import { Util } from "../shared/util";
+import { useAzureMonitor } from "@azure/monitor-opentelemetry";
 
 
 // We export these imports so that SDK users may use these classes directly.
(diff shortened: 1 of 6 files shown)


src/shim/applicationinsights.ts
```diff
@@ -139,6 +148,7 @@ export class Configuration {
      * @param collectExtendedMetrics if true, extended metrics counters will be collected every minute and sent to Application Insights
      * @returns {Configuration} this class
      */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars 
     public static setAutoCollectPerformance(value: boolean, collectExtendedMetrics: any) {
         if (defaultClient) {
```
hectorhdzg, line 153: Do we need to log an unsupported warning here when collectExtendedMetrics is provided?

src/shim/telemetryClient.ts
```diff
@@ -30,13 +30,14 @@ export class TelemetryClient {
     public commonProperties: { [key: string]: string };
     public config: Config;
     private _options: AzureMonitorOpenTelemetryOptions;
+    public configWarnings: string[] = [];
```
hectorhdzg, line 33: Update this to be a private method and create a add warning method here so is easy to understand where this is being updated

hectorhdzg (approved):
LGTM

PR #1293, opened 2024-02-29
Fix Linux Attach Backoff Logic

JacksonWeber:
* Check for the user application insights package path in Linux attach scenarios when evaluating if we should backoff during auto-attach.

### DIFF
--- a/Bootstrap/Helpers.ts
+++ b/Bootstrap/Helpers.ts
@@ -1,7 +1,8 @@
+import { isLinux } from "../Library/PrefixHelper";
 import { DiagnosticLog, DiagnosticMessageId } from "./DataModel";
 import { DiagnosticLogger } from "./DiagnosticLogger";
 
-const USER_APP_PATH = "/home/site/wwwroot";
+const LINUX_USER_APPLICATION_INSIGHTS_PATH = "/node_modules/applicationinsights/out/applicationinsights.js";
 
 export function sdkAlreadyExists(_logger: DiagnosticLogger): boolean {
     try {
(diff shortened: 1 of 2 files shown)


Bootstrap/Helpers.ts
```diff
@@ -1,46 +1,44 @@
...
-            }
+        }
+        /** 
+         * If loaded instance is in Azure machine home path do not attach the SDK, this means customer already instrumented their app.
+         * Linux App Service doesn't append the full cwd to the require.resolve, so we need to check for the relative path we expect
+         * if application insights is being imported in the user app code.
+        */
+        if (
+            appInstance.indexOf("home") > -1 ||
+            (appInstance === LINUX_USER_APPLICATION_INSIGHTS_PATH && isLinux())
```
hectorhdzg, line 25: I don't fully understand this check, in normal circumstances when ai is installed in customer app,is not app instance part of home path?

JacksonWeber: It is under the `home` path, but `require.resolve()` behaves differently in Linux vs. Windows app services. in Windows it returns the full path, something like `/home/site/wwwroot/<path-to-user-instance-of-app-insights-package>` but in Linux it returns just the path starting from under `wwwroot`, which is why our original check wouldn't back off in cases where the customer had instrumented their code.

This new check is just aware of the fact that Linux behaves that way and knows that the `/node_modules/applicationinsights/out/applicationinsights.js` path is always from under `/home`. In cases where `require.resolve()` recognizes the other agent in the app service it'll be under some `/usr/...` path, which won't stop our attach.

PR #1300, opened 2024-03-21
[Beta] Clean Up Shim Files & Fix Check for Extended Metrics Env Var

JacksonWeber:


src/shim/shim-jsonConfig.ts
```diff
@@ -77,7 +77,7 @@ export class ShimJsonConfig implements IJsonConfig {
     constructor() {
         // Load environment variables first
         this.connectionString = process.env[ENV_connectionString];
-        this.disableAllExtendedMetrics = !!process.env[ENV_nativeMetricsDisableAll];
+        this.disableAllExtendedMetrics = process.env[ENV_nativeMetricsDisableAll] !== undefined ? !!process.env[ENV_nativeMetricsDisableAll] : undefined;
```
hectorhdzg, line 80: Why setting this value if feature is not supported?

hectorhdzg: Maybe some extra code cleaning is needed here

JacksonWeber: I was reading it here to alert users later that the configuration option isn't supported, but I can clean it up a bit by just checking if the value is populated at all in the check later on.

JacksonWeber: Refactored to remove logic parsing the env var input and simply check if they're populated at all to throw the warning. I've also cleaned up any reference to the extended metrics in the standard beta flow - only parsing the values in the shim so we can check to throw the warning once the JSON & standard shim configs have been unified.

hectorhdzg: My comment was more about cleaning dead code, I can see https://github.com/microsoft/ApplicationInsights-node.js/blob/beta/src/metrics/nativeMetrics.ts still there, we should remove all dead code and this kind of PRs are the perfect opportunity to clean up related files

hectorhdzg (approved):
LGTM

src/shim/shim-jsonConfig.ts
JacksonWeber: Ah, I see. Killed the `NativeMetrics` class and our dependency on the native-metrics library. The corresponding package-lock file will be fixed in the beta 12 release PR since all of the OTel package updates live there.

PR #1312, opened 2024-04-22
Fix Telemetry To Log Conversion & Update OTel Deps

JacksonWeber:
Sending Log bodies as a string was breaking with the Azure SDK's serialization (being broken apart into individual objects per letter of the string). Fix upgrades to the latest OTel logs instrumentations which now allow for passing an object as the Log body, thereby fixing the log body serialization in Azure SDK.


hectorhdzg (approved):
LGTM

PR #1313, opened 2024-04-23
3.0.1 Release

JacksonWeber:


src/types.ts
```diff
@@ -7,7 +7,7 @@ import { InstrumentationConfig } from "@opentelemetry/instrumentation";
 import { OTLPExporterNodeConfigBase } from "@opentelemetry/otlp-exporter-base";
 
 
-export const AZURE_MONITOR_OPENTELEMETRY_VERSION = "1.3.0";
+export const AZURE_MONITOR_OPENTELEMETRY_VERSION = "1.4.0";
```
hectorhdzg, line 10: How this is being used? do we want to use distro version or something else? in this case should be 3.0.1 right?

hectorhdzg (approved):

src/types.ts
JacksonWeber: Looks to be used for two different things through the project that conflict with one another, setting the ""AZURE_MONITOR_DISTRO_VERSION" env var (used in statsbeat in the exporter, and also already handled by the distro which will always be loaded anyway), and to set the sdkVersion in this project (which like you mentioned, should be 3.0.1). I've removed the duplicate logic with the distro that sets the DISTRO_VERSION env var and renamed this constant to be more obvious in meaning.

PR #1314, opened 2024-04-23
Update Tracer Provider Typing

JacksonWeber:
Tracer provider being of the wrong type would cause the flush method to error with a message saying that tracerProvider had no forceFlush() method.


hectorhdzg (approved):
LGTM

PR #1319, opened 2024-04-26
Add Collection of Metrics when Measurements are passed to Events

JacksonWeber:
* When one or more measurements are passed to manual `.trackEvent()` calls, we should collect those as individual metrics.
* Added tests to ensure that metrics are created when measurements are passed.


JacksonWeber:
Closing in favor of the solution in the exporter: https://github.com/Azure/azure-sdk-for-js/pull/29511

PR #1325, opened 2024-05-05
Update Performance Counter Names

JacksonWeber:
Update perf counter names to OTel valid names that will be converted in the exporter to those that are valid for Breeze.


hectorhdzg (approved):
LGTM

PR #1326, opened 2024-05-10
Fix Default Logger Settings and Add Console Log Level Support

JacksonWeber:
Updated defaults such that:
* Winston & bunyan are on by default in general
* Console log collection is off by default
* Console log collection is on by default in AKS attach scenario

Also added support for Console logging log level.

### DIFF
--- a/src/agent/agentLoader.ts
+++ b/src/agent/agentLoader.ts
@@ -69,6 +69,9 @@ export class AgentLoader {
                     redis: {
                         enabled: true
                     },
+                    bunyan: {
+                        enabled: true
+                    },
                 }
             };
 
@@ -259,10 +262,10 @@ export class AgentLoader {
                 distroInstance = require.resolve(`${process.cwd()}/node_modules/@azure/monitor-opentelemetry`);
             }
             /** 
-         * If loaded instance is in Azure machine home path do not attach the SDK, this means customer already instrumented their app.
-         * Linux App Service doesn't append the full cwd to the require.resolve, so we need to check for the relative path we expect
-         * if application insights is being imported in the user app code.
-        */
+             * If loaded instance is in Azure machine home path do not attach the SDK, this means customer already instrumented their app.
+             * Linux App Service doesn't append the full cwd to the require.resolve, so we need to check for the relative path we expect
+             * if application insights is being imported in the user app code.
+            */
             if (
                 shimInstance.indexOf("home") > -1 || distroInstance.indexOf("home") > -1 ||
                 (shimInstance === LINUX_USER_APPLICATION_INSIGHTS_PATH && this._isLinux)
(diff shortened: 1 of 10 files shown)


src/agent/agentLoader.ts
```diff
@@ -69,6 +69,9 @@ export class AgentLoader {
                     redis: {
                         enabled: true
                     },
+                    bunyan: {
```
hectorhdzg, line 72: Looks like you are missing winston here, right? default should be false for all other agents, true only for AKS

src/shared/configuration/config.ts
```diff
@@ -69,15 +72,13 @@ export class ApplicationInsightsConfig {
             redis: { enabled: false },
             redis4: { enabled: false },
             console: { enabled: false },
-            bunyan: { enabled: false },
-            winston: { enabled: false },
+            bunyan: { enabled: true },
```
hectorhdzg, line 75: Default values should change only for AKS

src/agent/agentLoader.ts
JacksonWeber: Removed here and added to where winston is also set in the AKS loader. New functionality should fall into 3 categories that I've tested:
1. Non-AKS auto attach - all instrumentations that aren't logging are enabled, but console, winston, and bunyan are disabled.
2. AKS auto attach - instrumentations and logging is enabled by default.
3. Manual instrumentation - all instrumentations, bunyan, and winston are enabled by default, but console is disabled.

src/shared/configuration/config.ts
```diff
@@ -109,6 +110,19 @@ export class ApplicationInsightsConfig {
...
+                this.instrumentationOptions = {
+                    ...this.instrumentationOptions,
+                    console: {
+                        ...this.instrumentationOptions.console,
+                        logSendingLevel: logLevelParser(process.env[loggingLevel]),
+                    },
+                }
+            }
+
+            this._mergeConfig();
```
hectorhdzg, line 125: This looks like a very risky change, what is going on here with config prioritization? why making this change now?

JacksonWeber: I believe this is effectively a bug today because the `mergeConfig()` method parses the JSON config and environment variables. Env vars need to take precedence over standard in-code configs so that auto instrumentation fixes like setting the `APPLICATIONINSIGHTS_CONFIGURATION_CONTENT` can override any default configs that may already be set in auto-instrumentation scenarios (this is why setting the `console: {enabled: false}` in that env var wasn't working originally).

This also deviates from the way that the distro works, so we end up in a situation where env vars take precedence in the distro (https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/monitor/monitor-opentelemetry/src/shared/config.ts#L111) but not in App Insights. So managed log configurations (winston & bunyan) work as expected while App Insights managed console cannot be correctly overridden by the env var.

hectorhdzg: Sounds good, we need to validate JSON config before releasing agent, this should be a key scenario to support

hectorhdzg (approved):
LGTM

PR #1334, opened 2024-05-23
Update SDK Already Attached Check

JacksonWeber:
* The SDK Already Attached Check should look for the exporter instead of the distro package such that we don't attach in cases where the customer already has a dependency on either the exporter or the distro.


hectorhdzg (approved):
LGTM

PR #1337, opened 2024-05-30
Update Sampling Percentage to Accept 0 as Valid

### LINKED ISSUE #1336: Unable to Disable Logging by Setting samplingRate to Zero
In certain scenarios, it is useful to completely suppress logging by setting the `samplingRate` to zero. However, the current code prevents this by taking the default value when `samplingRate` is set to zero, as shown in the following segment:

https://github.com/microsoft/ApplicationInsights-node.js/blob/main/src/shared/configuration/config.ts#L112

To address this, we need to modify the code to allow samplingRate to be set to zero without reverting to the default value. This change will provide greater flexibility in controlling logging behavior.

JacksonWeber:
* Previously setting the sampling percentage as 0 would actually set sampling to 1 as 0 is falsy and would revert to the default value of the sampling percentage.

Resolves #1336

### DIFF
--- a/src/shared/configuration/config.ts
+++ b/src/shared/configuration/config.ts
@@ -112,6 +112,7 @@ export class ApplicationInsightsConfig {
             if (typeof(options.samplingRatio) === "number") {
                 this.samplingRatio = options.samplingRatio;
             }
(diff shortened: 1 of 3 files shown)


hectorhdzg (approved):
LGTM

PR #1340, opened 2024-06-04
Update Live Metrics to be on by Default

JacksonWeber:
* Updates the live metrics setting to be on by default in RP auto-attach scenarios.
* Updates imports from the sem conv package.


hectorhdzg (approved):
LGTM

PR #1362, opened 2024-07-12
Fix Crashing with Old Node.JS Versions When Manually Instrumenting

JacksonWeber:
Fix node.js versions that are not supported by the Azure SDK by conditionally loading Azure SDK packages when manually instrumenting.

https://github.com/Azure/azure-sdk-for-js/issues/30217


Library/AuthorizationHandler.ts
```diff
@@ -1,32 +1,42 @@
...
-import azureCore = require("@azure/core-rest-pipeline");
+import { TokenCredential } from "@azure/core-auth";
+import { PipelineRequest, PipelineResponse, PipelinePolicy } from "@azure/core-rest-pipeline";
+import Logging = require("./Logging");
 
 const applicationInsightsResource = "https://monitor.azure.com//.default";
 
+let azureCore: any;
+try {
+  azureCore = require("@azure/core-rest-pipeline");
```
hectorhdzg, line 11: This package is needed to send the actual telemetry to Breeze, if not available or not functional the SDK would do nothing, so this is a not necessary check.

README.md
```diff
@@ -32,7 +32,8 @@ You can manually track more aspects of your app and system using the API describ
 
 
 
-## Supported Node.JS versions 
+## Supported Node.JS versions
+> *Note:* Versions of Node.JS that are not supported by the [JS Azure SDK](https://github.com/Azure/azure-sdk-for-js/blob/main/SUPPORT.md) will not support AAD authentication.
```
hectorhdzg, line 36: This is a little bit confusing, we need azure/identity package to be included to use the feature, as customer need to pass an instance of a Credential object, so this extra note could cause more confusion than helping here.

package-lock.json
```diff
@@ -51,34 +51,25 @@
         }
       }
     },
-    "node_modules/@aashutoshrathi/word-wrap": {
```
hectorhdzg, line 54: package-lock.json Update added by mistake here?

Bootstrap/Default.ts
```diff
@@ -1,5 +1,4 @@
-import * as azureCoreAuth from "@azure/core-auth";
-
+import type { TokenCredential } from "@azure/core-auth";
```
hectorhdzg, line 1: Please rename the PR title to something that reflect the actual change

README.md
JacksonWeber: I'll reword to make it clearer what the exception to our support for older versions really is.

JacksonWeber: I moved the note to the aadCredential description. Hopefully that makes it clearer that old versions of Node don't impact using any other scenario with the SDK.

package-lock.json
JacksonWeber: Fixed.

Bootstrap/Default.ts
JacksonWeber: Updated.

hectorhdzg (approved):

PR #1366, opened 2024-08-14
[v2.x] Update @azure/core-rest-pipeline to latest

hectorhdzg:
There are issues with token refresh that should be resolved by using latest token acquisition and refresh code dependency


JacksonWeber (approved):
Thanks for noticing and cleaning up the unused core-util package too.

PR #1377, opened 2024-09-04
Bug Fix - Not add log processor repeatedly when there is any config warning

zhiyuanliang-ms:
The attribute log processor should only be initialized once. This should be a logic bug.


zhiyuanliang-ms:
@hectorhdzg @JacksonWeber @OsvaldoRosado Could you help review this PR?

JacksonWeber (approved):
Thanks for catching this unnecessary re-initialization of the loggerProvider in our console warning solution. LGTM

PR #1382, opened 2024-09-11
Allow Distro config to be specified

hectorhdzg:
JSON config should be covered already but JS options was not available


harsimar (approved):
LGTM

PR #1387, opened 2024-09-17
Remove unused undici import

### LINKED ISSUE #1386: Undeclared dependency “undici”
Hi,

i’m using Yarn pnp and after updating the 3.3.0 today O got an error that it could not resolve the “undici” package.

checking the source here it is imported in src/shim/correlationContextManager but is not defined in the package.json file and thus leading to yarn pnp disallowing our build to happen as we don’t allow undefined dependencies in our build.

hectorhdzg:
Apparently this was added by mistake as Headers is already defined in node types, removing to fix #1386


JacksonWeber:
I believe the reason the Node 16 tests are failing is that `instanceof` doesn't just use the type (which we could get from the @node/types 18.0.0 package) but requires a value, which explains why pulling the definition of the Header class from `undici` fixed this in Node 16 environments. We can either have an explicit dependency on `undici` (it's already pulled in as a dependency of the @azure/functions package) to solve the issue with yarn, or I can define a type-guard to check for the properties of the `Headers` type without the import.

package-lock.json
JacksonWeber, line 1: Can we remove undici-types entirely here now that we're not using the `Headers` type?

hectorhdzg: We are not directly referencing this package, is coming from a dependency so we cannot remove it

JacksonWeber (approved):
LGTM

PR #1416, opened 2025-02-28
Add setAzureMonitorOptions() to Configuration

huang-julien:
Hello :wave:

I'm coming from applicationinsights 2 and i like that we still have a shim in v3 while moving to OTEL.

This PR aims to improve a bit the DX by adding a `setAzureMonitorOptions()` static method to `Configuration` .
The reason is that those who upgraded from v2 to v3 may still use `setup()` which returns the `Configuration` class. In order to modify AZure Monitors options, users would have to do something like:

```ts
const configuration = setup('connection-string')
defaultClient.config.azureMonitorOpenTelemetryOptions = { /** ... **/ }
configuration.start()
```

this PR would allow to do this instead

```ts
setup('connection-string')
  .setAzureMonitorOptions({ /** ... **/ })
  .start()
```


JacksonWeber:
@huang-julien Just so I can understand your use-case a little better. You're adopting modifying the `azureMonitorOpenTelemetryOptions` directly, so I'm interested in why you're not migrating to the [Azure Monitor OpenTelemetry distro](https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/monitor/monitor-opentelemetry). Are you just using the shim for access to the `track` APIs?

huang-julien:
Yup I am. I'm actually the maintainer of [nitro-applicationinsights](https://github.com/huang-julien/nitro-applicationinsights) and [nuxt-applicationinsights](https://github.com/huang-julien/nuxt-applicationinsights).

The shim is very conveniant for SSR frameworks as it provide quasi-identical methods between appinsights node and appinsights web which allows for devs (such as Nuxt) to be able to call `track*` methods through a single property.

Issue #1422, opened 2025-04-16
setAutoCollectDependencies not working after upgrading from v2 to v3


qb-raj-lvmh:
Hi,

I'm working on migrating existing functions from Azure Function v3 to v4 along with node,js version upgrade. As part of this process, I've upgraded the package applicationinsights library from version "2.9" to "3.5". But after the upgrade, it seems the existing setAutoCollectDependencies false property doesn't work and it is creating hude logs for dependency logs.

`import * as appInsights from 'applicationinsights';

export class AppInsights implements ILoggerService {
	private readonly client: appInsights.TelemetryClient;
	constructor() {
	this.client = appInsights.defaultClient;

		
			appInsights
				.setup()
				.setAutoCollectDependencies(false) // seems not working.
				.setAutoCollectPerformance(false, false)
				.start();
			this.client = appInsights.defaultClient;
	
	}

	// Track Events
	/**
	 *
	 * @param trackEvent
	 */
	TrackEvent(trackEvent: IEventLog): void {
		this.client.trackEvent(trackEvent);
	}

	// Track Metric
	/**
	 *
	 * @param trackMetric
	 */

	TrackMetric(trackMetric: IMetricLog): void {
		this.client.trackMetric(trackMetric);
	}

	// Track Trace
	/**
	 *
	 * @param trackMessage
	 */
	TrackTrace(trackMessage: ITraceLog): void {
		this.client.trackTrace(trackMessage);
	}

	// Track Exception
	/**
	 *
	 * @param trackException
	 */
	TrackException(trackException: IExceptionLog): void {
		this.client.trackException(trackException);
	}

	// Track Dependency
	/**
	 *
	 * @param trackDependency
	 */
	TrackDependency(trackDependency: IDependencyLog): void {
		this.client.trackDependency(trackDependency);
	}

	// Track Request
	/**
	 *
	 * @param trackRequest
	 */
	TrackRequest(trackRequest: appInsights.Contracts.RequestTelemetry): void {
		this.client.trackRequest(trackRequest);
	}
}
`
My host.json is as follows
`{
	"version": "2.0",
	"logging": {
		"applicationInsights": {
			"samplingSettings": {
				"isEnabled": true,
				"excludedTypes": "Request;Dependency" // added Dependency here not working 
			},
			"enableDependencyTracking": false,
			"enablePerformanceCountersCollection": false
		},
// tried with adding below key and not working 
		"logLevel": {
			"default": "Warning",
[...]

PR #1425, opened 2025-04-28
Add Shim Support for Full Request and Dependency Filtering

JacksonWeber:
https://github.com/microsoft/ApplicationInsights-node.js/issues/1422

Adds support for disabling requests and dependencies across all instrumentations rather than just HTTP.

### DIFF
--- a/src/traces/requestProcessor.ts
+++ b/src/traces/requestProcessor.ts
@@ -1,0 +1,44 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Context, SpanKind, TraceFlags } from "@opentelemetry/api";
+import { ReadableSpan, Span, SpanProcessor } from "@opentelemetry/sdk-trace-base";
+
+/**
+ * Azure Monitor Incoming & Outgoing Request Processor.
+ * @internal
+ */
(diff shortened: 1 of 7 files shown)


src/traces/requestProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    forceFlush(): Promise<void> {
+        return Promise.resolve();
+    }
+
+    onStart(span: Span, _context: Context): void {
+        return;
+    }
+
+    onEnd(span: ReadableSpan): void {
```
MSNev, line 28: Based on the specification you are "not allowed" to alter the span or the span context in any way during the `onEnd` as this is "supposed" to be a readable span (not a mutable one).

There is a new `onEnding` which receives a "Mutable" Span to allow the setting / changing like this -- not sure if OTel-JS has this yet

MSNev: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onending

JacksonWeber: While `onEnding` doesn't appear to be implemented on the JS side, I've implemented and tested this logic in the `onStart()` method with good results.

MSNev (approved):

PR #1436, opened 2025-06-19
Remove Explicit Creation of Metric Exporter in AKS Auto-Attach Scenario

JacksonWeber:
Per the spec https://github.com/aep-health-and-standards/Telemetry-Collection-Spec/blob/main/ApplicationInsights/AKS_AutoAttach_Metrics.md#sending-otel-metrics-to-amw the OTEL_METRICS_EXPORTER env var should be populated when in AKS auto-attach scenarios. 

The OTel code that controls creation of the metric exporter in the even that this env var is populated exists here: https://github.com/open-telemetry/opentelemetry-js/blob/dac72912b3a895c91ee95cfa39a22a916411ba4c/experimental/packages/opentelemetry-sdk-node/src/sdk.ts#L115

### DIFF
--- a/src/agent/aksLoader.ts
+++ b/src/agent/aksLoader.ts
@@ -14,11 +14,6 @@ export class AKSLoader extends AgentLoader {
     constructor() {
         super();
         if (this._canLoad) {
-            // AKS specific configuration
-            this._options.otlpMetricExporterConfig = {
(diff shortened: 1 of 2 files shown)


hectorhdzg (approved):
LGTM

PR #1511, opened 2026-07-23
Replace diagnostic-channel console collection with @opentelemetry/instrumentation-console

JacksonWeber:
## Summary

Replaces the legacy `diagnostic-channel` / `diagnostic-channel-publishers` based console log collection with the OpenTelemetry [`@opentelemetry/instrumentation-console`](https://www.npmjs.com/package/@opentelemetry/instrumentation-console) package, which is now shipped and configured by the `@azure/monitor-opentelemetry` distro.

## Changes

- **Removed `src/logs/autoCollectLogs.ts`**, `src/logs/diagnostic-channel/console.sub.ts`, and `src/logs/diagnostic-channel/initialization.ts`. Console collection is no longer implemented in this repo.
- **`package.json`**: bumped `@azure/monitor-opentelemetry` to `^1.19.0` (which pulls in `@opentelemetry/instrumentation-console`) and dropped the `diagnostic-channel` / `diagnostic-channel-publishers` dependencies.
- **`src/main.ts`**: instead of instantiating `AutoCollectLogs`, `useAzureMonitor` now forwards the resolved console options to the distro via `instrumentationOptions.console`, mapping the public `console.logSendingLevel` option to the package's `logSeverity`. The `autoCollectLogs.shutdown()` calls were removed — the distro's `shutdownAzureMonitor()` disables the instrumentation and restores `console`.
- **`src/types.ts`**: `AzureMonitorOpenTelemetryOptions` now explicitly declares `instrumentationOptions`.
- **`src/shared/configuration/config.ts`**: `APPLICATIONINSIGHTS_INSTRUMENTATION_LOGGING_LEVEL` no longer overwrites a `logSendingLevel` that was set explicitly in code; the env var is only applied when the code value is `undefined`.
- **Deprecations / docs**: `noDiagnosticChannel` and `noPatchModules` are marked `@deprecated` in `src/shim/shim-config.ts` and `src/shim/types.ts`, and the README wording for those options (plus the "patches" → "instrumentations" language) was updated accordingly.
- **CI**: removed the `diagnostic-channel-publishers` install step from the Windows workflow.
- **Tests**: `test/unitTests/logs/console.tests.ts` was rewritten to assert on the config forwarded to the distro and to drive real `console.error` calls, plus a regression test that `console` is restored after shutdown. `aksLoader.tests.ts` now shuts down Azure Monitor between tests so the patched `console` and global providers don't leak into later suites. `main.tests.ts` gained a `getSpanProcessors` helper for the OpenTelemetry 2.x SDK property move. Functional test expectations were updated (`ConsoleError` now expects `Error`, and the `/diagChannelConsole` sequence entry was renamed to `/console`).
- Added a CHANGELOG entry.

## Behavioral note

Severity is now derived from the console method (`console.error` → ERROR, `console.warn` → WARN, `console.log`/`info` → INFO, etc.) instead of the old message-content / `stderr` heuristics. Notably, `console.error` now produces `Error` severity where it previously produced `Warning`.

## Validation

- `npm run build` and `npm run lint` pass with 0 errors.
- Full unit suite passes, including the new console forwarding + shutdown-restore coverage.
- All CI checks (build on Node 20/22/24, Windows + Ubuntu, integration, perf-regression) are green.


test/unitTests/agent/aksLoader.tests.ts
```diff
@@ -20,7 +19,6 @@ describe("agent/AKSLoader", () => {
     });
 
     afterEach(() => {
-        disposeConsole();
         process.env = originalEnv;
         sandbox.restore();
     });
```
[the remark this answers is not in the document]
JacksonWeber: Good catch - this was a real leak, and it was broader than the one test.

Fixed in b2ac0e8. Rather than restore cleanup for just the `initialize` test, I traced what was actually escaping the suite with a mocha root hook that compares `console.log` identity after every test:

- Before: `console.log` stayed patched with `patchedConsoleMethod` from test 7 onward through the end of the file.
- Two tests call `agent.initialize()`, not one - `initialize` and `initialize with OTLP metric reader creates multiple metric collectors`. The second was the one leaking.

I first tried an explicit `azureMonitorInitialized` flag set at the call site, but that only fixed the first test and would silently regress again whenever someone adds another `initialize()` test. Instead `afterEach` now detects the patch itself:

\\\	s
beforeEach(() => {
    originalEnv = process.env;
    originalConsoleLog = console.log;
});

afterEach(async () => {
    if (console.log !== originalConsoleLog) {
        await shutdownAzureMonitor();
    }
    process.env = originalEnv;
    sandbox.restore();
});
\\\

This is self-maintaining and self-clearing: it covers both current call sites plus any added later, and once `shutdownAzureMonitor()` restores console the condition goes false so it will not double-shutdown on subsequent tests.

I used the console check rather than the suggested `LoggerProvider` check because the global logger provider is not reset by shutdown, so that condition would stay true and re-trigger a shutdown on every later test in the file.

Verified: the per-test probe now reports zero tests leaving console patched, an end-of-suite probe reports `restored: true` (it was `false` on both the base commit and my first attempt), and the full unit suite is 248 passing. Note `shutdownAzureMonitor()` calls `autoCollectLogs.shutdown()` unguarded, so an unconditional teardown would throw for the tests that never initialize - hence the conditional.

package.json
```diff
@@ -77,17 +77,17 @@
     "@opentelemetry/exporter-metrics-otlp-http": "^0.220.0",
     "@opentelemetry/exporter-metrics-otlp-proto": "^0.220.0",
     "@opentelemetry/exporter-trace-otlp-http": "^0.220.0",
+    "@opentelemetry/instrumentation-console": "^0.2.0",
```
hectorhdzg: We should include this in both distros and consume here with a setting

package.json
```diff
@@ -77,17 +77,17 @@
     "@opentelemetry/exporter-metrics-otlp-http": "^0.220.0",
```
hectorhdzg, line 77: Can we use Msft distro as upstream to get rid of all OTLP direct dependencies and code?

package.json
```diff
@@ -77,17 +77,17 @@
...
     "@opentelemetry/sdk-metrics": "^2.9.0",
     "@opentelemetry/sdk-trace-base": "^2.9.0",
     "@opentelemetry/sdk-trace-node": "^2.9.0",
-    "@opentelemetry/semantic-conventions": "^1.43.0",
-    "diagnostic-channel": "1.1.1",
-    "diagnostic-channel-publishers": "1.0.8"
+    "@opentelemetry/semantic-conventions": "^1.43.0"
   },
   "overrides": {
+    "@opentelemetry/api-logs": "$@opentelemetry/api-logs",
```
hectorhdzg: Why this override? is there a breaking change we want to avoid?

README.md
```diff
@@ -157,8 +157,8 @@ separately from clients created with `new appInsights.TelemetryClient()`.
...
-| noPatchModules                | Disable individual monkey-patches. Set `noPatchModules` to a comma separated list of packages to disable. e.g. `"noPatchModules": "console,redis"` to avoid patching the console and redis packages. The following modules are available: `azuresdk, bunyan, console, mongodb, mongodb-core, mysql, redis, winston, pg`, and `pg-pool`. Visit the [diagnostic-channel-publishers' README](https://github.com/microsoft/node-diagnostic-channel/blob/master/src/diagnostic-channel-publishers/README.md) for information about exactly which versions of these packages are patched. |
+| noDiagnosticChannel           | In order to track context across asynchronous calls, some changes are required in third party libraries such as mongodb and redis. By default ApplicationInsights will use the appropriate OpenTelemetry instrumentation for each library. This property is to disable the feature. Note that by setting this flag, events may no longer be correctly associated with the right operation.  |
```
hectorhdzg: Lets add these two as deprecated as well

test/unitTests/agent/aksLoader.tests.ts
```diff
@@ -5,22 +5,27 @@ import { logs } from "@opentelemetry/api-logs";
...
     beforeEach(() => {
         originalEnv = process.env;
+        originalConsoleLog = console.log;
     });
 
-    afterEach(() => {
-        disposeConsole();
+    afterEach(async () => {
+        // initialize() installs global providers and patches console; both leak into later suites without a shutdown.
+        if (console.log !== originalConsoleLog) {
```
hectorhdzg, line 26: Why this only applicable to AKS and not other RP scenarios?

test/functionalTests/runner/taskExpectations.js
```diff
@@ -165,7 +165,7 @@ module.exports = {
         "MessageData",
         (telemetry) => {
             return telemetry.data.baseData.message === "Test console.error" &&
-            telemetry.data.baseData.severityLevel === "Warning";
+            telemetry.data.baseData.severityLevel === "Error";
```
rads-1996, line 168: why has the severity been changed?

package.json
JacksonWeber: Updating to use the instrumentation in the Az Mon distro instead.

package.json
JacksonWeber: Think the agent just started getting creative here. Reigned in.

test/unitTests/agent/aksLoader.tests.ts
JacksonWeber: Other RPs use `sandbox.stud(agent, "initialize")` so AKS is the only one with actual patched modules to clean up.

test/functionalTests/runner/taskExpectations.js
JacksonWeber: This was due to a behavior change in the old diagnostic channel. It guessed intent from the message text instead of just from understanding `console.warn` vs `console.error`

package.json
JacksonWeber: Happy to create this as a follow up PR, but wanted to keep this one focused if possible. I have removed our dependency on the console instrumentation in this package though.

hectorhdzg (approved):

